PR #12, opened 2023-08-02
Instantiate `python-project-template`

tothtamas28:


F-WRunTime:
We have a PYpi account if you wish to use it @tothtamas28 to publish your work I believe it must be setup as public. But I haven't looked too much into publishing private files to Pypi. Something I can look into if you're interested. No one has been interested yet in using this repository.

PR #54, opened 2023-09-27
Execution of contract initialization code

nwatson22:
Ports https://github.com/runtimeverification/evm-semantics/pull/2020 to this repo


pyproject.toml
```diff
@@ -15,8 +15,7 @@ kontrol = "kontrol.__main__:main"
 
 [tool.poetry.dependencies]
 python = "^3.10"
-kevm-pyk = { git = "https://github.com/runtimeverification/evm-semantics.git", tag = "v1.0.307", subdirectory = "kevm-pyk" }
-
+kevm-pyk = { git = "https://github.com/runtimeverification/evm-semantics.git", branch = "noah/contract-initialization-2", subdirectory = "kevm-pyk" }
```
ehildenb, line 18: Should this change be backed out?

flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "Kontrol";
 
   inputs = {
-    kevm.url = "github:runtimeverification/evm-semantics/v1.0.307";
+    kevm.url = "github:runtimeverification/evm-semantics/noah/contract-initialization-2";
     nixpkgs.follows = "kevm/nixpkgs";
     k-framework.follows = "kevm/k-framework";
```
ehildenb, line 7: This change should be backed out too.

ehildenb:
This should be merged after: https://github.com/runtimeverification/kontrol/pull/63 (and merge conflicts resolved).

flake.nix
nwatson22: It also depends on this branch, https://github.com/runtimeverification/evm-semantics/pull/2096, although I will back out these changes here now because the checks have passed.

pyproject.toml
nwatson22: Done.

src/kontrol/prove.py
```diff
@@ -192,11 +235,13 @@ def _run_cfg_group(
     trace_rewrites: bool,
     auto_abstract_gas: bool,
     port: int | None,
+    run_constructor: bool = False,
 ) -> dict[tuple[str, int], tuple[bool, list[str] | None]]:
+    print(f'run_constructor={run_constructor}')
```
palinatolmach, line 240: Just want to confirm: is it intentional or left over from debugging?

src/kontrol/solc_to_k.py
```diff
@@ -66,6 +66,70 @@ def solc_to_k(
...
+            sort: KSort,
+        ) -> None:
+            self.signature = f'{contract_name}_init'
+            self.arg_names = tuple([f'V{i}_{input["name"].replace("-", "_")}' for i, input in enumerate(abi['inputs'])])
+            self.arg_types = tuple([input['type'] for input in abi['inputs']])
+            self.contract_name = contract_name
+            self.contract_digest = contract_digest
+            self.contract_storage_digest = contract_storage_digest
+            self.sort = sort
+            # TODO: Check that we're handling all state mutability cases
```
palinatolmach, line 95: I understand it's copied from a similar `Method` class, but should we address or remove this `TODO`? Other state mutability cases are `pure`, `view`, and `nonpayable`, and I'm not sure if we want to handle them differently. It doesn't have to be addressed in this PR though.

src/kontrol/foundry.py
```diff
@@ -387,8 +387,14 @@ def get_optional_proof(self, test_id: str) -> Proof | None:
             return Proof.read_proof_data(self.proofs_dir, test_id)
         return None
 
-    def get_method(self, test: str) -> Contract.Method:
+    def get_method(self, test: str) -> Contract.Method | Contract.Constructor:
         contract_name, method_name = test.split('.')
+
+        if method_name == 'init':
+            constructor = self.contracts[contract_name].constructor
+            assert constructor is not None
```
palinatolmach, line 395: Should we throw an error if a constructor is not present? E.g., something like
```python
raise ValueError(Contract {contract.name} does not have a constructor.')
```

palinatolmach (approved):

src/kontrol/foundry.py
nwatson22: Fixed.

src/kontrol/prove.py
nwatson22: Removed.

src/kontrol/solc_to_k.py
nwatson22: I think `nonpayable` is handled implicitly by it not being payable, therefore we return a callvalue with a concrete value of zero. But I'm not aware of what we want to do with `pure` and `view`, or if we want to do something. Are those modifiers just something that is enforced by the solidity compiler or is it something we could take advantage of in verification?

This comment was originally added in https://github.com/runtimeverification/evm-semantics/pull/1318.

PR #59, opened 2023-09-29
Hotfix/docker development

F-WRunTime:


F-WRunTime:
@tothtamas28 This has produced a non issue image . I'd like to run through a formal run and determine if we can 
image

F-WRunTime:
Pushing to internal access registry on the left is an attempt to test if there is a caching issue with the internal pull through / caching server

.github/workflows/master-push.yml
```diff
@@ -98,12 +98,12 @@ jobs:
...
+          docker build . --no-cache --tag ${TAG} --build-arg K_VERSION=${K_VERSION}
 
       - name: 'Run Docker image'
         run: docker run --rm ${TAG} kontrol --help
 
       - name: 'Push Docker image to Docker Hub'
         run: |
           echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login --username rvdockerhub --password-stdin
-          docker image push ${TAG}
+          docker push ${TAG}
```
tothtamas28, line 109: Isn't one just an alias for the other?

tothtamas28 (approved):

F-WRunTime:
The --no-cache will not allow buildkit to use existing container layers to build the image; after a test I feel more confident this couldbe the issue; I was unable to reproduce the problem with our internal registry reliably to find an alternative solution. 
Docker does not change images after build and during push so this does not 100% explain the issue of working on the built machine but fails once pulled from dockerhub remote. I have noticed the image layers are not the same post push and pull. 
Why this is, is unclear; and unable to reproduce currently without using the remote dockerhub (internal registry did not do this.)

--no-cache produced a valid image without permission issues after pull down on a client. Using no cache will at least make sure each build uses fresh layers and does not rely on any remote or locally corrupted layers that could be causing this issue.

PR #62, opened 2023-10-02
Store external calls as subproofs

anvacaru:
First pull request for: https://github.com/runtimeverification/kontrol/issues/130

- [x] Extend `Contract.Method` class to contain the list of the external functions that the current method calls.
- [x] Add external calls as subproofs in`method_to_apr_proof`.


anvacaru:
Closing this due to staleness.

PR #63, opened 2023-10-02
Refactor proof setup

tothtamas28:
This PR refactors the first step of `foundry_prove`, namely selecting and initializing tests for proving.
* Extract class `FoundryTest` that encapsulates a `Contract`, a `Method` and a version `int`. This way steps that follow don't have to pass and validate test signatures as `str`.
* Extract functions for loading `FoundryTest` instances for test and `setUp()` methods.
* Pull loading the `setUp` final state from `_get_final_accounts_cell` to `method_to_apr_proof`. This makes `_method_to_cfg` a pure function.
* Extract function `method_to_initialized_cfg` that computes the initial proof `KCFG`. 
* Other small code quality improvements.


ehildenb:
It looks like `foundry_prove` is now completely ignoring the `exclude_tests` parameter. I guess that's because of a prior code change that somehow made us ignore that. So I guess we can just remove that parameter to both `foundry_prove`, the CLI, nad `matching_tests` function, since no one has complained.

src/kontrol/prove.py
```diff
@@ -241,105 +246,132 @@ def init_and_run_proof(_init_problem: tuple[str, str, int]) -> tuple[bool, list[
...
+
+    _LOGGER.info(f'Expanding macros in target state for test: {test.id}')
+    target_term = kcfg.node(target_node_id).cterm.kast
+    target_term = KDefinition__expand_macros(foundry.kevm.definition, target_term)
+    target_cterm = CTerm.from_kast(target_term)
+    kcfg.replace_node(target_node_id, target_cterm)
+
+    if simplify_init:
+        _LOGGER.info(f'Simplifying KCFG for test: {test.id}')
+        kcfg_explore.simplify(kcfg, {})
```
ehildenb, line 356: We should always do this actually, and remove this option.

ehildenb (approved):

src/kontrol/prove.py
tothtamas28: Opened an issue: https://github.com/runtimeverification/kontrol/issues/67

nwatson22 (approved):

PR #77, opened 2023-10-05
Show kontrol version number with the  `kontrol version` command

spencerhaoxiao:
1. Add a constant `VERSION` defintion in `src/kontrol/__init__.py`
2. Output the version number for `kontrol version`
3. Add the CI script to update the value of `VERSION` in `src/kontrol/__init__.py`


src/kontrol/__main__.py
```diff
@@ -419,7 +420,9 @@ def parse(s: str) -> list[T]:
 
     command_parser = parser.add_subparsers(dest='command', required=True)
 
-    command_parser.add_parser('version', help='Print out version of Kontrol command.')
+    command_parser.add_parser(
+        'version', help='Print out version of Kontrol command.', parents=[kontrol_cli_args.logging_args]
```
palinatolmach, line 424: Do we need `logging_args` here? Running `kontrol version` with `--debug` or `--verbose` seems to produce the same output as without args, but maybe it's good to keep it consistent with other subcommands.

src/kontrol/__main__.py
```diff
@@ -72,7 +73,7 @@ def main() -> None:
 
 
 def exec_version(**kwargs: Any) -> None:
-    raise NotImplementedError()
+    print(f'Kontrol Version: {VERSION}')
```
palinatolmach, line 76:
```suggestion
    print(f'Kontrol version: {VERSION}')
```
Making the formatting consistent with e.g., how `K` or `kore-rpc` versions are returned:
```
kompile --version
K version:    v6.0.133
```
```
kore-rpc --version
Kore version 0.60.0.0
```

palinatolmach (approved):

src/kontrol/__main__.py
spencerhaoxiao: There is a problem is with the logic to decide the logging level. Anyway, I modified that part of the code to give the default logging level is neither 'debug' or 'verbose' is specified.

PR #79, opened 2023-10-05
Add a new entry point for the debugger.

RaoulSchaffranek:
Previously, the Solidity Debugger relied on syscalls to call Kontrol. In our continuous effort to optimize, we're transitioning to embed Kontrol directly as a Python library. To facilitate this transition, we've introduced a new entry point for the debugger to access the Prover instances. This PR introduces the `foundry_debug` entry point, which mirrors `foundry_prove` in functionality but returns the prover instance directly rather than the final result.

Notice, this PR should be merged after this related PR is merged into KEVM:
https://github.com/runtimeverification/evm-semantics/pull/2106


RaoulSchaffranek:
I'm closing this issue because the logic is better implemented in the debugger itself.

PR #88, opened 2023-10-09
Clarify `--test` functionality for kontrol prove

anvacaru:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/2063

Changes:
- improved help message
- improved error message when multiple signatures are identified
- refactored regex search to avoid repetition


src/kontrol/foundry.py
```diff
@@ -253,18 +253,14 @@ def _escape_brackets(regs: list[str]) -> list[str]:
...
-            if any(re.search(t, test) for t in tests):
-                matched_tests.add(test)
-        for test in all_non_tests:
-            if any(re.search(t, test) for t in tests):
-                matched_tests.add(test)
+        matched_tests = set()
+        unfound_tests = set(tests)
+        for test in tests:
+            for possible_match in all_tests + all_non_tests:
+                if re.search(test, possible_match):
```
anvacaru, line 261: @tothtamas28 would it make sense to have a single `re.search` on a string that has `all_tests + all_non_tests` instead?

palinatolmach (approved):

src/kontrol/foundry.py
tothtamas28: Not really, for two reasons:
* `re.search` returns the first match. You could use `re.findall` instead, but that still requires  post-processing to correlate matches to tests`.
* if there's a single string, the match might span the section of multiple tests (if I understand the question correctly), which doesn't make sense.

src/kontrol/__main__.py
```diff
@@ -492,7 +492,11 @@ def _parse_test_version_tuple(value: str) -> tuple[str, int | None]:
         dest='tests',
         default=[],
         action='append',
-        help='Limit to only listed tests, ContractName.TestName',
+        help=(
+            "Specify the contract method to test in the format 'ContractName.MethodName'. If a method is overloaded, "
```
palinatolmach, line 496: One minor comment I have here is that I usually see “functions” instead of “methods” in the context of Solidity smart contracts. I agree that it doesn’t seem right, cause, unlike methods, functions are typically not associated with objects (or something like this), but, in my experience, methods are usually referred to as “Solidity functions” most of the time — probably because of the `function` keyword used in the definition. That doesn't have to addressed, just a suggestion to consider.

src/kontrol/foundry.py
anvacaru: Thanks!

src/kontrol/__main__.py
anvacaru: Addressed this, thank you! I got a bit confused because we're using both `method` and `function` keywords in the codebase. Left `function` to align with all the other help messages.

PR #96, opened 2023-10-10
The order of notifying downstreams happens before artifacts are ready…

F-WRunTime:
Fixing here


.github/workflows/master-push.yml
```diff
@@ -107,3 +76,35 @@ jobs:
         run: |
           echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login --username rvdockerhub --password-stdin
           docker image push ${TAG}
+
+          
+  cut-release:
+    name: 'Cut Release'
+    runs-on: ubuntu-latest
+    steps:
```
tothtamas28, line 84: Did you forget to add the `needs`  key, or is it just a code movement?

F-WRunTime: It was just a code movement. But you're right that would ensure they don't run before the other. let me add this.

.github/workflows/master-push.yml
```diff
@@ -107,3 +76,36 @@ jobs:
         run: |
           echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login --username rvdockerhub --password-stdin
           docker image push ${TAG}
+
+          
+  cut-release:
+    name: 'Cut Release'
+    runs-on: ubuntu-latest
+    needs: [nix-cache, dockerhub]
```
tothtamas28, line 84: Until the `macos-13` Nix cache job is fixed, I think we should just go with

```suggestion
    needs: dockerhub
```

Otherwise, we won't be able to make a release.

tothtamas28 (approved):

PR #101, opened 2023-10-13
Parallel exploration

nwatson22:
Addresses https://github.com/runtimeverification/evm-semantics/issues/2009


nwatson22:
I'm trying to figure out now why this is still using so much more memory such that it is running out on CI.

PR #103, opened 2023-10-13
Make test selection more similar to Foundry

nwatson22:
Addresses #104
Matches test selection behavior more closely to foundry. Instead of selecting tests with `--test` where the user is required to use regex to select exactly one test with each supplied `--test` flag, with this change you now use `--match-test` which will run any number of tests that match the given regex.

The differences still from foundry are:
- Non-test functions are considered
- Foundry only allows one `--match-test` flag. We already support multiple so I don't see a reason to remove this ability.


palinatolmach:
We should probably update `help` for `--match-test` to reflect these changes, i.e., that if a function is overloaded, one can provide a full signature, but otherwise all tests matching the regex or string prefix will be executed: 
https://github.com/runtimeverification/kontrol/blob/b4f2513fed8b01e1dc8d00f605acdfcf8a40e698/src/kontrol/__main__.py#L505.

palinatolmach:
I realised that another diff from Foundry's interface is that we request the user to prove `ContractName.` in the prefix, but I don't know if we want to address it now.

palinatolmach:
@anvacaru @RaoulSchaffranek @lucasmt @PetarMax can I ask for your feedback on this change before we merge it? Basically, it's renaming `--test` to `--match-test` (to align with Foundry's options) and execute _all_ tests that match the provided string prefix, not a unique one. 

I.e., in `RollTest` with three functions `test_roll_set()`, `test_roll_set(uint256 x)`, `test_roll_setup()`, all three will be executed by 
```
kontrol prove --match-test RollTest.test_roll_set
```
But only `test_roll_set(uint256 x)` will be executed by
```
kontrol prove --match-test 'RollTest.test_roll_set(uint256)'
```

That's what `forge test --match-test RollTest.test_roll_set` does too ( I just checked), but please let me and @nwatson22  know if you have any usability concerns.

palinatolmach (commented):
Could you please update `help` for `--match-test` to reflect these changes, i.e., that if a function is overloaded, one can provide a full signature, but otherwise all tests matching the regex or string prefix will be executed: 
https://github.com/runtimeverification/kontrol/blob/b4f2513fed8b01e1dc8d00f605acdfcf8a40e698/src/kontrol/__main__.py#L505.

nwatson22:
> Could you please update `help` for `--match-test` to reflect these changes, i.e., that if a function is overloaded, one can provide a full signature, but otherwise all tests matching the regex or string prefix will be executed?
> 
> https://github.com/runtimeverification/kontrol/blob/b4f2513fed8b01e1dc8d00f605acdfcf8a40e698/src/kontrol/__main__.py#L505

@palinatolmach I updated it to say: "Specify contract function(s) to test using a regular expression. This will match functions based on their full signature,  e.g., 'ERC20Test.testTransfer(address,uint256)'. This option can be used multiple times to add more functions to test." I hope this gets across that they can use the full signature if necessary or omit it if it's not overloaded. It also removes the explicit request to include the contract name, as this is not actually required. (You  can omit the contract name and it will match methods of the specified name in all the available contracts).

palinatolmach (approved):

PR #105, opened 2023-10-16
Allow users to select version of solc via kup

goodlyrottenapple:
This change removes the default pinned `foundry` and `solc` versions and instead uses the system installed versions of these tools. If the user wants to pin down a specific version of `solc` (e.g. for reproducibility in engagements or CI), they can install a version of `kontrol` with a specific `solc` compiler via:

```
kup install kontrol.solc_0_8_13
```


flake.nix
```diff
@@ -12,58 +12,69 @@
...
+                      (solc.mkDefault final solc_version)
+                    ])
+                } --set NIX_LIBS "${nixLibs prev}" --set KEVM_DIST_DIR ${
+                  prev.kevm k-framework.packages.${prev.system}.k
+                }
+              '';
+            }).overrideAttrs (old: {
+              passthru = old.passthru // (if solc_version == null then {
+                # list all supported solc versions here
+                solc_0_8_13 = kontrol { solc_version = final.solc_0_8_13; };
```
goodlyrottenapple, line 73: if there are any other versions of solc we want to support, we should add them here as:

```nix
solc_0_x_yy = kontrol { solc_version = final.solc_0_x_yy; };
```

Baltoli (approved):
Looks good!

flake.nix
spencerhaoxiao: Can we run multiple versions of solc simultaneously in kup? There are cases in which different versions of solc to compile a solidity source file (depending on the solidity pragma in the source file), thus we need to run multiple versions of solc to compile all the source files in a foundry project. Foundry allows this through automatically installing the required versions.

PR #112, opened 2023-10-18
Enable manual and auto abstraction for cells

ehildenb:
~Blocked on: https://github.com/runtimeverification/kontrol/pull/113~

This PR enables a feature discussed with @geo2a for keeping configurations smaller. In particular:

- `KEVMSemantics` is now extended by `KontrolSemantics`, which overrides the abstraction operator.
- `kontrol prove ...` gets a new feature `--auto-abstract-cell CELL_NAME_1 --auto-abstract-cell CELL_NAME_2`, which allows the user to specify that the given cell should be auto-abstracted to a variable every iteration of exploration.
- A new CLI argument is added `kevm abstract TEST_ID --node NODE_ID_1 --node NODE_ID_2 --cell CELL_NAME_1 --cell CELL_NAME_2`, which allows users to manually abstract a given node in the KCFG.
- A test of the abstraction CLI functionality is added.
- The `--auto-abstract-gas` flag is implemented in terms of this new functionality, and a warning thrown for using the old flag.

An example of using this would be:

```
kontrol prove ... --test MyContract.test_foo --auto-abstract-cell localMem --auto-abstract-cell gas --auto-abstract-cell refund
kontrol abstract MyContract.test_foo --node 3 --node 5 --cell localMem --cell callData
```


ehildenb:
@PetarMax @lucasmt it would be good to know if this is useful for blockswap, @geo2a mentioned a given problem you guys are working on where exploration takes quite a long time. Perhaps this can be used to speed up exploration, and the following strategy can be taken:

- Run the proof with `kontrol prove ... --auto-abstract-cell ...` (list a bunch of cells which grow over time and cause problems).
- If the proof passes in this case, then we're done (as abstraction leads to over-approximation of the state space).
- If the proof doesn't pass, we can still use the generated basic blocks for faster symbolic execution:
  - Do `kontrol show --to-module ...`, and add the generated rules (basic blocks) to the `lemmas.k` file you guys use.
  - Rekompile.
  - Rerun the proof with auto abstraction turned off (this time using the generated basic blocks for faster execution).

lucasmt:
This could be useful I think. The issue we have is that the `<memoryUsed>` cell is not being simplified by the booster, so maybe abstracting it away could be an option until the simplification capabilities of the booster improve. I don't know if there's any point when we actually need the value stored in that cell.

PetarMax:
> This could be useful I think. The issue we have is that the `<memoryUsed>` cell is not being simplified by the booster, so maybe abstracting it away could be an option until the simplification capabilities of the booster improve. I don't know if there's any point when we actually need the value stored in that cell.

`<memoryUsed>` is simplified by the Haskell backend, though. From what I see, the only place it's used is:
```
rule <k> MSIZE    => 32 *Word MU         ~> #push ... </k> <memoryUsed> MU </memoryUsed>
```
and then `MSIZE` is only used for
```
rule <k> #gasExec(SCHED, MSIZE)          => Gbase < SCHED > ... </k>
```
so it doesn't seem very useful...

Am I missing something?

yale-vinson:
@ehildenb Does someone from Kontrol need to review this PR so that it can be merged?

PetarMax:
I think this PR brings a potentially useful feature to the table and I am ok with it being merged - it is not passing CI, however.

src/kontrol/foundry.py
```diff
@@ -784,6 +815,30 @@ def foundry_get_model(
...
+
+    for cell in cells:
+        cell_name = cell.upper() + '_CELL'
+        cell_contents = cterm.cell(cell_name)
+        new_cell_contents = cell_contents
+        if not _is_lone_var(cell_contents):
+            new_cell_contents = abstract_term_safely(kast, base_name=cell_name)
+            if cell_name == 'GAS_CELL' and type(cell_contents) is KApply and cell_contents.label.name == 'infGas':
+                if _is_lone_var(cell_contents.args[0]):
+                    new_cell_contents = cell_contents
```
PetarMax, line 833: Does this say that if we already have just `#gas(VARIABLE)`, we don't do further changes? Why don't we do similar checks for other cells?

EDIT: We do do the checks, I just need to look at the code. I've understood the reasoning.

PetarMax (approved):

anvacaru:
CI is failing because the expected output needs to be updated.

palinatolmach:
@JuanCoRo @PetarMax we're running into some issues seemingly caused by a huge configuration size in OP proofs. Do you think it's worth to bring this PR back?

PetarMax:
Possibly... but which cells specifically make the configuration huge and what are the precise issues?

PR #114, opened 2023-10-19
Simplify returning of proof results

nwatson22:
Instead of returning a complicated tuple containing the proof status and failure logs, we can just return the Proof objects and the receiver can inspect them for this info. Enabled by https://github.com/runtimeverification/pyk/pull/654.


src/tests/integration/test_foundry_prove.py
```diff
@@ -327,23 +328,21 @@ def test_foundry_remove_node(
...
 
 
-def assert_pass(test: str, prove_res: dict[tuple[str, int], tuple[bool, list[str] | None]]) -> None:
-    id = id_for_test(test, prove_res)
-    passed, log = prove_res[(test, id)]
-    if not passed:
-        assert log
-        pytest.fail('\n'.join(log))
+def assert_pass(test: str, prove_res: list[Proof]) -> None:
+    proof = single([proof for proof in prove_res if proof.id.split(':')[0] == test])
```
tothtamas28, line 332: Consider

```suggestion
def assert_pass(proof: Proof) -> None:
```

and checking for a single proof in the caller.

src/tests/integration/test_foundry_prove.py
```diff
@@ -327,23 +328,21 @@ def test_foundry_remove_node(
...
-    id = id_for_test(test, prove_res)
-    passed, log = prove_res[test, id]
-    assert not passed
-    assert log
-
-
-def id_for_test(test: str, prove_res: dict[tuple[str, int], tuple[bool, list[str] | None]]) -> int:
-    return single(_id for _test, _id in prove_res.keys() if _test == test and _id is not None)
+def assert_fail(test: str, prove_res: list[Proof]) -> None:
+    proof = single([proof for proof in prove_res if proof.id.split(':')[0] == test])
```
tothtamas28, line 342:
```suggestion
def assert_fail(proof: Proof) -> None:
```

src/tests/integration/test_foundry_prove.py
```diff
@@ -386,20 +385,20 @@ def test_foundry_resume_proof(
         port=server.port,
         bug_report=bug_report,
     )
-    id = id_for_test(test, prove_res)
 
-    proof = foundry.get_apr_proof(f'{test}:{id}')
+    proof = foundry.get_apr_proof(f'{single(prove_res).id}')
```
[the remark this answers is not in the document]
nwatson22: You're right, this is redundant. I've fixed it.

src/tests/integration/test_foundry_prove.py
nwatson22: Done.

src/tests/integration/test_foundry_prove.py
nwatson22: Done.

src/tests/integration/test_foundry_prove.py
```diff
@@ -386,20 +385,20 @@ def test_foundry_resume_proof(
...
     assert proof.pending
 
     prove_res = foundry_prove(
         foundry_root,
-        tests=[(test, id)],
+        tests=[(test, None)],
         auto_abstract_gas=True,
-        max_iterations=6,
+        max_iterations=10,
```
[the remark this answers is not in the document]
nwatson22: It was intentional. The additional iterations are needed since #54 but I think the test was broken so it wasn't failing properly. This PR should fix that.

tothtamas28 (approved):

PR #125, opened 2023-10-24
Add maude backend to Kontrol

dwightguth:
Blocked on https://github.com/runtimeverification/evm-semantics/pull/2118

This adds the option to build the maude backend with `kontrol build`, and adds support for using the maude backend with `kontrol prove`.


anvacaru:
1. Is there any additional setup required that should be mentioned or documented in the Readme?
2. Could you include a CI stage to build and run this new backend?

dwightguth:
There is additional setup in order to actually build the backend and run it, yes, but that process has not yet been streamlined to the point where I'm comfortable including it in a downstream build. Similarly, enough of the tests we are working with still fail that I would prefer to wait on adding it to CI. For example, right now on the latest master of the maude backend, the test suite I am working with does not even terminate.

The work you are talking about will come, but it's not yet time to do it.

src/kontrol/cli.py
```diff
@@ -52,3 +53,9 @@ def k_gen_args(self) -> ArgumentParser:
             help='Extra modules to import into generated main module.',
         )
         return args
+
+    @cached_property
+    def kompile_target_args(self) -> ArgumentParser:
+        args = ArgumentParser(add_help=False)
+        args.add_argument('--target', type=KompileTarget, help='[haskell-booster|maude]')
```
anvacaru, line 60: I suppose this flag could be used to pass other targets by mistake. You could restrict them using `choices=[KompileTarget.HASKELL_BOOSTER, KompileTarget.MAUDE],`.

anvacaru (approved):
Left one suggestion, other than that it looks good to me 👍 !

src/kontrol/cli.py
dwightguth: I believe I tried that and it broke, but I'll try again and see if I am misremembering.

Issue #58, opened 2023-09-29
Update documentation link in output


palinatolmach:
As identified by @reekamaharaj, the link to documentation included in the output of `kontrol prove` is outdated:
``` 
Access documentation for KEVM foundry integration at https://docs.runtimeverification.com/kevm-integration-for-foundry/
```
It should be changed to http://docs.runtimeverification.com/kontrol.

PR #129, opened 2023-10-25
Updated documentation link in output

lisandrasilva:
The link to the documentation included in the output of `kontrol prove` was outdated, as pointed out in [Issue 58](https://github.com/runtimeverification/kontrol/issues/58). This PR updates that documentation link to  http://docs.runtimeverification.com/kontrol.


src/kontrol/foundry.py
```diff
@@ -353,7 +353,7 @@ def help_info() -> list[str]:
             )
         res_lines.append('')
         res_lines.append(
-            'Access documentation for KEVM foundry integration at https://docs.runtimeverification.com/kevm-integration-for-foundry/'
+            'Access documentation for KEVM foundry integration at http://docs.runtimeverification.com/kontrol'
```
palinatolmach, line 356: I'd rather use `https` link:
```suggestion
            'Access documentation for KEVM foundry integration at https://docs.runtimeverification.com/kontrol'
```
I think there's auto-redirection from `http://docs.runtimeverification.com/kontrol` to `https://docs.runtimeverification.com/kontrol`, so it's mostly aesthetics.

palinatolmach (changes requested):

src/kontrol/foundry.py
lisandrasilva: No worries, I could have noticed as well. I just changed to `https`.

palinatolmach (approved):

PR #132, opened 2023-10-26
Warn user if an outdated K version is being used

lisandrasilva:
According to [Issue 2097](https://github.com/runtimeverification/evm-semantics/issues/2097), KEVM doesn't throw an error if an older version of K is being used, which leads to incorrect or outdated results produced by KEVM and Kontrol. This PR addresses the issue by warning the user if an outdated K version is being used.


src/kontrol/__main__.py
```diff
@@ -62,6 +64,14 @@ def main() -> None:
     args = parser.parse_args()
     logging.basicConfig(level=_loglevel(args), format=_LOG_FORMAT)
 
+    expected_k_version = KVersion.parse(f'v{pyk.K_VERSION}')
+    actual_k_version = k_version()
+
+    if expected_k_version != actual_k_version:
+        print(
+            f'WARNING: An outdated version of K (v{actual_k_version}) is being used. Updating K to v{expected_k_version} is RECOMMENDED'
+        )
```
tothtamas28, line 74:
```suggestion
        _LOGGER.warning(
            f'An outdated version of K (v{actual_k_version}) is being used. Updating K to v{expected_k_version} is RECOMMENDED'
        )
```

src/kontrol/__main__.py
```diff
@@ -62,6 +64,14 @@ def main() -> None:
     args = parser.parse_args()
     logging.basicConfig(level=_loglevel(args), format=_LOG_FORMAT)
 
+    expected_k_version = KVersion.parse(f'v{pyk.K_VERSION}')
+    actual_k_version = k_version()
+
+    if expected_k_version != actual_k_version:
```
tothtamas28, line 71: `KVersion` has an optional attribute `git` that is set for a source build. To avoid emitting the warning for all source builds, a custom comparison should be implemented.

src/kontrol/__main__.py
tothtamas28: The wording suggests the issue is with an outdated K version, but actually the warning is emitted on a version that does not match the expected one.

tothtamas28, line 74: Please extract a function `_check_k_version`.

lisandrasilva: From where?

tothtamas28: I mean you can put this code in a function definition:

```python3
def _check_k_version() -> None:
    expected_k_version = ...
    ...
```

and then call the function in `main`:

```python3
def main() -> None:
    ...
    _check_k_version()
```

This improves readability.

lisandrasilva: Ahh ok got it! Thanks! I just added the function in `__main__.py` in the Kontrol repo, but maybe it would make sense to add this function in the pyk repo here: https://github.com/runtimeverification/pyk/blob/master/src/pyk/kbuild/utils.py
What do you think?

lisandrasilva: I addressed this. Please see if it's better now.

src/kontrol/__main__.py
```diff
@@ -71,6 +75,14 @@ def main() -> None:
     execute(**vars(args))
 
 
+def _check_k_version() -> None:
+    expected_k_version = KVersion.parse(f'v{pyk.K_VERSION}')
+    actual_k_version = k_version()
+
+    if expected_k_version != actual_k_version:
+        _LOGGER.warning(f'K version {expected_k_version} was expected but K version {actual_k_version} is being used. ')
```
tothtamas28, line 83:
```suggestion
        _LOGGER.warning(f'K version {expected_k_version} was expected but K version {actual_k_version} is being used.')
```

src/kontrol/__main__.py
tothtamas28: Looks good.

src/kontrol/__main__.py
lisandrasilva: I do not really understand this comment. Can you please clarify?

lisandrasilva: Ok, so we just need to compare `major`, `minor` and `patch` attributes, is that correct?

lisandrasilva: I think I understood but please check the latest commit! Thanks!

src/kontrol/__main__.py
```diff
@@ -71,6 +75,25 @@ def main() -> None:
...
 
+def _check_k_version() -> None:
+    expected_k_version = KVersion.parse(f'v{pyk.K_VERSION}')
+    actual_k_version = k_version()
+
+    if (
+        expected_k_version != actual_k_version
+        and not _compare_versions(expected_k_version, actual_k_version)
+        and not _compare_versions(actual_k_version, expected_k_version)
+    ):
```
tothtamas28, line 86: Please continue moving logic into `_compare_versions` until this condition simplifies to
```suggestion
    if not _compare_versions(expected_k_version, actual_k_version):
```

Then we can focus on simplifying the code in that function.

lisandrasilva: Ok, I just refactored! Please check that the logic is ok. I can refactor the `_compare_versions` function even further to have just one `return True` and `return False`. Let me know.

src/kontrol/__main__.py
```diff
@@ -71,6 +75,29 @@ def main() -> None:
...
+def _check_k_version() -> None:
+    expected_k_version = KVersion.parse(f'v{pyk.K_VERSION}')
+    actual_k_version = k_version()
+
+    if not _compare_versions(expected_k_version, actual_k_version):
+        _LOGGER.warning(f'K version {expected_k_version} was expected but K version {actual_k_version} is being used.')
+
+
+def _compare_versions(ver1: KVersion, ver2: KVersion) -> bool:
+    if ver1 != ver2:
```
tothtamas28, line 87: I'd structure this as follows:

```python3
if ver1.major != ver2.major or ver1.minor != ver2.minor or ver1.patch != ver2.patch:
    return False

if ver1.git == ver2.git:
    return True

if ver1.git and ver2.git:
    return False

git = ver1.git or ver2.git
assert git  # git is not None for exactly one of ver1 and ver2
return not git.ahead and not git.dirty
```

lisandrasilva: Nice! Thank you! I just pushed.

tothtamas28 (approved):

PR #138, opened 2023-10-27
Fix kup version of kontrol by building the foundryx target

goodlyrottenapple:


tothtamas28:
I tested this with both `nix shell` and `kup shell`, works as expected.

tothtamas28 (approved):

PR #146, opened 2023-10-27
Add job to test Nix build 

### LINKED ISSUE #76: Run Nix tests on CI
Currently, it seems that we don't run any nix tests on CI. We should be testing that it can build/install/run tests via Nix as well as on normal distros.

tothtamas28:
Fixes #76


.github/workflows/test-pr.yml
```diff
@@ -180,3 +180,49 @@ jobs:
...
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 60
+    steps:
+      - name: 'Check out code'
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: 'Install Foundry'
+        uses: foundry-rs/foundry-toolchain@v1
```
tothtamas28, line 200: Installing Foundry separately is necessary in the current setup.

goodlyrottenapple: If you use `nix build .#kontrol.solc_0_8_13`, foundry and solc will be added to the kontrol executable's PATH

.github/workflows/test-pr.yml
```diff
@@ -180,3 +180,49 @@ jobs:
...
         run: |
           docker stop --time=0 ${CONTAINER_NAME}
+
+  nix:
+    needs: code-quality-checks
+    name: 'Nix Tests'
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ubuntu-latest, macos-13]
```
goodlyrottenapple, line 190: we should be testing this on our m1 machines as well, since we want to set up the cachix cache next for fast kontrol installs via kup

goodlyrottenapple (changes requested):

.github/workflows/test-pr.yml
tothtamas28: That did the trick, thanks!

.github/workflows/test-pr.yml
tothtamas28: Done.

PR #157, opened 2023-11-03
Update master-push.yml

F-WRunTime:
This workflow keeps failing on release. 
This might be a fix, we use this as part of the publish methods in avm-semantics / K and there has been no issue with losing the key signing key in the middle of a push  Example here: https://github.com/runtimeverification/k/blob/13a89f4c6a5c5ed2780ce1cffb6bd62bcfbc6be7/.github/workflows/test-pr.yml#L253C1-L264C1


F-WRunTime:
Maybe there is something odd that action is doing, but this could be worth a try to properly get this fixed.. Disregard if the problem exists for another known reason.

tothtamas28:
Based on the build logs, to me the problem seems to be a segfault in this step:

```
cachix push k-framework
```

Can you log into the runner and try the steps manually? Maybe then we'll know more.

F-WRunTime:
@tothtamas28 I have logged into the runner. The problem is that the signing key is dropped in the middle of the actions execution push. I have not seen this happen / occur using push locally on CLI.

F-WRunTime:
I think another way I can test this will be to create a new binary cache and run the release here in the PR. I will go ahead with this to reproduce and demonstrate the difference.

PR #159, opened 2023-11-07
mark freshUInt and freshBool as preserving definedness

jberthold:
Use of `#bufStrict` is safe in these rules because the argument range is explicitly restricted (and the argument is symbolic anyway)


jberthold:
This is a comparison of the `kontrol` branch with master (from last Friday).
image

anvacaru (approved):

PR #162, opened 2023-11-08
Add test with high branching factor

lucasmt:
Adds a test with a high branching factor in order to test new features like node merging and branch parallelization.

The test splits into 8 branches in total (and this can be increased by increasing the length of the `proof` array). It takes 1h to run on my machine.


ehildenb:
We should add an actual test to the testing harness that merges the nodes and makes assertions about them. For example, it should be added to this test harness that the node merging is happening correctly: https://github.com/runtimeverification/kontrol/blob/master/src/tests/integration/test_foundry_prove.py#L236

lucasmt:
@ehildenb Currently I don't think it's feasible to apply node merging to this example, because of this issue: https://github.com/runtimeverification/kontrol/issues/194.

palinatolmach:
@nwatson22 brought up another challenge in adding this example to our test suite: the merge node functionality requires the user to accurately specify nodes to be merged, and it might be nontrivial to identify and add such nodes to the kcfg in this example. This should be partially addressed in https://github.com/runtimeverification/kontrol/pull/163, but it requires additional changes and is not being worked on at the moment.

Issue #23, opened 2023-08-24
Usability improvements to `foundry-merge-node`


lucasmt:
With [this PR](https://github.com/runtimeverification/evm-semantics/pull/1934), we can now use the `foundry-merge-nodes` command to merge two nodes in different branches into a single node that subsumes both. This lets us merge branches back after a split, for example after an if statement or a loop, continuing the execution as a single branch, which has the potential of greatly improving the scalability of the tool in code with many loops or if statements.

However, this command has a few limitations in terms of usability:

1. The corresponding nodes in each branch (same PC) need to be present in the KCFG in order to be merged. This is not necessarily easy to ensure.
2. Because it requires manual intervention, the result of the verification becomes harder to reproduce.

We should think about ways to address these limitations to make this functionality easier to use. For example, suppose there was a cheatcode `kevm.mergePoint()` that denotes the point where we want to merge branches:
```
if (condition) {
    // if branch
} else {
    // else branch
}

kevm.mergePoint();
```
If we could implement this cheatcode such that the point where it's called is always added as a node to the KCFG, this would take care of (1). Going one step further, if we had an option to automatically merge such nodes if they have the same PC, this would take care of (2) (but this heuristic might not always work, for example if the merge point itself is inside a loop or a function that is called multiple times). But I don't know if something like that could actually be implemented, because it is a cheatcode that would not be handled solely from inside `kevm`, instead needing to somehow communicate with `pyk`.

I'd like to hear people's thoughts on this and brainstorm possible options.

lucasmt:
For now, the alternative that I'm using to `mergePoint` is to make an external call that does nothing (it can be implemented as a dummy cheatcode, for example) and turn on `--break-on-calls`. But this will break on every external call, which will slow down execution a bit.

ehildenb:
I would call it a `kevm.cut()` instead, in the same notion of a `cut-rule`, and we can just make the rule which implements it a cut-rule. This is in line with the idea of cut-points from @daejunpark's thesis: https://www.semanticscholar.org/paper/Cut-Bisimulation-and-Program-Equivalence-Park-Kasampalis/be608a0dcfe23747c4146ed0f10a544ab015e059

lucasmt:
One potential complication that just occurred to me based on the current engagement: the point where we ideally want to merge can be inside the client code, in which case calling a cheatcode is not really possible.

palinatolmach:
[...]

PR #163, opened 2023-11-08
Add support for cut cheat code

anvacaru:
Fixes: #23 

Added a new cut cheat code that takes a program counter and will create a new KCFG node whenever that program counter is executed.

- When the cut is executed in a proof/test, the program counter argument is added to a set in the <cutPC> cell.
- The #pc [ OP ] rule is overwritten to identify when a program count in the <cutPC> cell is executed.
- Added an empty foundry.cut rule and marked it as a cut-point rule to create the new node in the kcfg.

TODO:

- [ ] add test
- [ ] extract `cut` cheat code in the cheat code library after this PR is merged

### DIFF
--- a/src/kontrol/kdist/foundry.md
+++ b/src/kontrol/kdist/foundry.md
@@ -119,6 +119,8 @@ The configuration of the Foundry Cheat Codes is defined as follwing:
     - `<isStorageWhitelistActive>` flags if the whitelist mode is enabled for storage changes.
     - `<addressSet>` - stores the address whitelist.
     - `<storageSlotSet>` - stores the storage whitelist containing pairs of addresses and storage indexes.
+6. The `<cutPC>` cell stores a set of program counters inserted using the `cut` cheat code.
+Each program counter in the set will end up creating a new node in the KCFG.
 
 ```k
 module FOUNDRY-CHEAT-CODES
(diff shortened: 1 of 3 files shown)


lucasmt:
I tried the `cut` cheatcode at the example from PR #162, but it didn't seem to be working. I used `kontrol section-edge` to look closer at what's going on, and I found that it was just skipping from `#next` to `#execute` without going through `#pc` (which is what should trigger the `#cut` instruction):
```
┃  │  (1 step)
┃  ├─ 122
┃  │   k: #next [ PUSH ( 2 ) ] ~> #execute ~> CONTINUATION:K
┃  │   pc: 4690
┃  │   callDepth: 0
┃  │
┃  │  (1 step)
┃  ├─ 123
┃  │   k: #execute ~> CONTINUATION:K
┃  │   pc: 4693
┃  │   callDepth: 0
```
I think this is because of the optimizations in the [optimizations.md](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/optimizations.md?plain=1#L58) file. Not sure what's the best way to get around this.

lucasmt:
Additionally, I think the cheatcode also won't work if the PC is arrived at from a `JUMP` instruction. Although this might not be as big of a problem since one can just use the PC immediately following the `JUMPDEST`.

anvacaru:
Closing this for now.

Issue #133, opened 2023-10-26
`<number>` not carried over from `setUp` to test


palinatolmach:
The value of `<number>` (corresponding to `block.number` in Solidity) set in `setUp()` via `vm.roll(...)` is not carried over to the test.  E.g., `test_roll_setup` fails:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity >=0.6.2 <0.9.0;

import "forge-std/Test.sol";
import "forge-std/Vm.sol";

import "../test/examples/MyErc20.sol";

contract RollTest is Test {
    function setUp() external {
        vm.roll(123);
    }

    // function testFail_run1() external {
    function test_roll_setup() external {
        assert(block.number == 123);

    }
}
```  
with the following path condition:
```
#And #Not ( { NUMBER_CELL:Int #Equals 123 } ) )
```

In the test configuration, `<number>` has symbolic value:
```
  <number>
    NUMBER_CELL:Int
  </number>
```
while in the terminal node of `setUp` the `<number>` value is set to the one provided in `vm.roll`:
```
  <number>
    123
  </number>
```

nwatson22:
Right now we only save the `<accounts>` cell. Are there any other cells that it would also make sense to keep, e.g. in `<block>`? Does foundry have any sort of guarantee about what the block number would be for an example like this? I guess it would make sense for it not to be reset to 0, but could it possibly be higher than 123 when `test_roll_setup()` is executed?

PR #166, opened 2023-11-09
Save number cell from setUp function

anvacaru:
Fixes: #133


src/tests/integration/test-data/foundry/test/BlockParamsTest.t.sol
```diff
@@ -36,3 +36,15 @@ contract BlockParamsTest is Test {
         assert(x >= 0);
     }
 }
+
+contract RollTest is Test {
+    function setUp() external {
+        vm.roll(123);
+    }
+
+    // function testFail_run1() external {
```
palinatolmach, line 45: Sorry, that came from the contract I was running — we probably don't need it here :)

anvacaru, line 45:
```suggestion
```

anvacaru: Done. I need to update the expected output once more.

palinatolmach:
Created a separate issue for tracking other similar cells as a follow-up to this PR: https://github.com/runtimeverification/kontrol/issues/169.

palinatolmach (approved):

PR #179, opened 2023-11-15
Move expectRevert check outside of `CALL*`/`CREATE*`logic

### LINKED ISSUE #176: Refactor `expectRevert` so that it occurs after external function call completes
In the context of external function calls, the current implementation of `expectRevert` performs its checks before the call is completed. This negatively impacts the ability to write compositional functions specifications, which need to capture the `expectRevert` cell, and will effectively double the number of specifications created. Since Foundry is a layer on top of KEVM, I would expect this particular check to occur only when the call has been completed.

[Kontrol documentation](https://github.com/runtimeverification/kontrol/blob/ab04b75350ac0fa47a0b9f733d5760b74525a935/src/kontrol/kdist/foundry.md?plain=1#L540) states that `The `#halt` production is used to examine the end of each call in KEVM.`, but that is not how external calls end. The complete external call is captured by `CALL _ _ _ _ _ _ ~> .`.

CAVEAT: I am not certain what the correct approach would be with respect to internal function calls.

anvacaru:
Fixes #176
**TL;DR** Changes:
 - use `#next[OP:Opcode]` production to insert a revert check when a revert is expected.
 - now that we check for reverts after a call has finished, we have to:
     - (1) update the local memory at the location where the output is stored to remove the error output (only for `CALL*` opcodes).
     - (2) update the word stack to change the value pushed by revert.
- fixed a bug in which `expectRevert` would incorrectly trigger at a call depth deeper than where it is invoked instead of only at the same level as the cheat code invocation.

**Description**:
 The rules used to check if a `CALL`/`CREATE` opcode has reverted or not are introduced before the call execution has finished, resulting in duplicated rules in a functions summary based on whether `<isRevertExpected>` is `true` or `false`. 
For context, this is how we currently check for expected reverts:
```k
    rule [foundry.handleExpectRevert]:
         <k> (. => #checkRevertReason ~> #clearExpectRevert) ~> #halt ... </k>
         <statusCode> SC </statusCode>
         <callDepth> CD </callDepth>
         <expectedRevert>
           <expectedDepth> ED </expectedDepth>
           ...
         </expectedRevert>
      requires CD <=Int ED
       andBool SC =/=K EVMC_SUCCESS
      [priority(40)]
```

We instead use the `#next[OP:Opcode]` to insert a check for revert after the execution of a call if a revert is expected and if `OP` is a `CALL`/`CREATE` opCode.
The drawback is that we can no longer `expectRevert` for internal functions with these changes.
This way, we can generate a function summary with `expectRevert` set to `false`.


src/kontrol/kdist/foundry.md
```diff
@@ -537,37 +537,44 @@ All cheat code calls which take place while `expectRevert` is active are ignored
...
+         <wordStack> _ : _ : _ : _ : _ : RETSTART : RETWIDTH : _WS </wordStack>
          <expectedRevert>
            <isRevertExpected> true </isRevertExpected>
            <expectedDepth> ED </expectedDepth>
            ...
          </expectedRevert>
-      requires CD <=Int ED
-       andBool SC =/=K EVMC_SUCCESS
+       requires CD <=Int ED
+       andBool (OP ==K CALL orBool OP ==K CALLCODE)
```
anvacaru, line 554: This might be redundant since `CallOp` contains only `CALL` and `CALLCODE`.

src/kontrol/kdist/foundry.md
```diff
@@ -537,37 +537,44 @@ All cheat code calls which take place while `expectRevert` is active are ignored
...
+    rule <k> #next [ OP:CallSixOp ] ~> (. => #checkRevert ~> #updateRevertOutput RETSTART RETWIDTH) ~> #execute ... </k>
+         <callDepth> CD </callDepth>
+         <wordStack> _ : _ : _ : _ : RETSTART : RETWIDTH : _WS </wordStack>
+         <expectedRevert>
+           <isRevertExpected> true </isRevertExpected>
+           <expectedDepth> ED </expectedDepth>
+           ...
+         </expectedRevert>
+       requires CD <=Int ED
+       andBool (OP ==K DELEGATECALL orBool OP ==K STATICCALL)
```
anvacaru, line 566: Same thing for `CallSixOp`.

src/kontrol/kdist/foundry.md
anvacaru, line 554: This might be redundant since CallOp contains only CALL and CALLCODE.
```suggestion
```

src/kontrol/kdist/foundry.md
anvacaru, line 566: Same thing goes for `CallSixOp`.
```suggestion
```

PetarMax:
A major benefit of this approach is that summaries of Solidity functions become decoupled from the Foundry layer. Currently, these are intertwined, resulting in multiple summaries where there ostensibly should be one.

When it comes to internal functions, this becomes a question of use cases. In real-world scenarios, would one expect to check for `expectRevert` of an internal function? What would that look like?

src/kontrol/kdist/foundry.md
```diff
@@ -537,37 +537,42 @@ All cheat code calls which take place while `expectRevert` is active are ignored
       [priority(35)]
 ```
 
-The `#halt` production is used to examine the end of each call in KEVM.
-If the call depth of the current call is lower than the call depth of the `expectRevert` cheat code and the `<statusCode>` is not `EVMC_SUCCESS`, the `#checkRevertReason` will be used to compare the output of the call with the expect reason provided.
+We use the `#next[OP]` to identify OpCodes that can revert and insert a `#checkRevert` production used to examine the end of each call/create in KEVM.
+We also expect the call depth of the current call to be lower than the call depth of the `expectRevert`.
```
palinatolmach, line 541: Should it be "lower than or equal"?
```
requires CD <=Int ED
```

```suggestion
We also expect the call depth of the current call to be lower than or equal to the call depth of the `expectRevert`.
```

palinatolmach:
> When it comes to internal functions, this becomes a question of use cases. In real-world scenarios, would one expect to check for `expectRevert` of an internal function? What would that look like?

I don't think it's very common to call other functions within a test contract, but that might happen; internal calls within a test might also occur if the test calls a function of contract it's inherited from. A workaround to make an internal call `call_assert_false()` external is to prepend this call with `this.` (i.e., `this.call_assert_false()`) — Andrei tests it with one of the tests he added: https://github.com/runtimeverification/kontrol/blob/cecb785d5aa792a2b853c7f4208e5b8c0886a9b4/src/tests/integration/test-data/foundry/test/Simple.t.sol#L49

IIUC, Foundry supports `expectRevert` for internal functions, but it uses `JUMP` to detect them, so it doesn't really distinguish between conditional statements and function calls.

src/kontrol/kdist/foundry.md
anvacaru: It should be only `==`. @PetarMax suggested that I should write a test to check this behaviour:
 - expectRevert set with an expected message at depth 0.
 - there is a revert at depth 2 with some non-expected message which will be caught at depth 1
 - there is a revert at depth 1 with the expected message.

So I created a test like this that passes with Foundry
```solidity
pragma solidity =0.8.13;

import "forge-std/Test.sol";

contract TestExpectRevert is Test {
    function revertDepth2() public {
        revert ("This should be at depth 2");
    }
    function revertDepth1() public  {
        try this.revertDepth2()
        {} catch {}
        revert ("This should be at depth 1");
    }

    function test_expectRevert_inDepth() public {
        vm.expectRevert("This should be at depth 1");
        this.revertDepth1();
    }
}
```
image

Now, due to the `ED <=Int CD` (expected depth leq than the current call depth), the expect revert check would trigger before the `catch` in the bytecode and depth 1 and fail the `expectRevert`.
The point in having `ED ==Int CD` is that the expect revert check would only trigger at the same level as the call to the `expectRevert` cheat code invocation, which should be the foundry test.

palinatolmach (approved):

src/kontrol/kdist/foundry.md
palinatolmach: Thanks for the clarification! LGTM

Issue #93, opened 2023-10-10
`kontrol list` crshes when there are no proofs


geo2a:
The `kontrol list` command crashes with an ugly Python stack trace when there are no proofs. Perhaps it would be better to catch the `FileNotFoundError` exception and print a friendlier error message.

**Observed behavior**
```
$ kontrol list
Traceback (most recent call last):
  File "/nix/store/zg6kd4c083df1vigcj2nlasvnd8vc003-python3.10-kontrol-0.1.20/bin/..kontrol-wrapped-wrapped", line 9, in <module>
    sys.exit(main())
  File "/nix/store/zg6kd4c083df1vigcj2nlasvnd8vc003-python3.10-kontrol-0.1.20/lib/python3.10/site-packages/kontrol/__main__.py", line 69, in main
    execute(**vars(args))
  File "/nix/store/zg6kd4c083df1vigcj2nlasvnd8vc003-python3.10-kontrol-0.1.20/lib/python3.10/site-packages/kontrol/__main__.py", line 254, in exec_list
    stats = foundry_list(foundry_root=foundry_root)
  File "/nix/store/zg6kd4c083df1vigcj2nlasvnd8vc003-python3.10-kontrol-0.1.20/lib/python3.10/site-packages/kontrol/foundry.py", line 556, in foundry_list
    for test_id in listdir(foundry.proofs_dir):
FileNotFoundError: [Errno 2] No such file or directory: 'out/proofs'
```
**Desired behavior**
```
$ kontrol list
No proofs found, run kontrol prove to start proving
```

PR #204, opened 2023-11-24
Error handling outside of a foundry project

anvacaru:
Fixes: https://github.com/runtimeverification/kontrol/issues/93

Gracefully ends execution when `kontrol` is used outside of a Foundry project.


src/kontrol/foundry.py
```diff
@@ -58,9 +58,12 @@ def __init__(
...
         self._root = foundry_root
-        with (foundry_root / 'foundry.toml').open('rb') as f:
-            self._toml = tomlkit.load(f)
-        self._bug_report = bug_report
+        try:
+            with (foundry_root / 'foundry.toml').open('rb') as f:
+                self._toml = tomlkit.load(f)
+        except FileNotFoundError:
+            _LOGGER.error('No Foundry project found. Exiting.')
+            sys.exit(1)
```
tothtamas28, line 66: This exception should be handled by the command line layer. Otherwise, this would force a library user to handle `SystemExit`.

tothtamas28 (changes requested):

src/kontrol/foundry.py
anvacaru: @tothtamas28 Then in this particular case I should surround with a try except the foundry-list call in 

```python
def exec_list(foundry_root: Path, **kwargs: Any) -> None:
    stats = foundry_list(foundry_root=foundry_root)
    print('\n'.join(stats))
```

But this should apply for all `exec` options of kontrol like `prove` and `build`.

tothtamas28: This indicates that refactoring is necessary before introducing this change.

* Methods like `foundry_list` should take `foundry: Foundry`.
* You can add a helper in `__cli__` that attempts to load a `Foundry` and exists with an error message if `foundry.toml` is not found.

anvacaru: Done! Thank you for the suggestion!

src/tests/profiling/test_foundry_prove.py
```diff
@@ -28,15 +29,15 @@
 
 def test_foundy_prove(profile: Profiler, use_booster: bool, bug_report: BugReport | None, tmp_path: Path) -> None:
     foundry_root = tmp_path / 'foundry'
-
+    foundry = load_foundry(foundry_root=foundry_root)
```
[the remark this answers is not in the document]
anvacaru: Oh, my bad. I'll switch the two instructions.

src/kontrol/foundry.py
```diff
@@ -835,3 +825,12 @@ def foundry_node_printer(foundry: Foundry, contract_name: str, proof: APRProof)
     if type(proof) is APRProof:
         return FoundryAPRNodePrinter(foundry, contract_name, proof)
     raise ValueError(f'Cannot build NodePrinter for proof type: {type(proof)}')
+
+
+def load_foundry(foundry_root: Path, bug_report: BugReport | None = None) -> Foundry:
```
tothtamas28, line 830: I think this belong to `__main__`, printing to `stderr` and exiting with an error code is only for command line use.

src/kontrol/foundry.py
```diff
@@ -835,3 +825,12 @@ def foundry_node_printer(foundry: Foundry, contract_name: str, proof: APRProof)
...
+    try:
+        foundry = Foundry(foundry_root=foundry_root, bug_report=bug_report)
+    except FileNotFoundError:
+        _LOGGER.error('foundry.toml file not found. Are you running kontrol in a Foundry project?')
```
tothtamas28, line 834:
```suggestion
        print(f'File foundry.toml not found in: {foundry_root}. Are you running kontrol in a Foundry project?', file=sys.stderr)
```

src/tests/integration/test_foundry_prove.py
```diff
@@ -66,7 +73,7 @@ def foundry_root(tmp_path_factory: TempPathFactory, worker_id: str) -> Path:
             run_process(['forge', 'build'], cwd=foundry_root)
 
             foundry_kompile(
-                foundry_root=foundry_root,
+                foundry=load_foundry(foundry_root),
```
tothtamas28, line 76: This is fine here:

```suggestion
                foundry=Foundry(foundry_root),
```

src/tests/profiling/test_foundry_prove.py
```diff
@@ -28,15 +29,15 @@
 
 def test_foundy_prove(profile: Profiler, use_booster: bool, bug_report: BugReport | None, tmp_path: Path) -> None:
     foundry_root = tmp_path / 'foundry'
-
     _forge_build(foundry_root)
+    foundry = load_foundry(foundry_root=foundry_root)
```
tothtamas28, line 33: Consider returning `Foundry` from `_forge_build`. It's fine to instantiate it with the constructor directly there.

src/kontrol/foundry.py
anvacaru: Moving this in `__main__` would prevent me from importing the function in the test files.

anvacaru: But I guess this function is not needed in there.

src/kontrol/foundry.py
anvacaru: Why `print()` and not `_LOGGER.error()`?

tothtamas28: `logging.error()` is for indicating suppression of an exception, `print` is for normal output: https://docs.python.org/3/howto/logging.html#when-to-use-logging

anvacaru: Thank you!

src/tests/integration/test_foundry_prove.py
anvacaru: Done!

src/tests/profiling/test_foundry_prove.py
anvacaru: Done!

src/kontrol/__main__.py
```diff
@@ -238,7 +250,7 @@ def exec_prove(
     )
 
     results = foundry_prove(
-        foundry_root=foundry_root,
+        foundry=(_load_foundry(foundry_root, bug_report)),
```
tothtamas28, line 253:
```suggestion
        foundry=_load_foundry(foundry_root, bug_report),
```

src/tests/integration/test_foundry_prove.py
```diff
@@ -66,28 +65,28 @@ def foundry_root(tmp_path_factory: TempPathFactory, worker_id: str) -> Path:
...
 
 
-def test_foundry_kompile(foundry_root: Path, update_expected_output: bool, use_booster: bool) -> None:
+def test_foundry_kompile(foundry: Foundry, update_expected_output: bool, use_booster: bool) -> None:
     if use_booster:
         return
     # Then
     assert_or_update_k_output(
-        foundry_root / 'out/kompiled/foundry.k',
+        foundry._root / 'out/kompiled/foundry.k',
```
tothtamas28, line 84:
```suggestion
        foundry.main_file,
```

src/tests/integration/test_foundry_prove.py
```diff
@@ -66,28 +65,28 @@ def foundry_root(tmp_path_factory: TempPathFactory, worker_id: str) -> Path:
...
         TEST_DATA_DIR / 'foundry.k.expected',
         update=update_expected_output,
     )
     assert_or_update_k_output(
-        foundry_root / 'out/kompiled/contracts.k',
+        foundry._root / 'out/kompiled/contracts.k',
```
tothtamas28, line 89:
```suggestion
        foundry.kompiled / 'contracts.k',
```

But introducing a new property `contracts_file` would be even better.

tothtamas28 (approved):
Left a few comments. Other than that looks nice and clean.

src/tests/integration/test_foundry_prove.py
anvacaru: Done!

src/kontrol/__main__.py
anvacaru: Done!

anvacaru:
**TODO**: `FoundryMock()` has to be refactored for `test_foundry_list`.

PR #211, opened 2023-11-28
[DRAFT] Implement no-gas mode

anvacaru:
Needs more testing.

Taking the `test_ExpectRevert_increasedDepth` as the first benchmark.
 - kontrol (master): `10m 27s`
 - kontrol (master) with `--auto-abstract-gas`: `10m 06s`
 - kontrol (nogas branch) with `--no-gas` option: `7m 02s`
 - kontrol (nogas branch) without `--no-gas` option: `11m 54s`
 - kontrol (with kevm nogas) with `--no-gas` option: `7m 48s`
 - kontrol (with kevm nogas) withouth `--no-gas` option: `10m 49s`

32.70% speed up with `--no-gas` than `master`.
13.88% slow down without `--no-gas` than `master`.
These numbers will increase with longer execution.


PetarMax:
I think It would be a good idea to create a token, e.g., `IMPERMISSIBLE`, and then have the functions that we would like to ensure are not executed, such as `#gasExec`, `#refund`, and `#deductGas`, transition to it unconditionally, in addition to disabling all of their usage. This would allows us to easily track down any missed cases.

src/kontrol/kdist/foundry.md
```diff
@@ -25,7 +25,363 @@ module FOUNDRY
...
+      <callState>
+        <wordStack>
+          ( WS => 0 : WS )
+        </wordStack>
+        <pc>
+          ( PCOUNT => ( PCOUNT +Int 1 ) )
+        </pc>
+        ...
+      </callState>
+      <noGas> true </noGas>
```
ehildenb, line 259: I also wouldn't just adjust these rules only.

anvacaru:
@ehildenb drafted this PR in KEVM with only the change for `#gas [ _ , _ ]`. I have to test it in kontrol and then I will reply with updates on the speedup.

anvacaru:
Overriding `#gas [ _. _]` and `#deductGas` in kevm led to a time of `7m48s` vs `7m02` (with all rules removed).

PetarMax:
> This seems like more rules to duplicate than needed. Could we just override the rules for either:
> 
>     * `#gas [ _ , _ ]` https://github.com/runtimeverification/evm-semantics/blob/767b866d5463d143d5864934e64e397d902f17f1/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md?plain=1#L1847
> 
> 
> Or we can override:
> 
>     * `#refund`: https://github.com/runtimeverification/evm-semantics/blob/767b866d5463d143d5864934e64e397d902f17f1/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md?plain=1#L1421
> 
>     * `#deductGas` and `#deductMemoryGas`: https://github.com/runtimeverification/evm-semantics/blob/767b866d5463d143d5864934e64e397d902f17f1/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md?plain=1#L1869
> 
> 
> Do we still get the same speedup? This should at least prevent building up large gas terms, and doing any gas comparisons, which I guess is a big chunk of the time.

Everything in this PR was done systematically to ensure that  the `<gas>`, `<callGas>`, and `<refund>` cells are never inspected or modified, thereby completely disabling gas reasoning. The introduction of the `IMPERMISSIBLE` token is there to ensure that the functions that we want not to be executed actually are not executed.

Which rules specifically seem duplicated, but need not be?

anvacaru:
Closing this as it is no longer needed. The no-gas logic is moved in kevm in https://github.com/runtimeverification/evm-semantics/pull/2207 . I will open a new PR to incorporate these changes in kontrol once we update the kevm release to the one using the no-gas logic.

PR #220, opened 2023-11-30
Add `pytest` option `foundry-root`

### LINKED ISSUE #214: Add `pytest` option that turns off kompilation during a test run
As discussed with @tothtamas28, we should add a `pytest` option that allows dropping the kompilation step when running Kontrol tests and using pre-kompiled files instead. That is needed for performance profiling performed by the Haskell Backend team, as recompiling introduces noise into the performance data.

Ideally, we want to introduce it to KEVM too, where the setup is more complicated.

tothtamas28:
Fixes #214


tothtamas28:
One way to use this feature:

```bash
# Initialize and kompile test project
$ make test-integration TEST_ARGS='-n0 -k test_foundry_kompile'
$ cp -r /tmp/pytest-of-user/pytest-current/foundry .
# This should be fast now
$ make test-integration TEST_ARGS='-n0 -k test_foundry_kompile --foundry-root foundry'
# Run tests with the pre-kompiled project
$ make test-integration TEST_ARGS='--foundry-root foundry'
```

PR #221, opened 2023-12-01
Deployment summaries

nwatson22:
Adds command `foundry summary` which takes in a state diff object (see https://book.getfoundry.sh/cheatcodes/stop-and-return-state-diff) in a json format, and generates a contract with a function `recreateDeployment` which reproduces the deployment.

Checked to work properly with the following versions of:
- [Foundry](https://github.com/foundry-rs/foundry/commit/e972bf70412e51b1e15041341f7c29725bb6ac17)
- [forge-std](https://github.com/foundry-rs/forge-std/commit/80a8f6ea9362849b2a8f2dc28df40c77a64f9c16)


src/tests/integration/test-data/deployment-code.expected
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.8.13;
+
+import { Vm } from "forge-std/Vm.sol";
+
+contract DeploymentSummary1Code {
```
JuanCoRo, line 5: Where does the `1` come from?

src/tests/integration/test-data/deployment-main.expected
```diff
@@ -0,0 +1,23 @@
+pragma solidity ^0.8.13;
+
+import { Vm } from "forge-std/Vm.sol";
+
+import 'src/DeploymentSummary1Code.sol';
```
JuanCoRo, line 5:
```suggestion
import { DeploymentSummary1Code } from 'src/DeploymentSummary1Code.sol';
```
This should be as suggested to have a consistent style

src/tests/integration/test-data/deployment-main.expected
```diff
@@ -0,0 +1,23 @@
...
+
+contract DeploymentSummary1 is DeploymentSummary1Code {
```
JuanCoRo, line 7: Same question as above with the `1`

src/kontrol/deployment.py
```diff
@@ -0,0 +1,126 @@
...
+        lines = []
+        lines.append(f'pragma solidity {self.SOLIDITY_VERSION};\n')
+        lines.append('import { Vm } from "forge-std/Vm.sol";\n')
+        return lines
+
+    def generate_code_contract(self) -> list[str]:
+        lines = []
+        lines.append(f'contract {self.name}Code ' + '{')
+        for code_alias, code in self.code.items():
+            lines.append(f'\tbytes constant public {code_alias}Code = hex{code!r};')
```
JuanCoRo, line 35:
```suggestion
            lines.append(f'\tbytes constant internal {code_alias}Code = hex{code!r};')
```
No need to create functions for the code

src/tests/integration/test-data/deployment-code.expected
nwatson22: I wanted to have some way of showing it was an example and not built-in, but I see why that's confusing, I've renamed it to simply `DeploymentSummary`.

src/tests/integration/test-data/deployment-main.expected
nwatson22: Fixed.

src/kontrol/deployment.py
nwatson22: Fixed.

JuanCoRo (approved):
LGTM! :+1:

src/kontrol/deployment.py
```diff
@@ -0,0 +1,123 @@
...
+        return lines
+
+    def generate_main_contract_file(self) -> list[str]:
+        lines = self.generate_header()
+        lines.append('import { ' + self.name + 'Code } from "./' + self.name + 'Code.sol";')
+        lines += self.generate_main_contract()
+        return lines
+
+    def generate_code_contract_file(self) -> list[str]:
+        lines = self.generate_header()
```
JuanCoRo, line 74: This will add `import { Vm } from "forge-std/Vm.sol";` to the code contract, but it's actually not needed. Better remove it

nwatson22: Fixed.

PR #226, opened 2023-12-04
Eliminate duplicate rule variable from `foundry.prank`

geo2a:
The duplicate rule variable triggers a unification bug in Booster (see https://github.com/runtimeverification/hs-backend-booster/issues/406) which causes it to not apply the rule. If we change the rule a little, by moving the constraint on the call depth into the side condition, then Booster aborts and reverts to Kore, which branches as expected.

We will work on Booster's rule application. Until then, I suggest we change the rule in Kontrol.


src/kontrol/kdist/cheatcodes.md
```diff
@@ -633,11 +633,12 @@ This is needed in order to prevent overwriting the caller for subcalls.
          <prank>
            <active> true </active>
            <newCaller> NCL </newCaller>
-           <depth> CD </depth>
+           <depth> D </depth>
```
anvacaru, line 636: There are a few other places where we duplicate variable names. Perhaps we should update those as well?

PR #227, opened 2023-12-05
Write `KompiledKore` data structures on `foundry_kompile`

tothtamas28:
Writes `pyk-definition.kore.json` on `foundry_kompile`. `KompiledKore` is loaded from there for KAST-to-KORE transformation during `foundry_prove`.

Related: https://github.com/runtimeverification/pyk/pull/740

Profiling results:

|             | `kompile` | `prove` |
|-------------|-----------|---------|
| Old version |       88s |    252s |
| New version |      109s |    235s |
| Delta       |      +21s |    -17s |


ehildenb:
Would be good to:

- indicate that it's Kore level information (maybe `*.kore.json`)
- Include versioning information (ideally in th efile name, so that we don't have to open the file to check the version), so something like `pyk-definition.HASH_OF_KORE_FILE.kore.json`.

tothtamas28:
> Would be good to:
>  * indicate that it's Kore level information (maybe `*.kore.json`)
>  * Include versioning information (...)

https://github.com/runtimeverification/pyk/pull/760

`KompiledKore.load` compares the timestamps now.

anvacaru (approved):

PR #231, opened 2023-12-06
[Draft] Init state for a function in CSE

anvacaru:
Changes:
 - add `--init-cse` flag to `kontrol prove` to initialize the kcfg of a function for CSE according to [this document](https://docs.google.com/document/d/1V8O2cX-akZXA9UGqnevdwCsZh4yyDTLu9-Ccu8cjkDU/edit?not_in_iframe=true&pli=1#heading=h.1vw8cs43in0b)
 - add a new method `_init_cse_cterm` that creates the init node in the kcfg with:
   - two symbolic accounts `ACCT_FROM` as caller  and `ACCT_TO` as callee
   - the contract bytecode as `ACCT_TO` code.
   - initializes the `<k>` cell with a production `CALL _GCAP ACCT_TO VALUE ARGSTART ARGWIDTH RETSTART RETWIDTH`, where `ARGWIDTH` and `RETWIDTH` are calculated based on the function abi.
 - updated the `Contract.Constructor` and `Contract.Method` class to implement `argwidth`  and `retwidth` properties.


src/kontrol/prove.py
```diff
@@ -494,6 +513,139 @@ def _init_cterm(
...
+    setup_cterm: CTerm | None = None,
+    calldata: KInner | None = None,
+    callvalue: KInner | None = None,
+) -> CTerm:
+    contract = KEVM.account_cell(
+        KVariable('CONTRACT'),
+        KVariable('CONTRACT_BAL'),
+        program,
+        KVariable('CONTRACT_STORAGE'),
+        KVariable('CONTRACT_STORAGE'),
```
ehildenb, line 532: It looks like we're duplicating a bunch of code, I wonder if we could reuse the `init_cterm` and just pass in the flag instead? Also, can we just make it so that whenever you call a non `test_*` function, this is the default?There is no other use-case of `init_cterm` on a non-`test_*` function _except_ for CSE, so we should just make that function do this whne this flag is passe.d

src/kontrol/prove.py
```diff
@@ -494,6 +513,139 @@ def _init_cterm(
...
+                        KApply(
+                            '#range(_,_,_)_EVM-TYPES_Bytes_Bytes_Int_Int',
+                            [KVariable('LM'), KVariable('ARGSTART'), intToken(4)],
+                        ),
+                    ),
+                    KEVM.abi_selector(method.signature),
+                ],
+            )
+        ),
+    ]
```
ehildenb, line 604: Should these constraints be added for _any_ proof? They seem prettygeneric. Maybe it can be added to `KEVM.add_constraints` instead?

ehildenb: This is where that `KEVM.add_invariant` method is, which does exactly that: https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kevm.py#L241

PR #243, opened 2023-12-12
Update README.md with Link to Kontrol Docs

reekamaharaj:


README.md
```diff
@@ -1,5 +1,8 @@
 # Kontrol
```
palinatolmach, line 2: @reekamaharaj can I ask you to add a description of Kontrol here please? I think we can just use the one we have in the documentation: "[Kontrol](https://github.com/runtimeverification/kontrol) combines KEVM and [Foundry](https://book.getfoundry.sh/) to grant developers the ability to perform [formal verification](https://en.wikipedia.org/wiki/Formal_verification) without learning a new language or tool. This is especially useful for those who are not verification engineers. Additionally, developers can leverage Foundry test suites they have already developed and use symbolic execution to increase the level of confidence."

palinatolmach (approved):

Issue #247, opened 2023-12-14
Failed proof isn't reported if function doesn't start with `test` or `testFail`


palinatolmach:
Update: the reason for the proof not being reported as failed is the fact that the function we're proving starts with `check`, and we only add a constraint on `foundry_success` predicate if the function name is prefixed with `test` or `fail`:

[prove.py#404](https://github.com/runtimeverification/kontrol/blob/921005c7a103aea50b93d568715d587590ba68ce/src/kontrol/prove.py#L404C1-L406C1):
```
    is_test = method.signature.startswith('test')
    failing = method.signature.startswith('testFail')
```

[prove.py#498](https://github.com/runtimeverification/kontrol/blob/921005c7a103aea50b93d568715d587590ba68ce/src/kontrol/prove.py#L498C1-L499C1):

```
    if is_test:
        if not failing:
            return final_cterm.add_constraint(mlEqualsTrue(foundry_success))
        else:
            return final_cterm.add_constraint(mlEqualsTrue(notBool(foundry_success)))
```

However, functions with other prefixes are still getting analyzed and reported as passing, which is misleading. I see at least two ways to address it: 
- throw an error if the function to be proven doesn't start with `setUp`, `test`, or `fail` — we can do that at the `--match-test` processing stage
- add a warning message if the function doesn't start with `test` or `fail`
- **use a different success predicate for functions not starting with `test` and `testFail`, e.g., assert that there is no reverting path**

 
Original issue:

The [following test](https://github.com/Cyfrin/sc-exploits-minimized/blob/d84d6f1cd73fad2c6dd49ca034f25a1e4bacbf83/src/invariant-break/FormalVerificationCatches.sol#L40) should fail if `number` equals `99`. The failing node has an `EVMC_REVERT` status, but it doesn't get reported as failing and the proof is considered passing. At the same time, `kontrol get-model` generates a counterexample for this node correctly.

palinatolmach:
As suggested by @RaoulSchaffranek, a nice way to address this issue is to introduce a different success predicate for functions not starting with `test` and `testFail`, e.g., assert that there is no reverting path.

PR #249, opened 2023-12-15
Verify `proveX` and `checkX` functions similarly to `testX`

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/247.

~Currently, this PR adds an error that is thrown if any of the function names provided with `--match-test` do not start with `test` or `testFail`, but should be refactored.~

At the moment, only functions starting with `test` and `testFail` are checked for Foundry success predicate and will fail if the function reverts. There are two issues with this approach:
- tests prefixed with `test`/`testFail` are picked up for fuzzing during a `forge test` run, and we don't provide a way to specify Kontrol-specific functions
- symbolic tests written for other tools (Halmos and hevm, in particular) 

The changes introduced in this PR include:
- the ability to handle functions that start with `prove` and `check` prefixes similarly to those starting with `test`
- a simple tests for this new functionality: [`prove_assert_true()`](https://github.com/runtimeverification/kontrol/blob/b1a384716b1936c2c6fb10663b77701f62302284/src/tests/integration/test-data/foundry/test/Simple.t.sol#L24); this test is skipped in the legacy CI in order to not increase it's runtime and since the change is relatively minor


src/kontrol/__main__.py
```diff
@@ -244,6 +244,10 @@ def exec_prove(
     if isinstance(kore_rpc_command, str):
         kore_rpc_command = kore_rpc_command.split()
 
+    for test_name, _ in tests:
+        if not test_name.startswith('test') or test_name.startswith('testFail'):
```
tothtamas28, line 248:
```suggestion
        if not test_name.startswith('test') or not test_name.startswith('testFail'):
```

(If this is the intended condition, then the second disjunct can be dropped, see comment below.)

tothtamas28: Btw this forbids

```
kontrol prove --match-test '.*'
```

Is this intended?

palinatolmach:
As highlighted by @RaoulSchaffranek and @anvacaru, some of our use-cases such as the debugger and CSE require executing functions that do not start with `test`. In this case, we should probably make it clear that we're not checking for violations in the documentation and/or through a warning message to the user.

palinatolmach:
Based on the discussion in today's standup, we decided to handle tests prefixed with `prove` and `proveFail` similarly to those starting with `test`. That would help us define Kontrol-specific functions that would _not_ get executed during a `forge test` run and would also make Kontrol compatible with Halmos and hevm symbolic tests.

src/kontrol/__main__.py
palinatolmach: It wasn't, thanks for noticing!

src/kontrol/__main__.py
```diff
@@ -244,6 +244,10 @@ def exec_prove(
...
+            raise RuntimeError(f"Test name should start with 'test' or 'testFail' to be verified: {test_name}")
```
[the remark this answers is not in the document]
palinatolmach: Right, thank you, Tamás! I'll mark these comments as resolved since we decided not to implement it in this way, but I appreciate your comments, I haven't realised neither of these things myself.

src/kontrol/prove.py
```diff
@@ -413,7 +413,8 @@ def _method_to_cfg(
         new_node_ids = [init_node.id]
         init_node_id = init_node.id
 
-    is_test = method.signature.startswith('test')
+    proof_prefixes = ['test', 'proof', 'check']
+    is_test = any(method.signature.startswith(prefix) for prefix in proof_prefixes)
     failing = method.signature.startswith('testFail')
```
JuanCoRo, line 418: Should we also have the dual logic for `proofFail` and `checkFail`?

palinatolmach: Yeah, I wasn't sure about that too, thanks @JuanCoRo! My reasoning behind not adding it was mostly the following:
- IIUC, the usage of `testFail` in Foundry itself is [generally discouraged](https://github.com/foundry-rs/foundry/issues/4437), considering that there is `vm.expectRevert()` cheatcode; it's a bit philosophical, but it also looks to me that proofs are not supposed to fail
- I don't think `hevm` and `Halmos` support the `...Fail` counterpart of these prefixes, so it not supporting them doesn't negatively affect compatibility with test suites written for these tools

But I agree they might be useful in some scenarios, so please let me know if you think I should add `proofFail` and `checkFail`, and I will!

anvacaru: I don't know, is it common to use `proofFail`/`checkFail`?

anvacaru:
Can you test this by renaming one of the existing tests (for which we check the output) with the new prefixes?

src/kontrol/prove.py
palinatolmach: Well, I don't think anyone's supporting (or using) `proofFail` and `checkFail` per se, but maybe you're right @JuanCoRo and t makes sense to add those for consistency with `test` and `testFail`.

src/kontrol/prove.py
```diff
@@ -413,7 +413,8 @@ def _method_to_cfg(
         new_node_ids = [init_node.id]
         init_node_id = init_node.id
 
-    is_test = method.signature.startswith('test')
+    proof_prefixes = ['test', 'proof', 'check']
```
tothtamas28, line 416:
```suggestion
    proof_prefixes = ['test', 'prove', 'check']
```

tothtamas28: By the way, why support both `prove` and `check`? If there's no specific reason, I think we should just pick one of the two.

palinatolmach: So we can automatically verify tests written for other symbolic testing tools: `hevm` (at least, the last time I checked) symbolically executes tests starting with `prove`, and `halmos` — tests starting with `check`.

palinatolmach: Thanks for noticing the wording issue! Fixed "prove" spelling in https://github.com/runtimeverification/kontrol/pull/249/commits/0b4e7357e6df8f1a089bdf779774b653cbc8734f.

src/kontrol/prove.py
```diff
@@ -419,8 +419,9 @@ def _method_to_cfg(
         new_node_ids = [init_node.id]
         init_node_id = init_node.id
 
-    is_test = method.signature.startswith('test')
-    failing = method.signature.startswith('testFail')
+    proof_prefixes = ['test', 'prove', 'check']
+    is_test = any(method.signature.startswith(prefix) for prefix in proof_prefixes)
+    failing = any(method.signature.startswith(prefix + 'Fail') for prefix in proof_prefixes)
```
tothtamas28, line 424: The logic for deciding what sort of KCFG to construct based on the function name should be pulled up from this function, and `is_test` and `failing` should become parameters (not necessarily in this PR though).

Additionally, `proof_prefixes` is the sort of thing that should be configurable through a configuration file. If it doesn't break Foundry we can consider adding a section to `foundry.toml`, otherwise maybe its time to introduce `kontrol.toml`. This would enable cleaning up the CLI a bit as well.

palinatolmach:
> Can you test this by renaming one of the existing tests (for which we check the output) with the new prefixes?

@anvacaru I renamed one of the tests for which we check expected output from `AssertTest.testFail_assert_false()` to `AssertTest.checkFail_assert_false()` in https://github.com/runtimeverification/kontrol/pull/249/commits/51848f445207df65586ca045f40f71fde6557722. To avoid increasing the CI load, I removed `testFail_assert_false`, so there's only `checkFail_assert_false` now. I find it sufficient, since the change is rather small and we're checking similar functionality with other `testFail` tests. `prove` prefix is also checked in `AssertTest.prove_assert_true()`, though without the expected output check.

Please let me know if you think we should test it more exhaustively, and thank you for all the suggestions!

src/kontrol/prove.py
tothtamas28: > The logic for deciding what sort of KCFG to construct based on the function name should be pulled up from this function

Then this feature could also be tested with a simple unit test.

palinatolmach: Yes, thanks Tamás, I agree! Halmos allows specifying the function name using the following argument, and we can do the same thing, introducing, e.g., `--function` and `--fail-function` as the first step:
```
    parser.add_argument(
        "--function",
        metavar="FUNCTION_NAME_PREFIX",
        default="check_",
        help="run tests matching the given prefix. Shortcut for `--match-test '^{PREFIX}'`. (default: '%(default)s')",
    )
```

And agreed on `.toml` configuration, we have in on a roadmap for February, but I wouldn't mind doing it earlier if I have time next week. Adding a new section (e.g., `[kontrol]` or `[symbolic]`) won't break Foundry but it will throw a warning:
```
warning: Found unknown config section in foundry.toml: [kontrol]
This notation for profiles has been deprecated and may result in the profile not being registered in future versions.
Please use [profile.kontrol] instead or run `forge config --fix`.
```

palinatolmach: @JuanCoRo needs this for an engagement, so I'm leaning towards leaving this PR as it right now, as it's the format we've agreed on so far, but I'll open a new PR right after that, if that works. Or please let me know what you think @JuanCoRo!

anvacaru (approved):

Skippyy96 (commented):
Thank you

PR #252, opened 2023-12-15
Disable gas computations in Kontrol

anvacaru:
~blocked on : https://github.com/runtimeverification/kontrol/pull/253~

Using the new `<useGas>` cell in evm-semantics, this PR turns off the computation of gas costs by default in Kontrol.

- adding a new `--use-gas` flag to `kontrol prove` that enables gas computations.
- modified the test harness to execute the test for the auto-abstraction of the gas cell with the `--use-gas` mode.
- added a `foundry-prove-with-gas` file to track tests that should be executed with the active `<use-gas>` cell.

It looks like with the `<use-gas>` cell set to `false,` we also lose information from the `<memoryUsed>` cell and `<accessedAccounts>.`


palinatolmach:
@JuanCoRo do we use `<accessedAccounts>` in deployment summary building? Is is a problem that we lose this information in no-gas mode?

PetarMax:
> @JuanCoRo do we use `<accessedAccounts>` in deployment summary building? Is is a problem that we lose this information in no-gas mode?

We shouldn't be losing this information for CSE for sure. It's important so that we know which information from the configuration we can drop in the summaries. 

In fact, we don't lose it for the `CALL*` opcodes (the accounts are accessed only when the call is actually made), but do lose it for
```
    rule <k> #gasAccess(SCHED, EXTCODESIZE ACCT)            => #accessAccounts ACCT ~> Caddraccess(SCHED, ACCT in ACCTS)                                           ... </k> <accessedAccounts> ACCTS </accessedAccounts>
    rule <k> #gasAccess(SCHED, EXTCODECOPY ACCT _ _ _)      => #accessAccounts ACCT ~> Caddraccess(SCHED, ACCT in ACCTS)                                           ... </k> <accessedAccounts> ACCTS </accessedAccounts>
    rule <k> #gasAccess(SCHED, EXTCODEHASH ACCT)            => #accessAccounts ACCT ~> Caddraccess(SCHED, ACCT in ACCTS)                                           ... </k> <accessedAccounts> ACCTS </accessedAccounts>
    rule <k> #gasAccess(SCHED, BALANCE ACCT)                => #accessAccounts ACCT ~> Caddraccess(SCHED, ACCT in ACCTS)                                           ... </k> <accessedAccounts> ACCTS </accessedAccounts>
    rule <k> #gasAccess(SCHED, SELFDESTRUCT ACCT)           => #accessAccounts ACCT ~> #if ACCT in ACCTS #then 0 #else Gcoldaccountaccess < SCHED > #fi            ... </k> <accessedAccounts> ACCTS </accessedAccounts>
```

PetarMax:
On second thought, it's probably fine to drop `<accessedAccounts>` information in no-gas mode, since it's only used for gas computations. There are other ways of figuring out which parts of the configuration are needed for CSE.

palinatolmach (approved):

palinatolmach:
LGTM!

JuanCoRo:
> @JuanCoRo do we use `<accessedAccounts>` in deployment summary building? Is is a problem that we lose this information in no-gas mode?

We don't use the `<accessedAccounts>` when building summaries. For now, we just take what [`stopAndReturnStateDiff`](https://book.getfoundry.sh/cheatcodes/stop-and-return-state-diff#accountaccess) gives us in a JSON and put it in cheatcodes to run them with Kontrol. I don't think we'll need the information of `<accessedAccounts>`.

PR #261, opened 2024-01-09
Enable breaking on all foundry actions

ehildenb:
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2242~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2246~

This PR is about making it possible to generate KEVM specifications from Kontrol proofs. When a Foundry rule fires in the middle of a basic block, that basic block is no longer provable directly with KEVM. So we add the option to intercept these calls so that basic blocks are formed before and after the Foundry rule. In particular:

- A handful more rules in `FOUNDRY` semantics are labelled for being able to break on them.
- The option `--break-on-foundry` is added, which makes sure that _every_ labelled rule in `FOUNDRY` is broken on.


src/kontrol/__main__.py
```diff
@@ -733,6 +735,13 @@ def _parse_test_version_tuple(value: str) -> tuple[str, int | None]:
     prove_args.add_argument(
         '--use-gas', dest='use_gas', default=False, action='store_true', help='Enables gas computation in KEVM.'
     )
+    prove_args.add_argument(
```
anvacaru, line 738: Can you rename this to `--break-on-cheatcodes` instead?

src/kontrol/prove.py
```diff
@@ -207,18 +207,65 @@ def init_and_run_proof(test: FoundryTest) -> Proof:
...
             )
 
+            cut_point_rules = KEVMSemantics.cut_point_rules(
+                prove_options.break_on_jumpi,
+                prove_options.break_on_calls,
+                prove_options.break_on_storage,
+                prove_options.break_on_basic_blocks,
+            )
+            if prove_options.break_on_foundry:
+                cut_point_rules += [
```
anvacaru, line 217: Adding only `FOUNDRY-CHEAT-CODES.foundry.call` should be enough, since it captures all the calls made to the cheat code address.

ehildenb: We want to break on _every_ rule that is added in this file, not just on "is something happening for Foundry". This is to make sure we're maknig basic blocks that don't contain _any_ rules from this file in it.

src/kontrol/__main__.py
ehildenb: Done.

src/kontrol/prove.py
anvacaru: Can we have `FOUNDRY-CHEAT-CODES.foundry` instead? Are we supporting partial matching? All the rules enumerated here make this list easy to get out of date.

anvacaru (approved):

src/kontrol/prove.py
ehildenb: I don't think partial matching is supported, but I guess that's something we could ask the backend team about.

I suppose we could have pyk instead read the names of the rules directly from the definition, let me see if that's possible.

ehildenb: I've done this, now it's enumerated directly from the definition instead.

PR #264, opened 2024-01-10
Refactor init and final terms for symbolic exploration

anvacaru:
This PR changes how the K configuration of a proof's initial and final nodes is generated for symbolic exploration (Solidity functions that do not start with `['proof,' 'check,' 'test,' 'proofFail' ..]`. *This does not include the `setUp()` and `constructor()` functions which are a special case*.

- The state is more general (required for CSE):
    - make the `callStack`, `callState`, `id`, and `logs` symbolic
    - save the function program bytecode in an account with a symbolic address, balance, nonce and storage instead of the concrete `Foundry.address_TEST_CONTRACT`.
  
**Other Changes**
- remove some cells from `init_subst` that get already initialized by the k rules from [cheatcodes.md](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/kdist/cheatcodes.md):
 - clean the `_init_cterm` function a bit:
     - remove unused args `contract_name` and `setup_cterm`
     - replace `KApply(.Set)` and `KApply(.List)` with `set_empty()` and `list_empty()`.
     - factor out the logic that generates the list of accounts in a private function.


src/kontrol/prove.py
```diff
@@ -500,6 +495,43 @@ def _init_cterm(
...
+
+        constraints = [
+            mlEqualsTrue(KEVM.range_uint(256, KVariable('CONTRACT_BAL'))),
+            mlEqualsTrue(KApply('#rangeNonce(_)_WORD_Bool_Int', KVariable('CONTRACT_NONCE'))),
+            mlEqualsTrue(
+                notBool(
+                    KApply('#isPrecompiledAccount(_,_)_EVM_Bool_Int_Schedule', [KVariable('CONTRACT_ID'), schedule])
+                )
+            ),
+        ]
```
anvacaru, line 533: These are captured by: https://github.com/runtimeverification/evm-semantics/pull/2249

src/kontrol/prove.py
```diff
@@ -571,6 +552,19 @@ def _init_cterm(
...
+    if symbolic_exploration:
+        init_subst.update(
+            {
+                'CALLSTACK_CELL': KVariable('CALLSTACK'),
+                'CALLDEPTH_CELL': KVariable('CALLDEPTH'),
+                'LOG_CELL': KVariable('LOGS'),
+                'ID_CELL': Foundry.address_TEST_SYMBOLIC(),
+            }
+        )
+        del init_subst['CALLSTACK_CELL']
```
palinatolmach, line 564: I may not fully understand what is being done here, but do we want to delete `CALLSTACK_CELL` and `LOG_CELL` that we've just made symbolic?

src/kontrol/prove.py
```diff
@@ -638,6 +664,27 @@ def _final_term(empty_config: KInner, contract_name: str, use_init_code: bool =
...
+            KVariable('CONTRACT_ORIG_STORAGE_FINAL'),
+            KVariable('CONTRACT_NONCE_FINAL'),
+        )
+        final_subst.update(
+            {
+                'ID_CELL': Foundry.address_TEST_SYMBOLIC(),
+                'ACCOUNTS_CELL': KEVM.accounts(
+                    [
+                        contract,  # test contract address
+                        Foundry.account_CHEATCODE_ADDRESS(KApply('.Map')),
```
palinatolmach, line 682: Is `store_var` supposed to be an empty map in this case? Should we use [`map_empty()`](https://github.com/runtimeverification/pyk/blob/5c625925da8408d691a962f20ef416750b3aebdb/src/pyk/prelude/collections.py#L43), in this case?

src/kontrol/prove.py
anvacaru: In the `init_subst` above, `callStack` is initialized with an empty list with `'CALLSTACK_CELL': list_empty(),.`
`list_empty()` is the same as `KApply(.List)`, so the change from `KApply(.List)` to `list_empty()` does not affect the initial configuration in any way.

Deleting the `CALLSTACK_CELL` entry from the `init_subst` will prevent initializing the cell with an empty list and keep it as a symbolic variable.

src/kontrol/prove.py
anvacaru: Yes, I missed that. Thank you!

src/kontrol/prove.py
anvacaru: Oh, right, I understood what you mean! Deleting a cell substitution after I replaced it with a symbolic variable was indeed redundant. Thanks!

src/kontrol/prove.py
```diff
@@ -613,6 +594,17 @@ def _init_cterm(
...
+    if symbolic_exploration:
+        init_subst.update(
+            {
+                'ID_CELL': Foundry.address_TEST_SYMBOLIC(),
+            }
+        )
+        del init_subst['CALLSTACK_CELL']
+        del init_subst['CALLDEPTH_CELL']
+        del init_subst['LOG_CELL']
+        del init_subst['INTERIMSTATES_CELL']
```
PetarMax, line 606: These could also be removed from `init_subst` for symbolic exploration:
```
        'TOUCHEDACCOUNTS_CELL': set_empty(),
        'ACCESSEDACCOUNTS_CELL': set_empty(),
        'ACCESSEDSTORAGE_CELL': map_empty(),
```
Alternatively, they could be added in post-production of obtained specifications.

src/kontrol/prove.py
```diff
@@ -680,6 +704,27 @@ def _final_term(empty_config: KInner, contract_name: str, use_init_code: bool =
...
+            KVariable('CONTRACT_ORIG_STORAGE_FINAL'),
+            KVariable('CONTRACT_NONCE_FINAL'),
+        )
+        final_subst.update(
+            {
+                'ID_CELL': Foundry.address_TEST_SYMBOLIC(),
+                'ACCOUNTS_CELL': KEVM.accounts(
+                    [
+                        contract,  # test contract address
+                        Foundry.account_CHEATCODE_ADDRESS(KApply('.Map')),
```
PetarMax, line 722: Should this be symbolic for symbolic exploration, just as for regular tests?

src/kontrol/prove.py
```diff
@@ -680,6 +704,27 @@ def _final_term(empty_config: KInner, contract_name: str, use_init_code: bool =
...
     }
+    if not is_test:
+        contract = KEVM.account_cell(
+            Foundry.address_TEST_SYMBOLIC(),
+            KVariable('CONTRACT_BAL_FINAL'),
+            program,
+            KVariable('CONTRACT_STORAGE_FINAL'),
+            KVariable('CONTRACT_ORIG_STORAGE_FINAL'),
+            KVariable('CONTRACT_NONCE_FINAL'),
+        )
```
PetarMax, line 715: This can be streamlined, see above comment.

src/kontrol/prove.py
```diff
@@ -624,10 +616,42 @@ def _init_cterm(
...
+def _create_initial_account_list(
+    program: KInner, symbolic_exploration: bool, summary: Iterable[SummaryEntry] | None
+) -> list[KInner]:
+    _contract = (
+        KEVM.account_cell(
+            Foundry.address_TEST_SYMBOLIC(),
+            KVariable('CONTRACT_BAL'),
+            program,
+            KVariable('CONTRACT_STORAGE'),
+            KVariable('CONTRACT_STORAGE'),
```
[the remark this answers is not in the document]
anvacaru: Done

PetarMax (approved):

PR #260, opened 2024-01-09
Enable more Kontrol tests on booster

ehildenb:
This PR enables more proofs on CI by removing them from the skip file. It also makes some other small changes:

- The commit 5413d5d sorts the skipped test list, and then commits fd8a157 and 77a8980 adjust the tests. Overall, 179 more tests are enabled, and none are disabled.
- The module `pytest-timeout` is added for being out to timeout individual tests by adding option `--timeout` as such: `make cov-integration ... TEST_ARGS="--timeout=7200"` for example.
- A small re-arrangement of `test_foundry_prove.py` is added.

Issue #197, opened 2023-11-22
Optimize Foundry test suite


anvacaru:
The Foundry test suite we use for CI proofs must be better organized. With duplicated test contracts (such as PlainPrankTest.t.sol and `PrankTest.t.sol`), contracts that do not use the `setUp` functionality repeatedly deploy auxiliary contracts in the tests. In addition, a lot of tests are skipped to keep the CI time low.

A few optimizations:
- [ ] Get a smaller test-suite codebase, resulting in lower kompile time:
    - [ ] Merge duplicated contracts
    - [ ] Remove/merge tests with duplicated logic

- [ ] Improve performance of the tests:
    - [ ] Use the `setUp` functionality instead of deploying auxiliary contracts in the test.
    - [ ] replace `assertEq` and `assertGt` assertion calls with `assert`. Since `assertEq` is a function from the forge test library,  this will reduce the number of kcfg nodes generated by removing calls to the forge test contract. We can add individual unit tests to check that `assertEq` and `assert` act similarly.
    - [ ] use `notBuiltinAddress()` in unit tests that use symbolic addresses to avoid unnecesary branchings.

By implementing these we could (1) reduce CI time and/or (2) run more tests that are currently skipped in the same amount of time.

PR #269, opened 2024-01-12
Disable legacy backend integration tests

### LINKED ISSUE #248: Disable integration legacy tests on CI
Factored out from https://github.com/runtimeverification/kontrol/pull/75.

The tests on the legacy backend take very long, and we are already testing conformance in KEVM. Any discrepancies between booster and legacy should be upstreamed as KClaim style tests to KEVM.

- Disable legacy tests on CI, for integration, maintain them for profiling (faster).
- Update timeouts.
- Update required status checks.

Note that there is a discrepancy between the output of `test_foundry_auto_abstraction` when the booster is used vs. not. This is mostly due to ensures and requires conditions being ordered differently, also sometimes using the booster seems to add ensures clauses to rule blocks in generated modules w[here](https://github.com/runtimeverification/kontrol/blob/336c92e8c27d2434a038b12b45f6b46aac63d70c/src/tests/integration/test-data/gas-abstraction.expected#L872) the legacy backend does not, for example here.

There are also other tests in [test_foundry_prove.py](https://github.com/runtimeverification/kontrol/pull/75/files#diff-b04daa00a79f80dab79c7ae97738b41df215cea6715fba22514a6fef3efadeaa) that are skipping if the booster is used. I guess in the case of `test_foundry_kompile` it was because there is no need to run it with both.

nwatson22:
closes #248


palinatolmach:
@nwatson22 please note that there's a related PR opened by @ehildenb: https://github.com/runtimeverification/kontrol/pull/260/files; and [this](https://github.com/runtimeverification/kontrol/issues/197) related suggestion by @anvacaru.

anvacaru:
Closing this as it got stale over time and there's still a need to test the legacy integration ([see this message](https://github.com/runtimeverification/kontrol/pull/331#issuecomment-1927271390)).
In addition, a few PR's have been merged that skip tests with a longer execution.

Issue #266, opened 2024-01-11
Enable support for fast summarizations


anvacaru:
**Context:**
Starting with https://github.com/runtimeverification/kontrol/pull/221, kontrol can take a JSON state-diff object([example here](https://github.com/runtimeverification/_audits_Ethereum-optimism_pausability/blob/kontrol-proofs/packages/contracts-bedrock/snapshots/state-diff/Kontrol-Deploy.json)), representing the summary of a deployment process, and generate a Solidity contract([example here](https://github.com/runtimeverification/_audits_Ethereum-optimism_pausability/blob/kontrol-proofs/packages/contracts-bedrock/test/kontrol/proofs/utils/DeploymentSummary.sol)).

**Problem:**
Executing the summarized Solidity contract through Kontrol to initialize the state is inefficient because of the large amount of foundry cheat codes that only alter the account state.

**Possible solution:**
One way to speed up is to process the account operations from the state-diff in Python during the proof initialization stage and insert the resulting accounts to the initial node of the kcfg. In this way, we avoid double processing the state diff: once to generate the Solidity file and once to execute it to get the changes in the KCFG.

**Technical details:**
[This](https://github.com/runtimeverification/kontrol/pull/221/files#diff-59cc0840d8befef59d646ac05de0d54e60f10330329cfb480e8b586751a70230R723-R724) is where each entry in the state diff is processed, and the logic that shows how these entries are processed is in [Deployment.add_cheatcode](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/deployment.py#L84).
Starting from here, we can alter the logic to apply the changes over a list of [account_cells](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kevm.py#L400). 
These accounts should then be included in the initial node of the kcfg [in here](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/prove.py#L472).

PR #271, opened 2024-01-12
Implement fast summarization

anvacaru:
fixes: #266

**Changes:**
- add a new `--init-node-from` flag that takes a JSON file containing the summary of a deployment process. Takes the accounts obtained after going through the summary and inserts them in the `<accounts>` cell of the initial node of the KCFG.

- Extract the logic that reads the JSON summary files in `foundry_summary()` to its own `read_summary()` function.

- Refactor `DeploymentSummary` class:
     - extract the logic that processes a summary entry into its own `SummaryEntry` class.
     -  refactoring the `DeploymentSummary.add_cheatcode(dict)` that would process a summary entry and create the Solidity code for the cheat codes required to update the state to `DeploymentSummary.extend(SummaryEntry)`.
 
 - Add `summary_to_account_cells(dict)` that goes through the summary JSON object and creates a list of `<account>` cells. 
 
**Minor changes:**
- replace occurrences of  `KApply(.Map)` to `map_empty()`
- remove unused `_write_cfg` function

Next steps:
 1. Apply the state diff changes over an account that has symbolic fields  like `{'balance': KVariable('ACCT_X_BAL'), 'nonce': KVariable('ACCT_X_NONCE') , ...}` instead of `{'balance': 0, 'nonce': 0, 'code': '', 'storage': {}}`
 2. Insert the set of account addresses in `<accessedAccounts>`,`<touchedAccounts>` (not sure actually?)
 3. Currently, the storage updates from the summary are placed only in `<storage>`, apply them in `<origStorage>` as well.


src/kontrol/options.py
```diff
@@ -28,6 +28,7 @@ class ProveOptions:
     fail_fast: bool
     reinit: bool
     use_gas: bool
+    summary: dict | None
```
tothtamas28, line 31: Will this eventually be transformed into a `Deploymentsummary`? If so, consider doing it right in `__init__` (or even earlier), as the sooner you get rid of data that's not validated / whose type is not statically known the better,

src/kontrol/prove.py
```diff
@@ -430,6 +440,57 @@ def _method_to_cfg(
...
+            _addr_account_storage = hex_string_to_int(account_storage)
+            _init_account(_addr_account_storage)
+            accounts[_addr_account_storage]['storage'][intToken(hex_string_to_int(slot))] = intToken(
+                hex_string_to_int(new_value)
+            )
+
+    return accounts
+
+
+def summary_to_account_cells(summary: dict) -> list[KApply]:
```
tothtamas28, line 476: This seems like the central feature in this PR, please add a (parametric) unit test that exercises it. This would catch changes to the `summary` schema.

src/kontrol/prove.py
```diff
@@ -439,15 +500,23 @@ def _init_cterm(
...
+        map_empty(),
+        map_empty(),
         intToken(1),
     )
+    init_account_list: list[KInner] = [
+        account_cell,  # test contract address
+        Foundry.account_CHEATCODE_ADDRESS(map_empty()),
+    ]
+    if summary is not None:
+        init_account_list.extend(summary_to_account_cells(summary))
```
tothtamas28, line 518: Is extending the list safe, i.e. is it possible for summary accounts to somehow interfere with the other two accounts (e.g. same address)? If so, consider doing some validation.

src/kontrol/prove.py
tothtamas28: It would be good to convert `summary` into a statically typed data structure soon (or right) after reading the data from disk.

src/kontrol/options.py
anvacaru: No, the `DeploymentSummary` is used only in the functionality of the `foundry_summary` function to generate the Solidity contract based on the `summary` dict.
But I get the idea and I can process the `summary` into a list of `SummaryEntry`s.

src/kontrol/deployment.py
```diff
@@ -1,3 +1,51 @@
...
+    @property
+    def is_create(self) -> bool:
+        return self.deployed_code != '0x' and self.kind == 'Create'
+
+    @property
+    def updates_balance(self) -> bool:
+        return self.new_balance != self.old_balance
+
+    @property
+    def storage_updates(self) -> list[tuple[str, str, str]]:
```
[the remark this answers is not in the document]
anvacaru: Implemented in https://github.com/runtimeverification/kontrol/pull/271/commits/751d5d70895e7db0f3ee38663ff8dc25a8819858.

src/kontrol/foundry.py
```diff
@@ -841,9 +836,17 @@ def foundry_get_model(
     return '\n'.join(res_lines)
 
 
-def _write_cfg(cfg: KCFG, path: Path) -> None:
-    path.write_text(cfg.to_json())
-    _LOGGER.info(f'Updated CFG file: {path}')
+def read_summary(accesses_file: Path, contract_names: Path | None = None) -> tuple[dict, dict]:
```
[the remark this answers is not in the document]
anvacaru: Done in https://github.com/runtimeverification/kontrol/pull/271/commits/2ba5fe5f39b7313e9b058f4183c37ecbbb35ded3.

src/kontrol/options.py
anvacaru: Now, `summary` is `list[SummaryEntry] | None`.

src/kontrol/prove.py
anvacaru: Added a unit test in https://github.com/runtimeverification/kontrol/pull/271/commits/55155001459a65c4c4db8ef038cb7c8f76ddc8c6, and`summary` is now a `list[SummaryEntry]`.

src/kontrol/deployment.py
```diff
@@ -1,3 +1,56 @@
...
+    @property
+    def is_create(self) -> bool:
+        return self.deployed_code != '0x' and self.kind == 'Create'
+
+    @property
+    def updates_balance(self) -> bool:
+        return self.new_balance != self.old_balance
+
+
 class DeploymentSummary:
```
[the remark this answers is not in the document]
anvacaru: Yes, this handles the generation of Solidity code. There currently is an [integration test in place].(https://github.com/runtimeverification/kontrol/blob/master/src/tests/integration/test_foundry_prove.py#L507)

src/kontrol/deployment.py
```diff
@@ -1,3 +1,56 @@
...
+            slot = storage_access['slot']
+            is_write = storage_access['isWrite']
+            previous_value = storage_access['previousValue']
+            new_value = storage_access['newValue']
+            reverted = storage_access['reverted']
+
+            if reverted or not is_write or new_value == previous_value:
+                continue
+
+            storage_changes.append(StorageUpdate(account_storage, slot, new_value))
```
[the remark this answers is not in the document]
anvacaru: Done in https://github.com/runtimeverification/kontrol/pull/271/commits/76e84517fbdd026166eedb63f972b932cf1c7cc7.

src/kontrol/prove.py
```diff
@@ -430,6 +439,55 @@ def _method_to_cfg(
     return cfg, new_node_ids, init_node_id, target_node.id
 
 
+def _process_summary(summary: list[SummaryEntry]) -> dict:
```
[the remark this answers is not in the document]
anvacaru: Moved it in https://github.com/runtimeverification/kontrol/pull/271/commits/76e84517fbdd026166eedb63f972b932cf1c7cc7.

ehildenb:
It would be good to test this on the Optimism codebase where the summaries are being used. @JuanCoRo can you confirm that this works for your needs?

palinatolmach:
Looks good to me too! I think @anvacaru tested it on the engagement proof of interest, but I'll wait for an additional confirmation before approving and merging.

palinatolmach (approved):

PR #279, opened 2024-01-16
Use special exit codes for passed and failed tests

palinatolmach:
Follow up to https://github.com/runtimeverification/evm-semantics/pull/2258.

This PR introduces different exit codes for three `kontrol prove` execution results, which should make it easier to automatically process these results, e.g., for benchmarking:
- `0` if there are no failed tests
- `2` if there are failed tests and a model has been generated for each of them
- `3` if there are failed tests, but at least one of them doesn't have a model

Exit code `1` is currently used for Foundry-related errors ([one](https://github.com/runtimeverification/kontrol/blob/fff05884eef5f61d92dffa2d0d7cb1116513dab3/src/kontrol/__main__.py#L73), [two](https://github.com/runtimeverification/kontrol/blob/fff05884eef5f61d92dffa2d0d7cb1116513dab3/src/kontrol/foundry.py#L251)).


palinatolmach:
@anvacaru @ehildenb can I ask for your feedback on this?

palinatolmach:
OTOH, @anvacaru suggested that we only exit with a non-zero code for failing tests w/no counterexample generated, as both passing and failing tests with counterexamples generated successfully are valid results. Which makes sense to me, but I'm not sure if it'd be as useful for automatically processing these results in benchmarking scripts etc.

PR #286, opened 2024-01-17
Annotating rules that preserve definedness

PetarMax:
This PR aims to annotate all of the Kontrol rules for which we know that they preserve definedness, but the backend is not able to prove this on its own.


src/kontrol/kdist/cheatcodes.md
```diff
@@ -345,6 +345,7 @@ The `<output>` cell will be updated with the value of the address generated from
          <k> #call_foundry SELECTOR ARGS => . ... </k>
          <output> _ => #bufStrict(32, #addrFromPrivateKey(#unparseDataBytes(ARGS))) </output>
       requires SELECTOR ==Int selector ( "addr(uint256)" )
+      [preserves-definedness]
```
PetarMax, line 348: In this rule, the construction of `#addrFromPrivateKey` ensures that the resulting value is in the range `[0, pow256)`.

src/kontrol/kdist/cheatcodes.md
```diff
@@ -896,6 +897,7 @@ The `ECDSASign` function returns the signed data in [r,s,v] form, which we conve
          <k> #call_foundry SELECTOR ARGS => . ... </k>
          <output> _ => #sign(#range(ARGS, 32, 32),#range(ARGS,0,32)) </output>
       requires SELECTOR ==Int selector ( "sign(uint256,bytes32)" )
+      [preserves-definedness]
```
PetarMax, line 900: In this rule, by definition of
```
rule #sign(BA1, BA2) => #parseByteStack(ECDSASignbytes(BA1, BA2)) [concrete]
```
and the totality of `ECDSASignbytes` (assuming its correct implementation), `#parseByteStack` should always be able to always unparse the result. With this in mind, could we mark `#sign` as total?

PetarMax:
[Related question](https://github.com/runtimeverification/evm-semantics/pull/2260/#issuecomment-1895610849).

src/kontrol/kdist/cheatcodes.md
jberthold: Is `#unparseDataBytes(ARGS)` total? From looking at its defining rule, I have some doubts.

PetarMax: We have that:
```
rule #unparseDataBytes( DATA ) => replaceFirst(Base2String(#asInteger(#asByteStack(1) +Bytes DATA), 16), "1", "0x")
```
Then, since `#asByteStack`, `+Bytes`, `#asInteger`, and `replaceFirst` are total by definition, the only possible issue would be that `Base2String` could be undefined for its arguments. From the documentation we have:
```
...
You can also convert between a `String` and an `Int` in a particular radix.
This radix can be anywhere between 2 and 36. For a radix 2 <= N <= 10, the
digits 0 to N-1 will be used. For a radix 11 <= N <= 36, the digits 0 to 9
and the first N-10 letters of the Latin alphabet will be used.
...
```
and so the partiality of `Base2String` seems to be in the second argument, which needs to be between 2 and 36, which holds, since it is 16. Another possible partiality would be that the first argument cannot be less than zero, but it isn't since `#asInteger` always returns a non-negative number. Therefore, I believe that `#unparseDataBytes` is effectively total.

palinatolmach:
The [Docker test](https://github.com/runtimeverification/kontrol/actions/runs/8037441270/job/21963872890?pr=286) is failing in `solc_to_k` with
```
  File "/home/user/.local/lib/python3.10/site-packages/kontrol/solc_to_k.py", line 580, in __init__
    self.contract_path = self.contract_json['ast']['absolutePath']
KeyError: 'ast' 
```
which means that it can’t find the `ast` field in the compilation artifact file generated by `forge build` — that doesn't seem related to the changes made in the PR, I'll look into why that might be happening.

anvacaru (approved):

PR #289, opened 2024-01-17
Avoid using the same priority for rules that apply on the same KItem

anvacaru:
Curently the rules for checking an expected revert and injecting a prank both use the `#next[ OP ]` K production to match on and rewrite the logic. In both cases, the rules have a higher `priority(40)` and their `require` clauses are different.
This means that in a case in which a test uses both `prank` and `expectRevert` these rules would branch because they have the same priority and can be applied.
```
    rule [foundry.set.expectrevert.1]:
         <k> #next [ _OP:CallOp ] ~> (. => #checkRevert ~> #updateRevertOutput RETSTART RETWIDTH) ~> #execute ... </k>
         <callDepth> CD </callDepth>
         <wordStack> _ : _ : _ : _ : _ : RETSTART : RETWIDTH : _WS </wordStack>
         <expectedRevert>
           <expectedDepth> CD </expectedDepth>
           ...
         </expectedRevert>
      [priority(40)]
```

```
    rule [foundry.prank]:
         <k> #next [ OP:OpCode ] => #injectPrank ~> #next [ OP:OpCode ] ~> #endPrank ... </k>
         <callDepth> CD </callDepth>
         <id> ACCT </id>
         <prank>
           <active> true </active>
           <newCaller> NCL </newCaller>
           <depth> CD </depth>
           ...
         </prank>
      requires ACCT =/=K NCL
       andBool (OP ==K CALL orBool OP ==K CALLCODE orBool OP ==K STATICCALL orBool OP ==K CREATE orBool OP ==K CREATE2)
      [priority(40)]
```

To avoid branchings, I'm changing the priority on the `expectedRevert` rules. I've tested this on the proofs from the foundry-demo repo and the issue is now fixed.


palinatolmach (approved):

ehildenb:
Please add a test in the Kontrol test-suite which demonstrates the problem as shown in the Foundry demo repo @anvacaru

anvacaru:
@ehildenb I added a new test, and I double checked that it fails on master and passes on this branch.

ehildenb:
Awesome, thank you!

PR #294, opened 2024-01-18
Publish Kontrol to Cachix binary cache

### LINKED ISSUE #290: Binary caching of installed Kontrol tool
When installing Kontrol, it can take quite a while to build the tool. It would be nice to do some similar binary caching that we do for K project on the Kontrol project, where we build and push that on CI.

@goodlyrottenapple is this easy to set up?

goodlyrottenapple:
Fixes #290


goodlyrottenapple:
@ehildenb i tested the nix binary cache run in this commit https://github.com/runtimeverification/kontrol/pull/294/commits/bc88e0051ee8338f2d2622355d9465aa9ffe4671 . We should be able to merge this, but i will be off tomorrow so may be best to merge on Monday so I can catch any issues then and there.

PR #298, opened 2024-01-18
Fix function default values

nwatson22:
Rectifies prove option function default values with default values with add_argument default values from `evm-semantics` `cli.py`


tothtamas28 (commented):
Having to specify default values in multiple places is not ideal. Consider the following design.

* In the argument parser, `default` is not set anywhere (amounts to `default=None`).
* In `exec_prove`, each optional argument is `Optional`. No default value is specified.
  ```python3
  break_on_calls: bool | None
  ```
* In `ProveOptions.__init__`, each optional argument is `Optional`. All default values are set to `None`.
  ```python3
    break_on_calls: bool | None = None
  ```
* `ProveOptions.__init__` takes care of setting default values.
  ```python
  self.break_on_calls = bool(break_on_calls)  # default value is False
  self.failure_info = True if failure info is None else failure_info  # default value is True
  ```

Now default values for CLI and library use are consistent as there is a single source of truth  (`ProveOptions.__init__`). Moreover, functions that instantiate `ProveOptions` do not have to be aware of default values to pass through as they can simply pass `None`.

nwatson22:
@palinatolmach 

> @nwatson22 is this PR still relevant, or should it be closed in favor of #501?

No, I am closing this in favor of #501.

PR #308, opened 2024-01-22
Improvements to proof-loading speed on reruns

ehildenb:
These are pulled out of the CSE PR: https://github.com/runtimeverification/kontrol/pull/284. This includes:

- Telling our tools to ignore various generated files in the Foundry test directory (ignore Python files from Foundry in our formatting, and ignore generated files in our gitignore).
- Avoid loading lots of versions of a given proof when only one is called for. This is done by changing `Foundry.proofs_with_test(...) -> list[Proof]` to `Foundry.proof_ids_with_test(...) -> list[str]`, and allowing the call-site to decide whether to load the proof or not. This significantly reduces startup time when there are several versions of a given proof present.
- Renames internal function `run_prover` to prefixed version `_run_prover` to avoid confusion with `run_prover` imported from KEVM.
- Avoids setting up `kore-server` for already passed proofs, which speeds up the re-run time on a batch of proofs where some proofs have passed and others haven't.
- Removes a little dead-code.


src/kontrol/foundry.py
```diff
@@ -375,13 +368,19 @@ def help_info() -> list[str]:
...
-    def proofs_with_test(self, test: str) -> list[Proof]:
-        proofs = [
-            self.get_optional_proof(pid)
-            for pid in listdir(self.proofs_dir)
-            if re.search(single(self._escape_brackets([test])), pid.split(':')[0])
+    def proof_ids_with_test(self, test: str, version: int | None = None) -> list[str]:
+        regex = single(self._escape_brackets([test]))
+        all_proof_ids: list[tuple[str, str, int]] = []
+        for pid in listdir(self.proofs_dir):
+            proof_type = pid.split('%')[0]
```
nwatson22, line 375: I'm not sure this works. I think the `%` notation is not meant to demarcate a proof type, but rather as a separator that replaces the `/` in file path where the solidity file the function is found in relative to the foundry dir, so if a test is in `src/tests/integration/test-data/foundry/test/a/b/c/MyTest.t.sol` it would be something like `test%a%b%c%MyTest.test_function():0`

nwatson22 (changes requested):

src/kontrol/foundry.py
ehildenb: Ahhhhh, I see. So I guess we should basically take all but the last segment for this one.

ehildenb: Done.

nwatson22 (approved):

PR #313, opened 2024-01-23
Refactor production of initial terms, factor out method identification

ehildenb:
This PR is pulled out of: https://github.com/runtimeverification/kontrol/pull/284.

This brings us closer to producing specifications that are usable for CSE when running individual functions directly. In particular:

- New methods `{Method,Constructor}.{is_test,is_testfail}` are added for binning Solidity methods by their type.
- The logic for deciding whether to have a more or less symbolic state is inverted. Instead, we start with the most general state (substituting the fewest number of cells), then if it's a `test`/`testFail`/`setUp`, the state is made more specific with it needs to be (behind guards about the type of function it is).
- Modifications are made to the produced initial states to make them compatible for use in CSE:
  - The `ORIGIN_ID` and `CALLER_ID` are made to be `Int`.
  - The `OUTPUT_CELL` is made to start as the empty buffer.
  - The cells `[OUTPUT_CELL, CALLSTACK_CELL, CALLDEPTH_CELL, ID_CELL, LOG_CELL, ACCESSEDACCOUNTS_CELL, ACCESSEDSTORAGE_CELL, INTERIMSTATES_CELL, STATIC_CELL, ACCOUNTS_CELL]` are left fully symbolic for normal function calls, only filled in for `setUp` and `test` methods.
  - Similar changes are made to the final term production.
- We make sure to add the KEVM invariant to terminal states in the `setUp` for producing initial states in the test methods.
- Adds `pytest-timeout` module for timing out individual tests.


src/kontrol/prove.py
```diff
@@ -539,44 +538,43 @@ def _init_cterm(
...
         'GAS_CELL': KEVM.inf_gas(KVariable('VGAS')),
         'K_CELL': KSequence([KEVM.sharp_execute(), KVariable('CONTINUATION')]),
-        'ACCOUNTS_CELL': KEVM.accounts(init_account_list),
         'SINGLECALL_CELL': FALSE,
         'ISREVERTEXPECTED_CELL': FALSE,
         'ISOPCODEEXPECTED_CELL': FALSE,
+        'EXPECTEDADDRESS_CELL': KApply('.Account_EVM-TYPES_Account'),
+        'EXPECTEDVALUE_CELL': intToken(0),
+        'EXPECTEDDATA_CELL': bytesToken(b''),
+        'OPCODETYPE_CELL': KApply('.OpcodeType_FOUNDRY-CHEAT-CODES_OpcodeType'),
```
anvacaru, line 577: These cells are auxiliary for `expectOpcode` cheat code. They are initialized one the [setExpectOpcode](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/kdist/cheatcodes.md?plain=1#L1132) rule is applied.
Having them symbolic will not cause any branchings, because the `ISOPCODEEXPECTED_CELL` is set to FALSE above. 
We can leave them out of `init_subst`.

src/kontrol/prove.py
```diff
@@ -585,6 +583,24 @@ def _init_cterm(
...
+            'LOG_CELL': list_empty(),
+            'ACCESSEDACCOUNTS_CELL': set_empty(),
+            'ACCESSEDSTORAGE_CELL': map_empty(),
+            'INTERIMSTATES_CELL': list_empty(),
+            'TOUCHEDACCOUNTS_CELL': set_empty(),
+            'STATIC_CELL': FALSE,
+            'ACCOUNTS_CELL': KEVM.accounts(init_account_list),
+        }
+        for k, v in init_subst_test.items():
+            init_subst[k] = v
```
anvacaru, line 602: Instead of the `for`, you can update the `dict` using  `update()`. If `init_subst` already contains keys present in `init_subst_test`, their values are also updated.
```suggestion
        init_subst.update(init_subst_test)
```

src/kontrol/prove.py
```diff
@@ -675,14 +665,9 @@ def _final_cterm(
...
-def _final_term(empty_config: KInner, contract_name: str, is_test: bool, use_init_code: bool = False) -> KInner:
-    program = (
-        KEVM.init_bytecode(KApply(f'contract_{contract_name}'))
-        if use_init_code
-        else KEVM.bin_runtime(KApply(f'contract_{contract_name}'))
-    )
+def _final_term(empty_config: KInner, program: KInner, is_test: bool, is_setup: bool) -> KInner:
     post_account_cell = KEVM.account_cell(
-        Foundry.address_TEST_CONTRACT() if is_test else Foundry.address_TEST_SYMBOLIC(),
+        Foundry.address_TEST_CONTRACT(),
```
anvacaru, line 670: Can you also remove [`Foundry.address_TEST_SYMBOLIC()` ](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/foundry.py#L343)from foundry.py since it is not used anymore?

src/kontrol/prove.py
```diff
@@ -539,44 +538,43 @@ def _init_cterm(
...
-        'ACCESSEDSTORAGE_CELL': map_empty(),
-        'INTERIMSTATES_CELL': list_empty(),
-        'LOCALMEM_CELL': KApply('.Bytes_BYTES-HOOKED_Bytes'),
+        'ORIGIN_CELL': KVariable('ORIGIN_ID', sort=KSort('Int')),
+        'CALLER_CELL': KVariable('CALLER_ID', sort=KSort('Int')),
+        'LOCALMEM_CELL': bytesToken(b''),
+        'PREVCALLER_CELL': KApply('.Account_EVM-TYPES_Account'),
+        'PREVORIGIN_CELL': KApply('.Account_EVM-TYPES_Account'),
+        'NEWCALLER_CELL': KApply('.Account_EVM-TYPES_Account'),
+        'NEWORIGIN_CELL': KApply('.Account_EVM-TYPES_Account'),
```
anvacaru, line 564: These cells can be left symbolic and won't cause any branchings as long as `ACTIVE_CELL` is initialized to a concrete value. The other ones are initialized [here ](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/kdist/cheatcodes.md?plain=1#L1162).

anvacaru (changes requested):
Left some comments that should be addressed.

poetry.lock
```diff
@@ -910,6 +910,20 @@ pytest = ">=5.0"
 [package.extras]
 dev = ["pre-commit", "pytest-asyncio", "tox"]
 
+[[package]]
```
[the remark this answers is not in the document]
ehildenb: It's used locally for testing. You add `TEST_ARGS=--timemout 600`, to tell it to timeout individual tests (this is how I disable tests that run too long). There isn't another way to timeout individual tests with timeout. This package has already been added to KEVM.

src/kontrol/prove.py
ehildenb: Done.

src/kontrol/prove.py
ehildenb: Done.

src/kontrol/prove.py
ehildenb: Done.

src/kontrol/prove.py
ehildenb: Done.

poetry.lock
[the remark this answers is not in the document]
anvacaru: Thanks! Can I copy your comment into https://github.com/runtimeverification/kontrol/issues/319?

anvacaru (approved):
Approved! You might need to update the expected output again after https://github.com/runtimeverification/kontrol/pull/313/commits/6d470cb72c924639d75e9cc7e9e1dcb211242a16 .

poetry.lock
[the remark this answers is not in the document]
ehildenb: Sure!

Issue #291, opened 2024-01-18
Disambiguate labels for same-named contracts and functions using their absolute paths


anvacaru:
Similar to the change applied to contract names in https://github.com/runtimeverification/kontrol/pull/219, we should disambiguate labels for same-named contracts and functions using their absolute paths.

_Originally posted by @palinatolmach in https://github.com/runtimeverification/kontrol/issues/275#issuecomment-1893551745_

PR #317, opened 2024-01-29
Added project root relative path as method label prefixes

ovatman:
- `Method` class now takes `contract_name_with_path` and filter `contract_name` out of it.
- `Method.klabel`, `Method.unique_klabel` and `Method.qualified_name` are now generated with `contract_name_with_path`
- `Contract` uses `self.name_with_path` when constructing `Method`

Fix for [#291](https://github.com/runtimeverification/kontrol/issues/291)


anvacaru:
Looks good! Can you add a unit test in `src/tests/unit/test_solc_to_k.py` that would fail without these changes?

anvacaru (approved):

palinatolmach (approved):

Issue #324, opened 2024-01-30
`kontrol view-kcfg` command is not working  


lisandrasilva:
With the latest version of Kontrol (0.1.34), the `view-kcfg` command is not working although the proof is in the proofs directory. 

**Steps to reproduce**
- Clone this [repo](https://github.com/runtimeverification/_audits_Ethereum-optimism_pausability/tree/develop)
- Go to `packages/contracts-bedrock`
- From that folder, to reproduce the proofs, run the following (further documentation in `test/kontrol`): 
  - `./test/kontrol/scripts/make-summary-deployment` creates necessary files to run the proofs
  - `./test/kontrol/scripts/run-kontrol dev` runs the proofs in dev mode
- To run `kontrol` commands prepend them with `FOUNDRY_PROFILE=kprove`.
- Run `FOUNDRY_PROFILE=kprove kontrol list` and all the proofs are displayed:
```
APRProof: test%kontrol%proofs%L1CrossDomainMessengerKontrol.prove_relayMessage_paused(uint256,address,address,uint256,uint256):0
    status: ProofStatus.PASSED
    admitted: False
    nodes: 15
    pending: 0
    failing: 0
    vacuous: 0
    stuck: 0
    terminal: 2
    refuted: 0
Subproofs: 0

APRProof: test%kontrol%proofs%L1ERC721BridgeKontrol.prove_finalizeBridgeERC21_paused(address,address,address,address,uint256):0
    status: ProofStatus.FAILED
    admitted: False
    nodes: 21
    pending: 0
    failing: 1
    vacuous: 0
    stuck: 0
    terminal: 2
    refuted: 0
Subproofs: 0

APRProof: test%kontrol%proofs%L1StandardBridgeKontrol.prove_finalizeBridgeERC20_paused(address,address,address,address,uint256):0
    status: ProofStatus.FAILED
    admitted: False
    nodes: 25
    pending: 0
    failing: 1
    vacuous: 0
    stuck: 0
    terminal: 2
    refuted: 0
Subproofs: 0

APRProof: test%kontrol%proofs%L1StandardBridgeKontrol.prove_finalizeBridgeETH_paused(address,address,uint256):0
    status: ProofStatus.FAILED
    admitted: False
    nodes: 25
    pending: 0
    failing: 1
    vacuous: 0
    stuck: 0
    terminal: 2
    refuted: 0
Subproofs: 0

[...]

PR #326, opened 2024-01-31
Fetch the correct proof id in case of nested dir structure

anvacaru:
Closes: https://github.com/runtimeverification/kontrol/issues/324

**Changes:**
1. use a `rsplit('%', 1)` instead to extract only the contract name and avoid extra splits that are redundant; also avoiding extra processing of  `%.join()` after `split('%')` for the `proof_dir`.
2. extracts the filtering of the `proof_ids` logic into its own staticmethod in order to write unit tests.
3. Adding docstrings and unit tests.

**The problem:**
for a `proof id` like 
```py
test%kontrol%proofs%L1ContractName.prove_functionName(address,address,uint256):0
```
the `proof_name` was extracted like
```py
proof_name = pid.split('%')[1].split(':')[0]
```

What happens here is that:
 1. `pid.split('%'):` This splits the string `pid` at each `%` character into a list of substrings.
 4. `[1]`: This selects the second element of the list resulting from the split by `%`.
 5. `.split(':')`: This further splits the selected string at each `:` sepparating the `proof_name` from the `proof_version`.
 6. `[0]`: Finally, this selects the first element of the list resulting from the split by `:`, which is the `proof_name`.

The problem is that in the case in which the proof is in a nested dir, like for the one above in `test/kontrol/proofs/`, the `proof_name` gets the second value in the list, using the `[1]` and  resulting in an incorrect proof name `kontrol`.


ehildenb (approved):

ehildenb:
@anvacaru please add a test that would catch this bug!!!

src/kontrol/foundry.py
```diff
@@ -372,19 +372,30 @@ def help_info() -> list[str]:
         )
         return res_lines
 
+    @staticmethod
+    def filter_proof_ids(proof_ids: list[str], regex: str, version: int | None = None) -> list[str]:
+        """
+        Searches for proof IDs that match a specified test name and, optionally,a specific version.
+        Each proof ID is expected to follow the format 'proof_dir_1%proof_dir_2%proof_name:version'.
+        Only proof IDs that match the given criteria are included in the returned list.
```
anvacaru, line 380:
```suggestion
        Searches for proof IDs that match a specified test name and an optional version number.
        
        Each proof ID is expected to follow the format 'proof_dir_1%proof_dir_2%proof_name:version'.
        Only proof IDs that match the given criteria are included in the returned list.
```

src/kontrol/foundry.py
```diff
@@ -372,19 +372,31 @@ def help_info() -> list[str]:
...
+                proof_name, proof_version_str = proof_name_version.split(':', 1)
+                proof_version = int(proof_version_str)
+            except ValueError:
+                continue
+            if re.search(regex, proof_name) and (version is None or version == proof_version):
+                matches.append(f'{proof_dir}%{proof_name}:{proof_version}')
+        return matches
+
     def proof_ids_with_test(self, test: str, version: int | None = None) -> list[str]:
         regex = single(self._escape_brackets([test]))
```
ehildenb, line 396: Building the `regex` should probably happen directly in `filter_proof_ids`, so that we can just pass in the `test`, or the list of `test`, rather than the regex. That way othre methods calling this one don't need to worry about escaping the brackets.

anvacaru: Done.

PR #331, opened 2024-02-04
Disable legacy CI stages

anvacaru:
Extracted logic from #269.
Related issue: #248.

**What:**
- test-pr.yml: removes the `legacy` stages for `profile` and `integration` stages.
- src/integration/test_foundry_prove.py:
    - enable show tests on booster backend
    -  remove the use of `SKIPPED_LEGACY_TESTS`
- removing file [src/tests/integration/test-data/foundry-prove-skip-legacy](https://github.com/runtimeverification/kontrol/compare/anvacaru/disable-legacy?expand=1#diff-e32f799352773cd01a7e778ad69cb0e93308a11c2f071503359c8969191eaecf)


anvacaru:
@ehildenb, if you approve this, could you update the required status checks in the repository's settings? The `legacy` checks are still showing.

ehildenb:
@anvacaru I'd rather first have a PR which updates to use the booster for all the `show` tests, make sure that passes, then a second PR to reduce the number of tests we run with the legacy backend to make it fit in 30 minutes. I still want to test the legacy backend, because we want to make sure it still works over time (in an engagement, if we get stuck, one thing we should always try is with the legacy backend, to see if it can make progress even when the booster cannot).

anvacaru:
@ehildenb Understood. I'll close this PR until then.

Issue #316, opened 2024-01-25
`vm.prank` flakiness


JuanCoRo:
In some cases (unclear the moment what triggers this), `vm.prank()` will not take effect if surrounded by a `vm.expectRevert()`. When attempting to reproduce the example on simple enough code, `vm.prank()` has worked as expected, but not in more convoluted cases such as the Optimism proofs.

anvacaru:
@JuanCoRo  on what version did  this happen? Could it be related to [this](https://github.com/runtimeverification/kontrol/pull/289)?

PR #339, opened 2024-02-06
Avoid pranking on cheat codes

anvacaru:
Fixes: #316 

**Problem**:
`vm.prank` should set the caller and/or origin only for the next call.
Kontrol counts a cheat code invocation as a valid external call and would disable the prank after.

**Changes:**
Apply the prank only if the account being called is not `address(vm)`.


palinatolmach (approved):
lgtm!

PR #342, opened 2024-02-06
Migrate tests over to booster backend, minimize legacy backend time

ehildenb:
This PR migrates all the testing to emphasize the booster testing, while leaving a few tests run with the legacy prover to ensure that it continues to work.

- All `show` tests (including `gas-abstraction.expected`, `contracts.k.expected`, and `foundry.k.expeccted`) are put in the `show` folder.
- Any show-based output allows the file to not exist, and creates it if it doesn't.
- All tests except `test_foundry_prove` are disabled fully on legacy backend (using `pytest.skip()`).
- All proofs that take >300s on the legacy backend are skipped.
- Updated expected output.
- Decrease CI timeout to 120min.


src/tests/integration/test-data/show/AssumeTest.testFail_assume_true(uint256,uint256).expected
```diff
@@ -5,64 +5,57 @@
...
 ├─ 5
-│   k: #gas [ MSTORE , MSTORE 192 46308022326495007027972728677917914892729792999299745 ...
-│   pc: 3667
+│   k: #addr [ ISZERO ] ~> #exec [ ISZERO ] ~> #pc [ ISZERO ] ~> #execute ~> CONTINUATI ...
+│   pc: 4206
 │   callDepth: 0
 │   statusCode: STATUSCODE:StatusCode
 │
-│  (1000 steps)
+│  (827 steps)
```
ehildenb, line 29: This change in the resulting KCFG is concerning, and should probably be isolated into a regression test for the booster backend.

ehildenb: I've tested locally and it turns out that the booster and legacy backends actually produce the same output on this test. This test wasn't running before (being used for any output), because it's listed in the tests for the legacy backend to skip, and so was being skipped for the show step as well.

src/tests/integration/test-data/show/SetUpDeployTest.test_extcodesize().expected
```diff
@@ -5,86 +5,79 @@
...
-│
-│  (70 steps)
-├─ 4
 │   k: #end EVMC_SUCCESS ~> #pc [ RETURN ] ~> #execute ~> #codeDeposit 4914609233421842 ...
 │   pc: 145
 │   callDepth: 1
 │   statusCode: STATUSCODE:StatusCode
 │
 │  (1 step)
-├─ 5
```
ehildenb, line 23: Similarly, this test is not exercised on the legacy backend, so the show output was actually not being tested against before. Now the output is actually being generated by the booster.

ehildenb: I've tested that this one actually produces the same output on booster and legacy, and that the update to the test output is correct.

src/tests/integration/test-data/foundry-show
```diff
@@ -8,5 +8,4 @@ AssertTest.test_revert_branch(uint256,uint256)
 AssumeTest.test_assume_false(uint256,uint256)
 AssumeTest.testFail_assume_false(uint256,uint256)
 AssumeTest.testFail_assume_true(uint256,uint256)
-LoopsTest.sum_N(uint256)
```
ehildenb, line 11: This test wasn't actually being run on the legacy backend (because it's in the `foundry-prove-skip-legacy`), so it's deleted here and its output removed.

anvacaru (approved):
Looks good to me :+1:

palinatolmach (approved):

PR #347, opened 2024-02-07
Automate CSE

anvacaru:
Add a new flag`--cse` that allows to execute proofs with compositional symbolic execution.
Modified `solc-to-k` to identify and store the external calls of a function. 
With `cse`, `foundry-prove` will be recursively called on the external calls first.


src/tests/integration/test_foundry_prove.py
```diff
@@ -332,32 +332,20 @@ def test_foundry_dependency(
...
         prove_options=ProveOptions(
             max_iterations=50,
             bug_report=bug_report,
+            cse=True,
         ),
         rpc_options=RPCOptions(
             port=server.port,
         ),
-        include_summaries=[(dependency, None)],
+        include_summaries=[],
```
ehildenb, line 348: I guess we don't need the `include_summaries` parameter at all anymore.

ehildenb:
Remove the `include_summaries` parameter to `foundry_prove`, since we don't need it anymore and it isn't used anywhere.

Also, add some more complicated tests:

- One where the `test_...` method calls 2 _different_ functions (eg. `add` and `sub`).
- One where the `test_...` method calls another function, which calls another external function. Make sure it builds the summaries all the way down first.

src/tests/integration/test_foundry_prove.py
anvacaru: @PetarMax suggested that we should keep `include_summaries` for a while for testing purposes.

src/kontrol/prove.py
```diff
@@ -63,14 +63,20 @@ def foundry_prove(
...
-        else []
-    )
+    summary_ids: list[str] = []
+    if prove_options.cse:
+        test_suite = collect_tests(foundry, tests, reinit=prove_options.reinit, return_empty=True)
+        for test in test_suite:
+            if not isinstance(test.method, Contract.Method) or test.method.function_calls is None:
+                continue
+
+            test_version_tuples = [parse_test_version_tuple(t) for t in test.method.function_calls]
```
anvacaru, line 73: Here, I would include only functions that are not already in `summary_ids` (functions that have not been proved already).

palinatolmach:
Running `kontrol prove --match-test ArithmeticCallTest.test_double_add --no-fail-fast --cse` from this branch locally throws the following error:
```    
AttributeError: Error! Cannot use both --cse and --include-summary.
```
although it looks like [it shouldn't](https://github.com/runtimeverification/kontrol/blob/81fa872c8a98cffa52a912276828284a30f4081e/src/kontrol/prove.py#L67) if `--include-summary` is not provided; I'll look into that and the CI error: https://github.com/runtimeverification/kontrol/actions/runs/7896530778/job/21550730288?pr=347, and will enable more complicated [tests](https://github.com/runtimeverification/kontrol/pull/347/commits/051e0053069321d9af340ae03c2707bf87cd5602) that Andrei wrote once I confirm that it works.

palinatolmach:
I enabled a [more complicated test](https://github.com/runtimeverification/kontrol/pull/347/commits/051e0053069321d9af340ae03c2707bf87cd5602) that Andrei added, which calls two different functions — it [passes](https://github.com/runtimeverification/kontrol/actions/runs/7957190875/job/21719563040?pr=347) in ~18 minutes on CI. I [added](https://github.com/runtimeverification/kontrol/pull/347/commits/c99b632c7ec5e194992c839ceb9156c07d741e7d) the original simpler `test_double_add()` test to the list of skipped dependency tests in the interest of CI execution time, but can reenable it. 

@ehildenb @PetarMax however, a test that calls a function, which calls another external function fails with the following error related to the `<code>` cell: 
```
ValueError: Could not convert ML predicate to sort Bool: 
KApply(label=KLabel(name='#Exists', params=(KSort(name='GeneratedTopCell'),)), args=(KVariable(name='CODE', sort=KSort(name='Bytes')), KApply(label=KLabel(name='#Equals', params=(KSort(name='CodeCell'), KSort(name='GeneratedTopCell'))), args=(KVariable(name='_Gen0', sort=KSort(name='CodeCell')), KApply(label=KLabel(name='<code>', params=()), args=(KVariable(name='CODE', sort=KSort(name='Bytes')),))))))
```
which doesn't occur if CSE is disabled. The functions, however, are executed in the correct order 
- `ArithmeticContract.add(uint256,uint256)`
- `ArithmeticContract.add_sub_external(uint256,uint256,uint256)`
- `ArithmeticCallTest.test_double_add_sub_external(uint256,uint256,uint256)`).

The test is as follows:
```solidity
    function test_double_add_sub_external(uint x, uint y, uint z) external {
        uint a = arith.add_sub_external(x, y, z);
        a = arith.add_sub_external(a, y, z);
        assert(a > x);
    }
```
where `arith` is
```solidity
    function add(uint256 x, uint256 y) external pure returns (uint256 z) {
        z = x + y;
    }

    function add_sub_external(uint256 x, uint256 y, uint256 z) external view returns (uint256 res) {
        uint256 sum = this.add(x, y);
        res = sum - z;
    }
```
and bug report is attached: 
[cse_test_double_add_sub_external.zip](https://github.com/runtimeverification/kontrol/files/14331550/cse_test_double_add_sub_external.zip)

I disabled this `test_double_add_sub_external` test (https://github.com/runtimeverification/kontrol/pull/347/commits/0eacb173e2ba1fb99b96f0f77331bad7cab935a9) so the CI should pass, and we can either merge this PR now and ask Andrei or Petar to fix the `<code>` issue in a follow-up PR, or wait till they're back and can look into it in the scope of this PR.

src/tests/integration/test-data/foundry-dependency-skip
```diff
@@ -0,0 +1,2 @@
+ArithmeticCallTest.test_double_add()
+ArithmeticCallTest.test_double_add_sub_external(uint256,uint256,uint256)
```
ehildenb, line 2: Let's run these on CI.

src/kontrol/prove.py
```diff
@@ -71,6 +74,22 @@ def foundry_prove(
...
+        for test in test_suite:
+            if not isinstance(test.method, Contract.Method) or test.method.function_calls is None:
+                continue
+
+            test_version_tuples = [
+                parse_test_version_tuple(t) for t in test.method.function_calls if t not in summary_ids
+            ]
+
+            if len(test_version_tuples) > 0:
+                print(f'For test {test.name}, found external calls: {test_version_tuples}')
```
ehildenb, line 88: We should be logging this, not printing it. Printing is for outputting information we expect the next tool to consume.

```suggestion
                _LOGGER.info(f'For test {test.name}, found external calls: {test_version_tuples}')
```

src/tests/integration/test-data/foundry-dependency-skip
ehildenb: Or at least, let's run the first one, `test_double_add`, on CI. Then we'll have 2 tests running (`test_double_add` and `test_double_add_double_sub`).

src/kontrol/prove.py
```diff
@@ -71,6 +74,22 @@ def foundry_prove(
...
+                summary_ids.extend(
+                    p.id for p in foundry_prove(foundry, prove_options, rpc_options, test_version_tuples)
+                )
```
ehildenb, line 91: A major issue here is that we'll be proving all subproofs is sequence, instead of taking advantage of our parallel prover abilities. So this will be much slower than in could be because we're not dispatching subproofs in parallel.

ehildenb (approved):
I'm approving, but we need to quickly fix the parallelization issue. When manually including something with `--include-summary ...`, it's assumed the usre has already proven the subproof, so we can still get parallelization. But with `--cse` it just does the dependencies all one at a time instead of discharging them with the same number of workers that the user specified at the top level.

Please make an issue to fix this (in addition to addressing the other comments on the PR).

src/tests/integration/test-data/foundry-dependency-skip
anvacaru: There are currently 2 tests running. `test_double_add(uint256,uint256)` and `test_double_add_double_sub(uint256,uint256)`. `ArithmeticCallTest.test_double_add_sub_external(uint256,uint256,uint256)` is currently crashing with `Could not convert ML predicate to sort Bool`.

src/kontrol/prove.py
anvacaru: Here we could have something like
```py
        prove_option.workers = min(len(test_version_tuples), MAX_WORKERS)
```
where `MAX_WORKERS` can be defined to somewhere between 8 or 12.

palinatolmach (approved):

src/kontrol/prove.py
ehildenb: Ah, thanks for the explanation below, I think I understand better now.

src/kontrol/prove.py
```diff
@@ -73,6 +76,22 @@ def foundry_prove(
...
+
+            if len(test_version_tuples) > 0:
+                _LOGGER.info(f'For test {test.name}, found external calls: {test_version_tuples}')
+                summary_ids.extend(
+                    p.id for p in foundry_prove(foundry, prove_options, rpc_options, test_version_tuples)
+                )
+
     test_suite = collect_tests(foundry, tests, reinit=prove_options.reinit)
     test_names = [test.name for test in test_suite]
     print(f'Running functions: {test_names}')
```
ehildenb, line 97: This print should also be changed, but maybe in a later PR.

PR #361, opened 2024-02-12
Tolga/tomlize kontrol arguments

### LINKED ISSUE #73: Enable managing kontrolx options with a config file
It is very error prone to keep in mind what command you just kompiled with, which modules are imported, etc.

Instead of requiring the user to always kompile with certain flags, we should offer a configuration file (similar to how foundry does, which lets you set things like number of retries, etc). If you need to try using different flags here and there that can be managed with profiles.

This is also how Foundry does it, so it would be a recognizable pattern for devs.
(As a small bonus, seeing commited `kontrol.toml` files in Solidity git repos would work as good marketing, the same way you see `Dockerfile`s all over and that immediately tells you "oh this team is using Docker".)

So instead of for example

```
kontrolx foundry-kompile --require "aux-lemmas.k" --module-import PinataTest:AUX-LEMMAS --rekompile --with-llvm-library --verbose --regen
```

and then 

```
kontrolx foundry-prove --max-iterations 50 --verbose --no-break-on-calls --test PinataTest.testA0 --max-depth 1000
```

we could have a file `kontrol.toml`:

```toml
[profile.a0]
test = 'PinataTest.testA0'
max-depth = 1000

kompile-require = [ 'aux-lemmas.k' ]
kompile-module-import = [ 'PinataTest:AUX-LEMMAS' ]
kompile-llvm-library = true

prove-max-iterations = 50
prove-break-on-calls = false
```

This would likely reduce errors due to forgetting a certain flags, and improve reproducibility since every engineer should be able to get the same result by just doing 

```
kontrolx foundry-prove --profile a0
```

The best order of priority is probably

1. command-line flags
2. config file
3. defaults

### LINKED ISSUE #90: Good automatization of `kontrol prove` with no arguments
Similarly to `forge test`, we could have a more complete default behavior for `kontrol prove`. As of now, `kontrol prove` is a no-op.
It would be good if running `kontrol prove` did (at least some of) the following:

- [ ] If a `run-kontrol.sh` script is present, execute `bash run-kontrol.sh 2>&1 | tee log.out`
- [ ] If no `run-kontrol.sh` script is present, then run Kontrol with the default values on all `test`s present in the `.k.sol` files, while recording the output of the runs in a `log.out` file
- [ ] If a configuration mechanism is added similar to that of `foundry.toml`, `kontrol prove` would run with the configuration specified in such file, say `kontrol.toml`

While this might not be a complete description of the things we could add, it's a good starting point.

ovatman:
This PR targets being able to specify `kontrol` arguments via a toml file. Resolves #73 and resolves #90


tothtamas28 (commented):
This PR seems to address three more or less independent issues, please split it in three accordingly.

ehildenb:
Agreed, a 400-600 line diff is too much for reviewing in one PR. Better target is 100-200 lines.

Issue #334, opened 2024-02-05
Add flag to not create multiple versions of the same proof


JuanCoRo:
It is possible that when developing proofs, we don't care about keeping track of multiple versions of the same proof. We could have a flag (e.g., `--single-version`) to automatically delete all existing versions of `Contract.proof` when `kontrol prove Contract.proof --single-version` is executed.

palinatolmach:
Based on the conversation on the standup 02/06, we've decided to
- add this option as a configuration file parameter instead of a CLI flag
- prioritize the completion of https://github.com/runtimeverification/kontrol/issues/333 first

palinatolmach:
Since it was decided to add this option as a configuration file parameter rather than a CLI flag, I'm marking it as blocked on https://github.com/runtimeverification/kontrol/pull/373.

palinatolmach:
Closed in favor of https://github.com/orgs/runtimeverification/projects/23/views/7?pane=issue&itemId=72718013.

PR #373, opened 2024-02-15
adds the ability to accept arguments from a toml file provided as argument

### LINKED ISSUE #73: Enable managing kontrolx options with a config file

### LINKED ISSUE #90: Good automatization of `kontrol prove` with no arguments

ovatman:
This PR targets being able to specify kontrol arguments via a toml file. Resolves #73 and resolves #90
- moves `_create_argument_parser()` to `cli.py`
- adds `config` and `config-profile` arguments
- adds toml file to argument extractor function `read_toml_args()` to `cli.py`
- adds unit tests and test harness for the toml arguments


src/kontrol/__main__.py
```diff
@@ -78,6 +73,7 @@ def main() -> None:
     sys.setrecursionlimit(15000000)
     parser = _create_argument_parser()
     args = parser.parse_args()
+    args = read_toml_args(parser, args, sys.argv[2:]) if hasattr(args, 'config_file') else args
```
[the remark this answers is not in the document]
palinatolmach: Sounds good, thanks for the clarification!

src/kontrol/cli.py
```diff
@@ -110,3 +136,437 @@ def rpc_args(self) -> ArgumentParser:
...
+            return toml_profile[profile_list[0]]
+        return get_profile(toml_profile[profile_list[0]], profile_list[1:])
+
+    toml_args = {}
+    if args.config_file.is_file():
+        with open(args.config_file, 'rb') as config_file:
+            try:
+                toml_args = tomli.load(config_file)
+            except tomli.TOMLDecodeError:
+                _LOGGER.error('Input config file is not in TOML format')
```
palinatolmach, line 553: Since we're not raising a `ValueError`, the execution will continue, right? Maybe, in this case, you could explicitly state in the error message that, since the file is not valid TOML, it's being ignored/the default or CLI-provided values are to be used.

src/tests/unit/test_toml_args.py
```diff
@@ -0,0 +1,91 @@
...
+    assert hasattr(args, 'debug')
+    assert args.verbose
+    assert hasattr(args, 'version')
+    assert args.version == 3
+    assert hasattr(args, 'node_deltas')
+    assert len(args.node_deltas) == 1
+    assert str(args.node_deltas) == "[(10, '#target')]"
+
+
+def test_cli_overrides_toml() -> None:
```
palinatolmach, line 61: It's a great test to have! But I don't think `foundry_root` is being set in the `kontrol_test.toml` file, so I'm not sure if it's checking the priority of CLI vs config file-provided values:
```
[build]
verbose                 = false
debug                   = false
```

palinatolmach:
I left a few more comments, but other than that it looks good to me! @tothtamas28 could you please also look through this PR when you have time?

src/kontrol/__main__.py
tothtamas28, line 76: What's the reason behind using `sys.argv` directly?

tothtamas28: In this setup, I think it no longer makes sense to let `argparse` handle default values, i.e. all `default` arguments should be left unspecified. If you can in addition also ensure that `None` is not a valid value for any option (which I think should be straightforward), then you can do the following:

1. First, parse the command line options to get the config file location. Keep the `Namespace` object around for later.
2. Handler 1: Defaults. Create a `dict` with default values. Pass on the result to the next handler.
3. Handler 2: Config file. Parse and process the config, update the `dict`. Pass on the result to the next handler.
4. Handler 3: Command line arguments. Update the `dict` from the `Namespace`, skipping keys with value `None`.

Here, each handler takes a `dict` and updates it with values obtained from a given source. (Note that the order can also be reversed if you'd like, i.e. start with the source with the highest priority.)

palinatolmach: @tothtamas28 @ovatman do you think we can get this PR merged with some minimal changes, before @nwatson22  merges the `pyk` default value update (https://github.com/runtimeverification/pyk/pull/916) to make these changes easier to merge?

src/kontrol/__main__.py
```diff
@@ -80,6 +75,7 @@ def main() -> None:
...
```
nwatson22, line 78: Would it be possible for `read_toml_args` to return a `dict` of arguments read from the toml file instead of passing them again to argparse `parse_args` again along with the original command line args? So we could do something like
```
    args_dict = vars(parser.parse_args())
    args_dict = args_dict | read_toml_args()
    ...
    execute(args_dict)
```
This would ideally remove the necessity of having to convert the toml args to a command line format.

nwatson22 (commented):
Would it be possible for `read_toml_args` to return a `dict` of arguments read from the toml file instead of passing them again to argparse `parse_args` again along with the original command line args? So we could do something like
```

```

nwatson22:
@ovatman 

Yes, I see how that could be an issue now, with the naming inconsistencies, at least for the command-line-only user. Maybe we could add a command that automatically generates an example/base/stub toml file with all the settings visible in toml/python var form.

As for the checking for existence of file paths, I'm hoping that's not a big deal and we can handle the possibility that a file doesn't exist within the logic of the commands. Theoretically we should already be doing this, since you can just call `exec_*` programmatically and it will not do those argparse-level checks.

@tothtamas28 @ehildenb do you have any thoughts on these?

palinatolmach:
@ovatman @nwatson22 thanks for the investigation! I'd probably be in favor of
- mapping each option name to its designated variable name when reading toml contents to a dictionary, i.e., mapping `foundry_project_root` in toml to `foundry_root` automatically in `read_toml_args()`
- checking existence of files manually even if it's less elegant, since, as @nwatson22 noticed, we might be calling `exec_*` from within the Python code
- I also realised that we're using dashes in variable names as in CLI (`foundry-project-root`) instead of underscores (`foundry_project_root`) which [is more common](https://stackoverflow.com/questions/51543501/naming-convention-for-keys-in-toml-file); I don't have a strong opinion about it, so if you disagree please feel free to ignore.

palinatolmach:
@ovatman if we do `args_dict = read_toml_args() | args_dict`, then we'll be overriding values specified in `toml` with the default CLI values, right? I agree that's undesirable and that (manually provided) CLI arguments should take precedence — I'm not sure how to implement it best in this case, but maybe what @tothtamas28 had suggested in https://github.com/runtimeverification/kontrol/pull/373#discussion_r1494333609 might work? 

Also, would this logic change once @nwatson22's PR in pyk changing default values is ready?

palinatolmach:
@tothtamas28 @nwatson22 @ovatman 

Do you think we can merge this PR as is, if there's no obvious more elegant solution, but the incoming [`pyk` PR](https://github.com/runtimeverification/pyk/pull/916) (once ready) should make it easier to refactor, _and_ if it'll be easier for @nwatson22 to propagate his changes from that PR to Kontrol if this `toml` support PR is merged first?

yale-vinson:
@ovatman What is blocking this one? As this is blocking #334 is there a clear path to resolving both of these?

palinatolmach:
Closed as outdated since https://github.com/runtimeverification/kontrol/pull/580 got merged instead.

Issue #329, opened 2024-02-02
JUnit XML Test Reports


mds1:
JUnit XML test reports are widely used in CI tools to provide more advanced test reporting, such as [CircleCI's Test Insights](https://circleci.com/docs/insights-tests) feature. Having `kontrol prove` support this standard test result format would enable projects to track individual proof failure rates, average individual proof speed, etc., with existing services that support the standard.

anvacaru:
@mds1 Thank you for opening this issue. It would certainly be a nice addition! We'll start by looking at the JUnit XML schema and see what information is required and what information we already have available.

lisandrasilva:
Hi @mds1, 
Thanks a lot for this suggestion, we are working on implementing it. Currently the generated report look like the following:
```xml
<?xml version='1.0' encoding='utf-8'?>
<testsuites tests="4" failures="1" errors="1" time="194.60815978050232" timestamp="2024-02-16 15:07:00.831986">
	<testsuite name="SimpleTest" tests="2" failures="0" errors="0" time="65.24263954162598" timestamp="2024-02-16 15:07:34.434294">
		<testcase name="test_assert_true()" classname="SimpleTest" time="33.59451103210449" />
		<testcase name="testFail_assert_false()" classname="SimpleTest" time="31.648128509521484" />
	</testsuite>
	<testsuite name="AssertTest" tests="1" failures="1" errors="0" time="58.05429410934448" timestamp="2024-02-16 15:09:04.418438">
		<testcase name="test_revert_branch(uint256,uint256)" classname="AssertTest" time="58.05429410934448">
			<failure message="2 Failure nodes. (1 pending and 1 failing)">
Pending nodes: [9]

Failing nodes:

  Node id: 10
  Failure reason:
    Implication check failed, the following is the remaining implication:
    ( ( { true #Equals 0 &lt;=Int CALLER_ID:Int }
    #And ( { true #Equals 0 &lt;=Int ORIGIN_ID:Int }
    #And ( { true #Equals 0 &lt;=Int NUMBER_CELL:Int }
    #And ( { true #Equals 0 &lt;=Int TIMESTAMP_CELL:Int }
    #And ( { true #Equals 0 &lt;=Int VV0_x_114b9705:Int }
    #And ( { true #Equals 0 &lt;=Int VV1_y_114b9705:Int }
    #And ( { true #Equals VV1_y_114b9705:Int &lt;=Int VV0_x_114b9705:Int }
[...]

PR #377, opened 2024-02-20
Generate xml report

lisandrasilva:
This PR addresses issue #329.
When running `kontrol prove` with the flag `--xml-test-report` the report is generated. For now, the report only contains the tests that were passed in the `kontrol prove`.


src/tests/integration/test_foundry_prove.py
```diff
@@ -633,3 +634,38 @@ def test_deployment_summary(
         TEST_DATA_DIR / 'foundry' / 'src' / 'DeploymentStateCode.sol',
         update=update_expected_output,
     )
+
+
+def test_foundry_xml_report(
```
lisandrasilva, line 639: @ehildenb thank you for your suggestion! I just added the test. 

> Let's add a test to `test_foundry_prove.py` which exercises the new functionality.
> 
> * Pick a list of 3-5 short-running proofs. Run them in a batch with `foundry_prove(...)`.
> * Generate the XML report.
> * Re-load the XML report with a junit-style reader, to make sure it's understandable as an XML test report.

lisandrasilva:
This PR is dependent on evm-semantics [PR #2304](https://github.com/runtimeverification/evm-semantics/pull/2304).

src/tests/profiling/test_foundry_prove.py
```diff
@@ -28,27 +29,42 @@
...
+
+    tree = Et.parse('kontrol_prove_report.xml')
+    testsuites = tree.getroot()
+    testsuite = testsuites.find('testsuite')
+    assert testsuite
+    assert testsuite.get('name', 'None') == 'SimpleTest'
+    assert testsuite.findall('testcase[@name="test_assert_true()"]')
+    failure = testsuite.findall('testcase[@name="test_assert_false()"]')
+    assert failure
+    assert failure[0].findall('failure')
```
ehildenb, line 67: Is this intentional or should be reverted? It seems taht the test is added already to the integration test?

lisandrasilva: It was not intentional. I just reverted it.

src/kontrol/__main__.py
```diff
@@ -806,6 +808,13 @@ def _parse_test_version_tuple(value: str) -> tuple[str, int | None]:
         action='store_true',
         help='Flag used by Simbolik to initialise the state of a non test function as if it was a test function.',
     )
+    prove_args.add_argument(
+        '--xml-test-report',
+        dest='xml_test_report',
+        default=False,
+        action='store_true',
+        help='Generate a JUnit xml report',
```
palinatolmach, line 816: One very minor spelling suggestion: 

```suggestion
        help='Generate a JUnit XML report',
```

src/tests/integration/test_foundry_prove.py
palinatolmach: The test looks great, thanks @lisandrasilva! But it looks like the legacy backend job timed out on the previous run: https://github.com/runtimeverification/kontrol/actions/runs/8138402331/job/22251131183?pr=377. I'd say we can skip `test_foundry_xml_report` on legacy backend runs, e.g., you can just add
```
    if no_use_booster:
        pytest.skip()
```
to the test (you'll also need to add `no_use_booster: bool,` to the function signature).

palinatolmach (approved):
Approved with a few minor comments.

Issue #382, opened 2024-02-21
Refactor test selection if no `--match-test` is provided


palinatolmach:
The logic for selecting tests for verification if no `--match-test` is provided should be refactored to:
- include tests that start with `check` and `prove`, in addition to `test`, following a similar change in https://github.com/runtimeverification/kontrol/pull/249
- include test contracts that do not end with `Test` (e.g., our Optimism proof contract is called `OptimismPortalKontrol`), but, instead include the contracts located in the `test` folder

The refactoring should happen in the [all_tests function](https://github.com/runtimeverification/kontrol/blob/0d35ef478297f7d3004406f5ac951a0e5060d8e1/src/kontrol/foundry.py#L260):
 ```solidity
    @cached_property
    def all_tests(self) -> list[str]:
        return [
            f'{contract.name_with_path}.{method.signature}'
            for contract in self.contracts.values()
            if contract.name_with_path.endswith('Test')
            for method in contract.methods
            if method.name.startswith('test')
        ]
```

PR #383, opened 2024-02-21
Fix default test selection in the absence of`--match-test`

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/382.

This PR refactors the logic for selecting tests for verification if no `--match-test` is provided to:
- include tests that start with check and prove, in addition to test, following a similar change in https://github.com/runtimeverification/kontrol/pull/249
- instead of selecting test contracts ending with `Test`, verify contracts located in the `test` folder (by default) or in the one specified in `foundry.toml`, e.g.
```
test = 'test/kontrol/proofs'
```
This seems to be the approach Foundry is using to select tests, e.g., `forge test` executes tests located in `OptimismPortalKontrol` (not ending with `Test`), but doesn't analyze test contracts located in folders other than `test`:
```
❯ forge test
...
Running 3 tests for test/OptimismPortal.k.sol:OptimismPortalKontrol
[PASS] test_finalizeWithdrawalTransaction_paused(...) (runs: 256, μ: 81151, ~: 82169)
```
[Adding `/` ](https://github.com/runtimeverification/kontrol/blob/09d71f5b5b3d307592d430224443350c58dfe89a/src/kontrol/foundry.py#L262)to such test directory is needed to ensure that we're matching a folder (not just the `test` keyword) against the contract path.


src/kontrol/foundry.py
```diff
@@ -258,12 +258,14 @@ def build(self) -> None:
 
     @cached_property
     def all_tests(self) -> list[str]:
+        test_dir = self.profile.get('test', 'test')
+        test_dir += '/' if not test_dir.endswith('/') else ''
```
anvacaru, line 262: you could drop the `if` statement in favor of `os.path.join`:

```suggestion
        test_dir = os.path.join(self.profile.get('test', 'test'), '')
```

anvacaru (approved):
Approved with a single suggestion!

src/kontrol/foundry.py
palinatolmach: Nice, I didn't know about it, thank you!

Issue #389, opened 2024-02-22
Display the proof execution time


lisandrasilva:
When running `kontrol prove`, after the execution ends, it displays each proof's status. It would also be helpful to print the execution time of each proof. The same thing applies to the `kontrol list` command; it would be helpful to print the execution time of each proof.

PR #391, opened 2024-02-22
print proof execution time after proof status

lisandrasilva:
Addresses #389.


src/kontrol/__main__.py
```diff
@@ -305,9 +305,11 @@ def exec_prove(
     for proof in results:
         if proof.passed:
             print(f'PROOF PASSED: {proof.id}')
+            print(f'time: {proof.exec_time//3600}h {(proof.exec_time%3600)//60}m {proof.exec_time%60}s')
```
palinatolmach, line 308:
```suggestion
            print(f'time: {proof.exec_time // 3600}h {(proof.exec_time % 3600) // 60}m {proof.exec_time % 60}s')
```

palinatolmach: Also, since the same statement is repeated below, maybe it makes sense to factor it out in a separate function. This string
```
f'{proof.exec_time//3600}h {(proof.exec_time%3600)//60}m {proof.exec_time%60}s'
```
could also be a property of the proof object (e.g., `formatted_time`, or something like that), which would require another PR in pyk.

palinatolmach (approved):

src/kontrol/__main__.py
lisandrasilva: Good idea! I just opened this [PR](https://github.com/runtimeverification/pyk/pull/941) in pyk.

Issue #205, opened 2023-11-24
Test signatures are ignored when checking test versions


palinatolmach:
Identified and reported by ljmanini on Discord.

`foundry.py#proofs_with_test()` returns all proofs that contain the specified test's name in their file, ignoring the test's parameter types, which makes `kontrol` throw error messages suggesting that two versions of a test are available, while, in fact, there's a single version for each of two tests with the same name but different parameters. We should take full function signature into consideration when retrieving the number of available proof versions.

palinatolmach:
Thank you @lmanini for reporting this issue and looking into it! We had a somewhat similar conversation on the format of `kontrol prove --match-test` input (https://github.com/runtimeverification/kontrol/pull/103#issuecomment-1783404744) — at that time, we decided that requesting a user to provide a full signature with parentheses might hurt usability, so we made the help message for `--match-test` more specific, saying that "if a function is overloaded, one can provide a full signature, but otherwise all tests matching the regex or string prefix will be executed". 

Maybe we can add more clarification to the description provided in `kontrol view-kcfg` and the documentation? Do you think that might help? Right now, we are just saying 
```
positional arguments:
  test                  Test to run
```
and 
```
❯ kontrol view-kcfg
usage: kontrol view-kcfg [-h] [--version VERSION] [--verbose] [--debug] [--foundry-project-root FOUNDRY_ROOT] test
kontrol view-kcfg: error: the following arguments are required: test
```

yale-vinson:
@qian-hu is working this one, but we cannot assign to her until she is added to Kontol.

yale-vinson:
In discussion with @qian-hu and @anvacaru we decided that this issue is better dealt with by imporving our documentation as opposed to modifying the codebase. I am going to reassign this to @reekamaharaj and Qian will reach out to bring her up to speed.

One other recommendation from Andrei is to update the "help" message for the "view-kcfg" command ([here](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/__main__.py#L828)).

PR #394, opened 2024-02-23
Update kontrol view-kcfg error messages

qian-hu:
This PR improves error messages to assist the user in providing the correct proof for visualization. It also addresses issue #205 along the proof management section in the Kontrol documentation.

When executing `kontrol view-kcfg` and encountering multiple proofs with the same name, ensure the correct proof is provided by: 1) specifying the version number of the intended proof, and 2) ensuring that the function variables align with the values of the desired proof.


src/kontrol/foundry.py
```diff
@@ -309,9 +309,14 @@ def get_test_id(self, test: str, version: int | None) -> str:
         if len(matching_proof_ids) == 0:
             raise ValueError(f'Found no matching proofs for {test}:{version}.')
         if len(matching_proof_ids) > 1:
-            raise ValueError(
-                f'Found {len(matching_proof_ids)} matching proofs for {test}:{matching_proof_ids}. Use the --version flag to choose one.'
-            )
+            if version == None:
```
palinatolmach, line 312:
```suggestion
            if version is None:
```
I think it's recommended to use `is None` instead of `== None`, e.g., https://discuss.python.org/t/why-is-it-better-to-write-if-a-is-none-instead-of-if-a-none/42203

src/kontrol/foundry.py
```diff
@@ -309,9 +309,14 @@ def get_test_id(self, test: str, version: int | None) -> str:
...
+                raise ValueError(
+                    f'Found {len(matching_proof_ids)} matching proofs for {test}:{version}. Use the --version flag to choose one.'
+                )
+            else:
+                raise ValueError(
+                    f'Found {len(matching_proof_ids)} matching proofs for {test}:{version}. Provide full signatures to the test.'
```
palinatolmach, line 318: Do you think we can make the error message a bit more specific? E.g., provide an example of the full test signature (I copied this one from our `kontrol prove --match-test` help message. I'm just a bit worried that users may not really know what's the format of a full signature (plus, it has to be wrapped in '' symbols).
```suggestion
                    f"Found {len(matching_proof_ids)} matching proofs for {test}:{version}. Provide a full signature of the test, e.g., 'ERC20Test.testTransfer(address,uint256)'."
```

palinatolmach (commented):
I left some minor comments on the wording, but otherwise lgtm!

src/kontrol/foundry.py
palinatolmach: Looks great, thank you!

palinatolmach (approved):

Issue #395, opened 2024-02-26
Concretize `<account>`, `<id>`, `callDepth>` cells for debugger


palinatolmach:
After https://github.com/runtimeverification/kontrol/pull/313, `<account>` and some other relevant cells are being generated only for test functions:
https://github.com/runtimeverification/kontrol/blob/9bd00e49a19a2df0d15ab4b44ac2e7470296b99e/src/kontrol/prove.py#L578

As noted by @RaoulSchaffranek, debugger relies on `<account>`, `<id>`, `callDepth>` being concrete, so we should add a flag that generates these cells for _all_ functions. 

_Note:_ it's needed for the Simbolik demo, so we should try to get it merged by end of week.

Issue #362, opened 2024-02-12
Contracts and storage layouts in CSE executions


PetarMax:
All contracts relevant to a CSE execution need to be present in the storage explicitly, and their storage layout needs to be made explicit as well. Consider the (contrived, but representative) example from #356 of `Double.applyOp`:
```
contract Double is UIntUnaryOp {

    Identity iv;

    function applyOp(uint256 x) external view returns (uint256) {
        return iv.applyOp(x) + iv.applyOp(x);
    }
}
```
which holds an instance of an `Identity` contract in its storage and calls the `applyOp` function from it. There are two issues with the current generality of the CSE initial state:
1. The `<accessedStorage>` cell is fully symbolic, resulting in branching on whether or not various slots of various contracts have previously been accessed, together with an infinite branching similar to that reported in [this KEVM issue](https://github.com/runtimeverification/evm-semantics/issues/2291) due to the lack of full coverage of the `<accessedStorage>` update rules. This can be seen, for example, in nodes 4, 5, and 6, of the attached KCFG, which, respectively have the path conditions:
```
{ true #Equals ( notBool CONTRACT_ID:Int in_keys ( ACCESSEDSTORAGE_CELL:Map ) ) }
```
and
```
{ ACCESSEDSTORAGE_CELL:Map #Equals ( ( CONTRACT_ID:Int |-> TS:Set ) _DotVar7:Map ) }
```
and
```
{ true #Equals CONTRACT_ID:Int in_keys ( ACCESSEDSTORAGE_CELL:Map ) }
#And #Not ( #Exists _DotVar7:Map . #Exists TS:Set . ( { false #Equals CONTRACT_ID:Int in_keys ( _DotVar7:Map ) }
#And { ACCESSEDSTORAGE_CELL:Map #Equals ( ( CONTRACT_ID:Int |-> TS:Set )
_DotVar7:Map ) } ) )
```

2. The storage is fully symbolic, meaning that the knowledge that `iv` is an `Identity` contract is lost, and the execution cannot figure out which function corresponds to `applyOp`.

The KCFG can be obtained by running
```
kontrol build --require src/cse/lemmas.k --module-import CSETest:CSE-LEMMAS --verbose --regen --rekompile
kontrol prove --verbose --match-test 'Double.applyOp'
```
from the `src/tests/integration/test-data/foundry` folder of Kontrol using the `petar/cse-exploration` branch of PR #355.

[...]

PR #398, opened 2024-02-27
Allow non-general init state for non-test functions

anvacaru:
After https://github.com/runtimeverification/kontrol/pull/313/, the initial terms for non-test functions became more general in order to support CSE.
However, this is a setback for Symbolik, as it needs that information for debugging purposes.
In this PR I'm a `--with-non-general-state` flag to `kontrol prove` to unblock Symbolik until https://github.com/runtimeverification/kontrol/issues/362 gets implemented.

closes: https://github.com/runtimeverification/kontrol/issues/395


src/kontrol/__main__.py
```diff
@@ -797,6 +799,13 @@ def _parse_test_version_tuple(value: str) -> tuple[str, int | None]:
         action='append',
         help='Specify a summary to include as a lemma.',
     )
+    prove_args.add_argument(
+        '--with-non-general-state',
+        dest='with_non_general_state',
+        default=False,
+        action='store_true',
+        help='Flag used by Symbolik to initialise the state of a non test function as if it was a test function.',
```
palinatolmach, line 807:
```suggestion
        help='Flag used by Simbolik to initialise the state of a non test function as if it was a test function.',
```

palinatolmach (approved):
lgtm, thanks @anvacaru! One minor note is that I think it's spelled as Simbolik (with 'i'): [Simbolik](https://docs.runtimeverification.com/simbolik/).

PR #403, opened 2024-02-28
Adds `--hevm` success predicate option to `kontrol prove`

### LINKED ISSUE #312: Add `hevm_success` predicate for compatbility with HEVM
HEVM fails on `ds-test` and `assert` violations but not on `require` or `revert` statements. This is useful to silence some of the noise coming from overflow errors. Maybe we can add another success predicate, `hevm_success`, for enhanced tool compatibility.

Of course, we need to expose these different success conditions via some cli-argument. For example, we could add an argument `--success` that defaults to `foundry`, but also accepts `hevm`.

lisandrasilva:
Closes #312.


src/kontrol/prove.py
```diff
@@ -656,23 +661,31 @@ def _final_cterm(
...
+        foundry_success = Foundry.success(
+            KVariable('STATUSCODE_FINAL'),
+            dst_failed_post,
+            KVariable('ISREVERTEXPECTED_FINAL'),
+            KVariable('ISOPCODEEXPECTED_FINAL'),
+            KVariable('RECORDEVENT_FINAL'),
+            KVariable('ISEVENTEXPECTED_FINAL'),
+        )
+        if is_test:
+            if not failing:
```
palinatolmach, line 679: Can you please check if `proveFail` is supported by `hevm` for declaring tests that are supposed to fail, similarly to Foundry? I saw that they used this convention in some of their materials, but maybe something changed since then.

src/kontrol/__main__.py
```diff
@@ -806,6 +808,13 @@ def _parse_test_version_tuple(value: str) -> tuple[str, int | None]:
         action='store_true',
         help='Flag used by Simbolik to initialise the state of a non test function as if it was a test function.',
     )
+    prove_args.add_argument(
+        '--hevm',
+        dest='hevm',
+        default=False,
+        action='store_true',
+        help='Uses the hevm success predicate',
```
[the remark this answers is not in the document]
lisandrasilva: Something like "Runs the tests against the Hevm success predicate instead of the default Foundry's success criteria" is ok? Or would you suggest something else?

src/kontrol/kdist/foundry.md
```diff
@@ -86,3 +86,32 @@ module FOUNDRY-SUCCESS
 
 endmodule
 ```
+
+The hevm success predicate option was implemented for [benchmarking](https://github.com/eth-sc-comp/benchmarks/tree/deb3faa7e42993a057ba52935368a89f08970f19) purposes.
```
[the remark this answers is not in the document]
lisandrasilva: Yes, I am working on that.

src/kontrol/foundry.py
```diff
@@ -322,6 +322,10 @@ def success(s: KInner, dst: KInner, r: KInner, c: KInner, e1: KInner, e2: KInner
     def fail(s: KInner, dst: KInner, r: KInner, c: KInner, e1: KInner, e2: KInner) -> KApply:
         return notBool(Foundry.success(s, dst, r, c, e1, e2))
 
+    @staticmethod
+    def hevm_success(s: KInner, dst: KInner, out: KInner) -> KApply:
```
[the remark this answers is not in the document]
lisandrasilva: Done!

src/kontrol/prove.py
lisandrasilva: It seems it is not supported, I just tried with this simple example:
```solidity
function proveFail_simple2(uint x) public pure {
      require(x < 1);
      assert(x != 0);
```
And it fails with the command `hevm symbolic --code $BYTECODE --sig "proveFails_simple2(uint)"`:
```
Exploring contract
Simplifying expression
Explored contract (4 branches)
Checking for reachability of 1 potential property violation(s)

Discovered the following counterexamples:

Calldata:
  0x22a261690000000000000000000000000000000000000000000000000000000000000000

Storage:
  Addr SymAddr "miner": []
  Addr SymAddr "origin": []

Transaction Context:
  TxValue: 0x0
```

palinatolmach: Thanks for checking! Can you please also try `hevm test` and `halmos`? It looks like that might be our reference for what we're checking by default. ty!

src/kontrol/__main__.py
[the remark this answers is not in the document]
palinatolmach: How about "Use `hevm` success predicate instead of `foundry` to determine if a test is passing", or something like that?

src/kontrol/prove.py
lisandrasilva: Yes, `hevm test` supports `proveFail`. But also, `hevm test` fails with `revert`, see the test `prove_revert` and `prove_expectRevert`.

src/tests/integration/test-data/foundry/test/HevmTests.t.sol
```diff
@@ -0,0 +1,134 @@
...
+      assert(x == 0);
+    }
+
+    function prove_expectRevert() public {
+      // Passes with `hevm symbolic --code $BYTECODE --sig "prove_expectRevert()"`
+      // Fails with `hevm test` 
+      vm.expectRevert();
+    }
+
+    function prove_revert() public pure {
```
lisandrasilva, line 79: @palinatolmach `hevm test` fails with the `revert` statement, whereas `hevm symbolic` does not.

src/tests/integration/test-data/foundry/test/HevmTests.t.sol
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    function prove_assume2(uint x) public pure {
+      // Passes with `hevm symbolic --code $BYTECODE --sig "prove_assume2(uint)"`
+      // Passes with `hevm test`
+      vm.assume(x < 1);
+      assert(x == 0);
+    }
+
+    function prove_expectRevert() public {
```
lisandrasilva, line 73: @palinatolmach here the results between `hevm test` and `hevm symbolic` are also different.

src/kontrol/prove.py
palinatolmach: Thank you for confirming that! I just tried `proveFail` with Halmos, and it looks like it it's supported either. I think I'm slightly inclined to add support for `proveFail` to `hevm` success predicate checking (since `hevm test` supports this), but if don't find it reasonable, please feel free to not add it :) 
@RaoulSchaffranek can I ask for your opinion on that too? ty!

lisandrasilva: Maybe we can also ask dxo what they think? I don't mind to add it :) 
What about the tests `prove_revert` and `prove_expectRevert`? Should we also make the results consistent with `hevm test`? According to the issue #312, it should not fail with `revert` statements, however `hevm test` fails. 

Also, just to confirm, should we check the hevm success predicate only for functions prefixed with `prove`?

palinatolmach: My suggestion is to:
- not report a test as failing on a failed `revert` or `expectRevert` (i.e., to not make it consistent with `hevm test` on `prove_revert` and `prove_expectRevert`)
- check the hevm success predicate on tests that are prefixed with `prove`, `check`, and `test`, as we do for Foundry tests (we did this to make our results compatible with those of Halmos and hevm, so I'd stick to that)
- support `proveFail` as a way to specify tests that are supposed to fail

But I don't have a strong opinion and would also check with @RaoulSchaffranek  and/or dxo, yes :)

lisandrasilva: Implementing the `proveFail` is not that easy... For instance, the following test:
```solidity
function proveFail_require_assert(uint x) public pure {
      // Fails with `hevm symbolic --code $BYTECODE --sig "proveFail_require_assert(uint)"`
      // Passes with `hevm test`
      require(x < 1);
      assert(x != 0);
    }
```
Has two branches: 
1. `x < 1` -  hevm predicate returns `false` because of the user-defined assertion
2. `x >= 1` - hevm predicate returns `true`, although it ends with `EVMC_REVERT`

If we only make `proveFail` the symmetric to `prove`, as we do for Foundry, i.e.:
```python
if failing:
    return final_cterm.add_constraint(mlEqualsTrue(notBool hevm_success))
```
Then, the second branch will return `false`. We can work around this by implementing a different `hevm-fail` predicate that returns `true` if the function ends with `EVMC_REVERT`, but that would mean that the following tests would both pass, which would be inconsistent:
```solidity
function prove_revert() public pure {
      revert("Just reverts");
}

function proveFail_revert() public pure {
      revert("Just reverts");
}
```
Please let me know what you think. Or if you have a better solution :)

palinatolmach: Ok, I understood, thank you! I think `proveFail` tests should only fail if the current `hevm_success` predicate evaluates to `false`, e.g., if there's a failed user-defined or DSTest assertion. The problem here is that, in this case, the `require` statement plays a role of `vm.assume`, so, instead of branching on it, we should (ideally, in this `hevm`-like execution mode) just add it to the list of path conditions.

palinatolmach: I agree that's problematic, so maybe we can leave it for another PR. My only worry is that there'll be some discrepancy between our Foundry and hevm modes. Do you think it makes sense to add an info message for the user if `--hevm` is on and an `xFail` test is being analyzed?

lisandrasilva: I agree that would be the best implementation. I'll try to do that in this PR, it shouldn't be that hard. In case it is too complex, I'll go with the second option - add an info message for the `xFail` - and leave that for another PR.

RaoulSchaffranek: > The problem here is that, in this case, the require statement plays a role of vm.assume, so, instead of branching on it, we should (ideally, in this hevm-like execution mode) just add it to the list of path conditions.

I'm strongly in favor of this implementation. It allows us to write meaningful specifications in pure Solidity without the forge-std library - which can be very handy.

From the discussion it appears that `hevm` uses different success predicates for `hevm test` and `hevm symbolic`. And we're trying to simulate `hevm test`, is my understanding correct?

palinatolmach: > From the discussion it appears that `hevm` uses different success predicates for `hevm test` and `hevm symbolic`. And we're trying to simulate `hevm test`, is my understanding correct?

It looks like we're trying to simulate some middle ground between `hevm test` and `halmos` :). DSTest assert violations are supported by `hevm test`, so here we follow it, yes; but @lisandrasilva also mentioned that `hevm test` supports, for example, `expectRevert` statements, which we should ignore in the "user-defined tests and dstest asserts" success predicate (if I understand correctly).

src/kontrol/kdist/hevm.md
```diff
@@ -0,0 +1,40 @@
+### Hevm Success Predicate
```
palinatolmach, line 1: I know it's a nitpick, but it looks like `hevm` is spelled in all lowercase in their documentation and repo. Can I please ask you to use the same here and in the next line? Or, if you'd prefer, all uppercase `HEVM` might be fine too.
```suggestion
### hevm Success Predicate
```

src/kontrol/hevm.py
```diff
@@ -0,0 +1,34 @@
...
+        print_foundry_success_info = any('hevm_success' in line for line in res_lines)
+        if print_foundry_success_info:
+            res_lines.append('')
+            res_lines.append('See `hevm_success` predicate for more information:')
+            res_lines.append(
+                'https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/kdist/hevm.md#hevm-success-predicate'
+            )
+        res_lines.append('')
+        res_lines.append(
+            'Access documentation for KEVM foundry integration at https://docs.runtimeverification.com/kontrol'
```
palinatolmach, line 24: Sorry, another nitpick: I guess, it's not really Foundry integration if we're not using Foundry's success predicate? Maybe we can say "documentation for KEVM and hevm integration"? or just "documentation for Kontrol"? I think we can probably just switch the _"Access documentation for Kontrol at ..."_ for both `foundry` and `hevm` modes :) 
wdyt?

palinatolmach (commented):
Left a few comments, but the rest looks good to me!

palinatolmach:
@lisandrasilva did you get a chance to confirm that `hevm test` reports a `proveFail` test as passing if it reverts on all branches?

lisandrasilva:
> @lisandrasilva did you get a chance to confirm that `hevm test` reports a `proveFail` test as passing if it reverts on all branches?

@palinatolmach yes, the following test fails with `hevm test`:
```solidity
function proveFail_all_branches(uint x) public pure {
      // Fails with `hevm symbolic --code $BYTECODE --sig "proveFail_all_branches(uint)"`
      // Fails with `hevm test`
      // Fails with `kontrol prove --hevm --match-test proveFail_all_branches`
      if (x<10)
        assert(true);
      else
        assert(false);
    }
```
The tests file [HevmTests.t.sol](https://github.com/runtimeverification/kontrol/pull/403/files#diff-c7f405129ed7d36130e1ed224f15a40331d11cae34f5c42d1bfa16cb2ac28511) contains the results of running `hevm symbolic`, `hevm test` and `kontrol prove` for each test.

src/kontrol/kdist/hevm.md
```diff
@@ -0,0 +1,50 @@
...
+       andBool #range(OUT, 35, 1) ==K b"\x01" //Error code for user defined assertions
+    rule hevm_success(EVMC_REVERT, _, _) => true
+    rule hevm_success(_, _, _)           => false [owise]
+
+    rule ( selector ( "Panic(uint256)" ) => 1313373041 )
+```
+
+### hevm Fail Predicate
+
+In order to support `proveFail` we also defined the hevm fail predicate. This predicate asserts that all branches are failing, i.e. there is not a branch ending with `EVMC_SUCCESS` and not violating a `DS-TEST` assertion.
```
palinatolmach, line 35:
```suggestion
In order to support `proveFail` we also defined the hevm fail predicate. This predicate asserts that all branches of the test are failing, i.e. there is not a branch ending with `EVMC_SUCCESS` and not violating a `DS-TEST` assertion.
```

palinatolmach (approved):
I left a couple of minor spelling suggestions, everything else lgtm. @RaoulSchaffranek @anvacaru if you have time, can you please look through this PR too?

src/tests/integration/test_foundry_prove.py
```diff
@@ -733,3 +733,68 @@ def test_foundry_refute_node(
 
     # Proof passes again
     assert_pass(test, single(prove_res_3))
+
```
anvacaru, line 736: Consider moving this in a new `test_hevm_prove.py`.

anvacaru (approved):
Approved with a single comment.

src/kontrol/kdist/hevm.md
```diff
@@ -0,0 +1,50 @@
...
+        statusCode: StatusCode ","
+        failed: Int ","
+        output: Bytes
+      ")" [function, klabel(hevm_success), symbol]
+ // ----------------------------------------------
+    rule hevm_success(EVMC_SUCCESS, 0, _)  => true
+    rule hevm_success(EVMC_REVERT, _, OUT) => false
+      requires #range(OUT, 0, 4)  ==K Int2Bytes(4, selector ("Panic(uint256)"), BE)
+       andBool #range(OUT, 35, 1) ==K b"\x01" //Error code for user defined assertions
+    rule hevm_success(EVMC_REVERT, _, _) => true
```
lisandrasilva, line 27: @anvacaru I just noticed this. Shouldn't I have the `[owise]` here as well?

anvacaru: I would keep a single `[owise]` rule for `hevm_success(EVMC_REVERT, _, _)` and have a `requires` clause for the `true` case of `(EVMC_REVERT, _, OUT)`.
```k
    rule hevm_success(EVMC_SUCCESS, 0, _)  => true
    rule hevm_success(EVMC_REVERT, _, OUT) => true
      requires notBool(#range(OUT, 0, 4)  ==K Int2Bytes(4, selector ("Panic(uint256)"), BE)
       andBool #range(OUT, 35, 1) ==K b"\x01") //Error code for user defined assertions
    rule hevm_success(_, _, _)           => false [owise]
```

lisandrasilva: Thank you :)

src/tests/integration/test_foundry_prove.py
lisandrasilva: Done. I am also running the tests only with the booster since one of the tests was failing without `failure_info` in the legacy backend.

anvacaru (approved):

src/tests/integration/test_hevm_prove.py
```diff
@@ -0,0 +1,172 @@
...
+            copy_tree(str(TEST_DATA_DIR / 'foundry'), str(foundry_root))
+
+            run_process(['forge', 'install', '--no-git', f'foundry-rs/forge-std@{FORGE_STD_REF}'], cwd=foundry_root)
+            run_process(['forge', 'build'], cwd=foundry_root)
+
+            foundry_kompile(
+                foundry=Foundry(foundry_root),
+                includes=(),
+                requires=[str(TEST_DATA_DIR / 'lemmas.k')],
+                imports=['LoopsTest:SUM-TO-N-INVARIANT'],
```
palinatolmach, line 78: Since `LoopsTest` is not being run here, we probably don't need to import it. I wonder if we're rekompiling files for hevm tests, or are they being shared with Foundry tests?

palinatolmach:
As recommended by @tothtamas28, let's move the fixtures shared between `test_hevm_prove` and `test_foundry_prove` to a separate file and import them in `test_hevm_prove` and `test_foundry_prove`:

```
@pytest.fixture(scope='module')
def server(foundry: Foundry, no_use_booster: bool) -> Iterator[KoreServer]:
    llvm_definition_dir = foundry.out / 'kompiled' / 'llvm-library' if not no_use_booster else None
    kore_rpc_command = ('kore-rpc-booster',) if not no_use_booster else ('kore-rpc',)

    yield kore_server(
        definition_dir=foundry.kevm.definition_dir,
        llvm_definition_dir=llvm_definition_dir,
        module_name=foundry.kevm.main_module,
        command=kore_rpc_command,
        smt_timeout=300,
        smt_retry_limit=10,
        fallback_on=None,
        interim_simplification=None,
        no_post_exec_simplify=None,
    )


@pytest.fixture(scope='session')
def foundry(foundry_root_dir: Path | None, tmp_path_factory: TempPathFactory, worker_id: str) -> Foundry:
    if foundry_root_dir:
        return Foundry(foundry_root_dir)

    if worker_id == 'master':
        root_tmp_dir = tmp_path_factory.getbasetemp()
    else:
        root_tmp_dir = tmp_path_factory.getbasetemp().parent

    foundry_root = root_tmp_dir / 'foundry'
    with FileLock(str(foundry_root) + '.lock'):
        if not foundry_root.is_dir():
            copy_tree(str(TEST_DATA_DIR / 'foundry'), str(foundry_root))

            run_process(['forge', 'install', '--no-git', f'foundry-rs/forge-std@{FORGE_STD_REF}'], cwd=foundry_root)
            run_process(['forge', 'build'], cwd=foundry_root)

            foundry_kompile(
                foundry=Foundry(foundry_root),
                includes=(),
            )

    session_foundry_root = tmp_path_factory.mktemp('foundry')
    copy_tree(str(foundry_root), str(session_foundry_root))
    return Foundry(session_foundry_root)
```

tothtamas28:
> let's move the fixtures shared between `test_hevm_prove` and `test_foundry_prove` to a separate file

Specifically, to `src/tests/integration/conftest.py`. `pytest` automatically loads fixtures from there, no importing will be necessary,

src/tests/integration/test-data/show/foundry.k.expected
```diff
@@ -1,6 +1,5 @@
 requires "contracts.k"
 requires "requires/lemmas.k"
-requires "requires/cse-lemmas.k"
```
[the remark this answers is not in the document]
lisandrasilva: I got that by running `make test TEST_ARGS="--update-expected-output"`, but I am rerunning it since I just committed some changes.

src/kontrol/hevm.py
```diff
@@ -0,0 +1,38 @@
...
+if TYPE_CHECKING:
+    from pyk.kast.inner import KInner
+
+
+class Hevm:
+    @staticmethod
+    def help_info(proof_id: str) -> list[str]:
+        res_lines: list[str] = []
+        _, test = proof_id.split('.')
+        if not test.startswith('proveFail'):
```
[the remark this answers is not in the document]
lisandrasilva: It is not, thanks for noticing!

PR #355, opened 2024-02-09
Extending CSE support

PetarMax:
This PR provides infrastructure and tests in order to extend the support we have for CSE. In particular:

- it adds a number of CSE-related tests into a separate CI job; some are not run because of a bug in the booster that is to be fixed by [this PR](https://github.com/runtimeverification/hs-backend-booster/pull/544).
- To relieve pressure off of the CI, we are running only the CSE versions of the CSE-related tests and commenting out not the non-CSE ones. Eventually, both should be run since they will allow for a precise comparison of the effects of CSE.
- It adds a file with lemmas useful for CSE. These lemmas may be upstreamable to KEVM in subsequent PRs.

PR #404, opened 2024-02-29
Fix run-constructor option

### LINKED ISSUE #335: `--run-constructor` fails when run on a contract w/o constructor
Reported by `secoabla` on Discord.

`kontrol prove --run-constructor` throws the following error 
```
No such file or directory: 'out/proofs/test%SimpleTest.init:None/proof.json'
```
when run on a contract without a constructor, e.g.:
```
contract SimpleTest is Test {
    bool flag = true;
    function test_init() external {
        assert(flag);
    }
}
```

If the contract contains an empty constructor (`constructor() {}`), there's no error. 

We should handle this case gracefully and also ensure that the initialization code (e.g., storage variable initialization) is executed.

qian-hu:
Closes #335.
Addresses #66.

The current implementation requires using the end of the constructor's proof as the initial state for the test proof when the '--run-constructor' option is activated. However, an issue arises when this option is enabled while running a proof on a contract without a constructor - the initialization code fails to execute. In such cases, it disrupts the execution continuity as it lacks the end of the constructor's initial proof.

This PR proposes two potential solutions to address the abovementioned issue. The initial solution involves raising an error message to guide the user in providing a contract constructor. A more practical alternative suggests treating a contract without an explicit constructor as one with an empty constructor. This modification would also enable initializing the value assignments outside of functions in such cases, as in:
```
contract ConstructorTest is Test {
    bool flag = true;
}
```

The second approach is adopted to address issues of initializing variables.


nwatson22:
Thanks for fixing this. I would suggest the second option. Right now the `Contract.constructor` field is set in `Contract.__init__` only if there is an explicit contract member where `type='constructor'`. It may be necessary to modify the `Constructor` class to be able to construct one not tied to a method, and it should still use the `init_bytecode` rather than the `bin_runtime` by the check in `_method_to_cfg`.

palinatolmach:
@PetarMax I think it's supported and checked in [ContructorTest](https://github.com/runtimeverification/kontrol/blob/6fb37c90d4dd2761b881a09b8e7e0e0566feb1f1/src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol):
```solidity
pragma solidity =0.8.13;

import "forge-std/Test.sol";

contract ConstructorTest is Test {
    bool flag = true;

    function test_constructor() public {
        assert(flag);
    }

    function testFail_constructor() public {
        assert(!flag);
    }
}
```
Or do you mean something else?

src/tests/integration/test-data/foundry-prove-skip-legacy
```diff
@@ -60,6 +60,7 @@ BytesTypeTest.testFail_bytes32(bytes32)
 BytesTypeTest.testFail_bytes4(bytes4)
 ChainIdTest.test_chainid_setup()
 CoinBaseTest.test_coinbase_setup()
+ConstructorTest.test_constructor()
```
palinatolmach, line 63: Thanks for adding this test! For it to be run on CI (in this case, only with the booster backend), it should also be added to [foundry-prove-all](https://github.com/runtimeverification/kontrol/blob/master/src/tests/integration/test-data/foundry-prove-all).

PetarMax:
> @PetarMax I think it's supported and checked in [ContructorTest](https://github.com/runtimeverification/kontrol/blob/6fb37c90d4dd2761b881a09b8e7e0e0566feb1f1/src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol):
> 
> ```solidity
> // SPDX-License-Identifier: UNLICENSED
> pragma solidity =0.8.13;
> 
> import "forge-std/Test.sol";
> 
> contract ConstructorTest is Test {
>     bool flag = true;
> 
>     function test_constructor() public {
>         assert(flag);
>     }
> 
>     function testFail_constructor() public {
>         assert(!flag);
>     }
> }
> ```
> 
> Or do you mean something else?

Yes, that's what I meant - does it also work for analysing non-tests (i.e., functions in isolation)?

ehildenb:
Let's add a test of the non-test function being used correctly with eth `--run-constructor` option as well, as that is a feature that has been requested and we want to make sure it is working.

ehildenb:
We need a new testing harness in `test_foundry_prove.py` (https://github.com/runtimeverification/kontrol/blob/master/src/tests/integration/test_foundry_prove.py), called `test_run_constructor`. You can set that test up to be the same as `test_foundry_prove(...)` function, but when it builds the `ProveOptions`, it passes `run_constructor=True`. Then add a new list at `tests/integration/test-data/foundry-run-constructor`, which lists at least 2 functions, one which is a `test_` function, one which is a normal non-test function, to be run in this test-suite. Make sure that we're getitng the correct behavior on both functions.

ehildenb:
@qian-hu I think you'll need to update the expected output, by running:

```
make test-integration TEST_ARGS='--update-expected-output'
make test-integration TEST_ARGS='--update-expected-output --use-booster'
```

That will take a long time to run, and should probably be run on a powerful machine.

palinatolmach:
@ehildenb @qian-hu maybe `test_run_constructor` can be merged with an existing [`test_foundry_init_code` harness](https://github.com/runtimeverification/kontrol/blob/d2a5828f4a68d1bdbe7b21be03942a59eb765be1/src/tests/integration/test_foundry_prove.py#L602C5-L602C27) that is run with `run_constructor=True`? 

Also, starting from https://github.com/runtimeverification/kontrol/pull/249, functions prefixed with `check` and `prove` are treated similarly to those starting with `test`, so they are also tests. Sorry for not mentioning that earlier. For testing whether variables are getting initialized in non-tests, can you please rename it in a way that it doesn't start with either of these three keywords, e.g., `run_constructor()`? Thanks!

palinatolmach:
@qian-hu  starting from https://github.com/runtimeverification/kontrol/pull/249, functions prefixed with `check` and `prove` are treated similarly to those starting with `test`, so they are also tests. Sorry for not mentioning that earlier. For testing whether variables are getting initialized in non-tests, can you please rename it in a way that it doesn't start with either of these three keywords, e.g., `run_constructor()`? Thanks!

ehildenb:
I think we still want to run those proofs with `run_constructor=False` to make sure that works? Or is it expected that those proofs fail with `run_constructor=False`? The idea is we should test both functionalities, but if it's not expected to work we can skip testing.

palinatolmach:
@ehildenb the tests are designed in way that they'll fail if initialization code is not executed, e.g., 
```
contract ConstructorTest is Test {
    bool flag = true;

    function test_constructor() public {
        assert(flag);
    }
```
It's expected to fail w/o `--run-constructor` since in this case `flag` will have the default value of `false`. In `test_foundry_prove` tests almost all (if not all) have no constructor and they're not failing when run with `run_constructor=False`, so in a sense that's being test too. But it should be possible to add more tests, if you think that's better.

src/tests/integration/test-data/foundry-prove-skip-legacy
```diff
@@ -62,6 +62,8 @@ BytesTypeTest.testFail_bytes32(bytes32)
...
+ConstructorTest.run_constructor()
```
palinatolmach, line 66: This `foundry-prove-skip-legacy` file is [used](https://github.com/runtimeverification/kontrol/blob/134b0a407385b769727bc55f291c3164782ee119/src/tests/integration/test_foundry_prove.py#L147) to determine which tests to not run with legacy backend in `test_foundry_prove`, and is not used in `test_foundry_init_code`. 

At the same time, all [`test_foundry_init_code`](https://github.com/runtimeverification/kontrol/blob/c1104375d608b20496a2c94435a34f82a16e28c9/src/tests/integration/test_foundry_prove.py#L576
) tests are being skipped in the legacy backend job:
```
def test_foundry_init_code(test: str, foundry: Foundry, bug_report: BugReport | None, no_use_booster: bool) -> None:
    if no_use_booster:
        pytest.skip()
```
If that's intentional, you can just remove these `ConstructorTest` tests from `foundry-prove-skip-legacy`. If you'd prefer to run some of the tests on the legacy backend (which might be reasonable if they aren't taking more than 300s to execute there — that's the threshold we currently use), you can do something similar to
```
    if (no_use_booster and test_id in SKIPPED_LEGACY_INIT_CODE_TESTS)):
        pytest.skip()
```
where `SKIPPED_LEGACY_INIT_CODE_TESTS` can be read from a file such as `foundry-init-code-skip`.

src/tests/integration/test-data/foundry-init-code-tests
```diff
@@ -0,0 +1,6 @@
+InitCodeTest.test_init()
```
palinatolmach, line 1: I know that's nitpicking, but to be consistent with other similar files I'd suggest to rename `foundry-init-code-tests`  to `foundry-init-code` or `foundry-init-code-all` (the latter is useful if there'll be a file with tests we'd like to skip such as `foundry-init-code-skip`).

palinatolmach:
@PetarMax do we make any assumptions on storage being symbolic in non-tests? If the init code is executed, the `flag` in the contract above be `true` even for non-tests, right?

palinatolmach:
@PetarMax yes, generally, it's being inlined in the runtime bytecode if the variable is marked as `constant` and `immutable` (immutable variables can be initialized during declaration _or_ in the constructor, so I'm not certain if they'll be inlined in the latter case during symbolic execution). That sounds reasonable, and I think we can do some static analysis on the AST we use as input — e.g., we can determine if there are assignments to the variable outside of the declaration or a constructor. But it should probably be another PR :) 

Do you know if right now we're assuming that all storage in non-tests is symbolic? I'm wondering if the `flag`'s value is not being set or gets overwritten.

PetarMax:
I would think it's not being set, since we don't do any explicit storage manipulation, it is currently just set as symbolic.

palinatolmach:
Actually, @PetarMax @qian-hu, my apologies — I just re-ran this example and it _is_ being set (I'm just not used to [view-kcfg](https://github.com/runtimeverification/kontrol/pull/406) loading the latest version by default, sorry). But it still branches on `CALLVALUE_CELL` being zero (is it normal?), and it doesn't fail on a revert, so I'm not sure how can a concise test for non-test functions look like.

palinatolmach:
Besides, something strange still happens even though I see that in the `<storage>` cell slot 27 corresponding to `flag` is set to `1`. The proof fails with this failure info:
```
  Node id: 18
  Failure reason:
    Matching failed.
    The following cells failed matching individually (antecedent #Implies consequent):
    K_CELL: SLOAD 27
    ~> #pc [ SLOAD ]
    ~> #execute #Implies #halt
    ~> CONTINUATION:K
  Path condition:
    { true #Equals notBool CALLVALUE_CELL:Int ==Int 0 }
  Model:
    TIMESTAMP_CELL = 0
    NUMBER_CELL = 0
    ID_CELL = 0
    ORIGIN_ID = 10
    CALLER_ID = 10
```

And looks like it can resolve `SLOAD 27` to `1`:
```
┃  │  (181 steps)
┃  ├─ 11
┃  │   k: SLOAD 27 ~> #pc [ SLOAD ] ~> #execute ~> CONTINUATION:K
┃  │   pc: 1890
┃  │   callDepth: CALLDEPTH_CELL:Int
┃  ┃
┃  ┃ (1 step)
┃  ┣━━┓
┃  ┃  │
┃  ┃  ├─ 13
┃  ┃  │   k: 1 ~> #push ~> #pc [ SLOAD ] ~> #execute ~> CONTINUATION:K
┃  ┃  │   pc: 1890
┃  ┃  │   callDepth: CALLDEPTH_CELL:Int
┃  ┃  │   statusCode: STATUSCODE:StatusCode
┃  ┃  │
...
┃  ┃
┃  ┣━━┓
┃  ┃  │
┃  ┃  ├─ 14
┃  ┃  │   k: 0 ~> #push ~> #pc [ SLOAD ] ~> #execute ~> CONTINUATION:K
┃  ┃  │   pc: 1890
┃  ┃  │   callDepth: CALLDEPTH_CELL:Int
┃  ┃  │   statusCode: STATUSCODE:StatusCode
┃  ┃  │
┃  ┃  │  (83 steps)
┃  ┃  ├─ 20 (terminal)
┃  ┃  │   k: #halt ~> CONTINUATION:K
┃  ┃  │   pc: 2527
┃  ┃  │   callDepth: CALLDEPTH_CELL:Int
┃  ┃  │   statusCode: EVMC_REVERT
```
That might be a problem that's not directly related to this PR though.

PetarMax:
Yes, if the function is not a test, `CALLVALUE` is symbolic and the execution may branch if the function in question is not `payable` (I think).

PetarMax:
There is only one rule for `SLOAD`:
```
    rule [sload]:
         <k> SLOAD INDEX => #lookup(STORAGE, INDEX) ~> #push ... </k>
         <id> ACCT </id>
         <account>
           <acctID> ACCT </acctID>
           <storage> STORAGE </storage>
           ...
         </account>
```
and that rule requires `ACCT` to be in accounts. What is the `<id>` and does it exist in accounts? I think we might need to merge [this PR first](https://github.com/runtimeverification/kontrol/pull/355) for that functionality to work.

palinatolmach:
> What is the id and does it exist in accounts? 

Ok, that explains it, thanks @PetarMax! `<id>` is `ID_CELL:Int` and it doesn't exist in `<accounts>`. Shouldn't it be the same as this Foundry test getting executed? 

But, in any case, if the support provided by this PR is sufficient, maybe we can get it merged without the test for non-tests, and instead add it as part of the CSE PR?

PetarMax:
I've now extracted that part of the CSE PR [into a separate PR](https://github.com/runtimeverification/kontrol/pull/427), which could ostensibly get merged much faster.

palinatolmach:
Thanks @PetarMax and @qian-hu! It all looks good to me now; my only remaining request is to a non-test, now that it won't crash on `SLOAD`. We can't test it as a `test`-prefixed function since it won't fail, but @PetarMax suggested that we can execute something like
```
    function run_constructor() public {
        assert(flag);
    }
```
and assert that — as `flag` gets set to `1` — in the KCFG, there are no execution paths that end with the `EVMC_REVERT` status. That might be similar to how we perform [`foundry_show`](https://github.com/runtimeverification/kontrol/blob/2c4ce81d2b17b845536cd24891dd7715f3a5cb1a/src/tests/integration/test_foundry_prove.py#L173) and then [compare expected output](https://github.com/runtimeverification/kontrol/blob/2c4ce81d2b17b845536cd24891dd7715f3a5cb1a/src/tests/integration/test_foundry_prove.py#L187) in `test_foundry_prove`.

src/tests/integration/test_foundry_prove.py
```diff
@@ -569,15 +569,68 @@ def test_foundry_resume_proof(
...
+    if test_id != test:
+        assert_pass(test_id, single(prove_res_1))
+        return
+    else:
+        assert_fail(test, single(prove_res_1))
+
+    # Continue checking non-test function
+    check_pending(foundry, test, [19, 20, 21])
+
+    foundry_step_node(
```
palinatolmach, line 613: I think the execution of `ConstructorTest.run_constructor()` terminates early due to `fail_fast` being turned on by default. You can pass `fail_fast=False` to the first `foundry_prove`:
```py
    prove_res_1 = foundry_prove(
        foundry,
        tests=[(test_id, None)],
        prove_options=ProveOptions(
            run_constructor=True,
            bug_report=bug_report,
            fail_fast=False,
        ),
        rpc_options=RPCOptions(
            smt_timeout=300,
            smt_retry_limit=10,
            use_booster=not no_use_booster,
        ),
    )
```
which shouldn't affect other `init_code` tests since they are passing. The test can then look like this:
```py
@pytest.mark.parametrize('test_id', ALL_INIT_CODE_TESTS)
def test_foundry_init_code(
    test_id: str,
    foundry: Foundry,
    update_expected_output: bool,
    bug_report: BugReport | None,
    no_use_booster: bool,
    server: KoreServer,
) -> None:
    if no_use_booster:
        pytest.skip()

    test = 'ConstructorTest.run_constructor()'

    prove_res_1 = foundry_prove(
        foundry,
        tests=[(test_id, None)],
        prove_options=ProveOptions(
            run_constructor=True,
            bug_report=bug_report,
            fail_fast=False
        ),
        rpc_options=RPCOptions(
            smt_timeout=300,
            smt_retry_limit=10,
            use_booster=not no_use_booster,
        ),
    )

    # Then
    if test_id != test:
        assert_pass(test_id, single(prove_res_1))
        return
    else:
        assert_fail(test, single(prove_res_1))

    # And when
        show_res = foundry_show(
            foundry,
            test=test,
            to_module=True,
            sort_collections=True,
            omit_unstable_output=True,
            pending=True,
            failing=True,
            failure_info=True,
            counterexample_info=True,
            port=server.port,
        )

        # Then
        assert_or_update_show_output(show_res, TEST_DATA_DIR / f'show/{test}.expected', update=update_expected_output)
```

src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol
```diff
@@ -0,0 +1,43 @@
...
+
+    function test_constructor() public {
+        assert(flag);
+    }
+
+    function testFail_constructor() public {
+        assert(!flag);
+    }
+
+    function test_run_constructor_false() public {
```
palinatolmach, line 17: I feel like there's some redundancy in these tests. For example, `test_run_constructor_false()` now asserts that a variable set to `false` is `false`, and it doesn't really test the init code functionality, so I think it (and `testFail_run_constructor_false`) can be safely removed. `testFail_run_constructor` and `test_run_constructor` test the same functionality as `test_constructor` and `testFail_constructor` except for the getter function that I feel like is out of scope of this change and PR. 

@qian-hu @ehildenb I think having `test_constructor()`, `run_constructor()`, and maybe `testFail_constructor()` would be sufficient, but if you disagree please feel free to leave it as is.

src/tests/integration/test-data/foundry-prove-all
```diff
@@ -71,6 +71,9 @@ BytesTypeTest.testFail_bytes32(bytes32)
 BytesTypeTest.testFail_bytes4(bytes4)
 ChainIdTest.test_chainid_setup()
 CoinBaseTest.test_coinbase_setup()
+ConstructorTest.test_run_constructor_false()
+ConstructorTest.testFail_run_constructor_false()
+ConstructorTest.run_constructor()
```
palinatolmach, line 76: Would `run_constructor()` work and have the same output when initialization code is not executed?

palinatolmach (approved):
I approved this PR to unblock, but left several suggestions on tests.

src/tests/integration/test_foundry_prove.py
palinatolmach: That would requite updating expected output, although it's almost the same.

palinatolmach:
@qian-hu the PR looks great, many thanks for addressing all the comments!

PR #412, opened 2024-03-05
Implement `kontrol split-node` command

lucasmt:
This PR implements `kontrol split-node`, which allows manually splitting a node on a user-provided branching condition. This feature is useful, for example, for introducing assumptions without having to restart a proof, by splitting a node on the assumption and then refuting the negative branch.

A test for the new feature is also included, along with its expected output.


src/kontrol/__main__.py
```diff
@@ -375,6 +376,16 @@ def exec_unrefute_node(foundry_root: Path, test: str, node: NodeIdLike, version:
...
 
 
+def exec_split_node(
+    foundry_root: Path, test: str, node: NodeIdLike, branch_condition: str, version: int | None, **kwargs: Any
+) -> None:
+    node_ids = foundry_split_node(
+        foundry=_load_foundry(foundry_root), test=test, node=node, branch_condition=branch_condition, version=version
+    )
+
+    print(f'Node {node} has been split into {node_ids} on condition {branch_condition}.')
```
anvacaru, line 386: Seems like this message is duplicated. It's present here and in `foundry_split_node`. Is this intended?

PR #423, opened 2024-03-08
Print link to `foundry_success` predicate in `help_info`

### LINKED ISSUE #414: `Foundry.help_info()` does not print link to the `foundry_success` predicate
The function `Foundry.help_info()` has an `if` condition that always evaluates to `false`. The link to the `foundry_success` predicate is added to the `res_lines` inside this `if` block, which means it is never printed to the user. 

https://github.com/runtimeverification/kontrol/blob/dd539214b22103cceb4aec00cc85055c23b5abd1/src/kontrol/foundry.py#L367-L380

lisandrasilva:
Fixes #414.


src/tests/integration/test-data/show/AssertTest.checkFail_assert_false().expected
```diff
@@ -263,8 +263,8 @@ module SUMMARY-TEST%ASSERTTEST.CHECKFAIL-ASSERT-FALSE():0
        andBool ( ORIGIN_ID:Int <Int pow160
        andBool ( NUMBER_CELL:Int <=Int maxSInt256
        andBool ( TIMESTAMP_CELL:Int <Int pow256
-       andBool ( ( notBool ( 0 <Int CALLER_ID:Int andBool CALLER_ID:Int <=Int 9 ) )
```
palinatolmach, line 266: I'm a bit surprised by this change in the expected output for this PR that doesn't include any reasoning changes (although that diff probably makes sense?), but let's see if the CI passes.

palinatolmach (approved):
Approved with a note on the expected output update!

src/tests/integration/test-data/show/AssertTest.checkFail_assert_false().expected
lisandrasilva: Yes, I was also surprised by this change. I just ran `make test-integration TEST_ARGS="--numprocesses=4 --update-expected-output"`. But I'll rerun it.

palinatolmach: Are you running the same `pyk`/`kevm-pyk` version as the current Kontrol master?

lisandrasilva: But I just merged `master` into the hevm PR, and I am getting the same changes when running the tests locally with `--update-expected-output` flag.

palinatolmach: Yeah, that looks like the right version. The tests we're checking output for started failing, I think: https://github.com/runtimeverification/kontrol/actions/runs/8201372467/job/22430073352?pr=423. Let me try to run it locally too!

lisandrasilva: I know, it's weird.  I'll clean my `.cache` and run them again. I had problems before with updating the expected files because I had an older version of KEVM in `.cache`.

palinatolmach: It might be the cache issue, cause I don't get this rewrite. Shall I push the output now to make this PR pass, or would you prefer to do it yourself?

lisandrasilva: Ahh ok, thank you! Yes, if you don't mind pushing it it would be great, otherwise I can push it later :)

lisandrasilva: Which version of k do you have installed with `kup`? I cleaned the cache and I still get the same changes :woman_shrugging:

lisandrasilva: I am running the following sequence of commands:
```
kup install k.openssl.procps --version v$(cat deps/k_release)
poetry install
poetry run kdist clean
CXX=clang++-14 poetry run kdist --verbose build -j2 evm-semantics.haskell kontrol.foundry
make test-integration TEST_ARGS="--numprocesses=4 --update-expected-output"
```

palinatolmach: Oh good question! I have `v6.3.24` installed, and it's actually a bit outdated — it looks like `6.3.25` is the one we have in `deps` now.

lisandrasilva: Yes, that's the one I am using. But it's weird because you are right, the tests fail on CI because of these changes! 
Well, it seems I am disabled from updating the expected files again :slightly_smiling_face:

palinatolmach: Can you please send a message to the kontrol channel, in case someone encountered that before? 

Also, all tests are passing now expect for `AssumeTest.test_assume_false` — I think that's caused by the `failure_info` issue (I have it locally too) that'll get fixed once the `pyk` version is updated in KEVM (https://github.com/runtimeverification/evm-semantics/pull/2341) and then lands in Kontrol.

palinatolmach:
Looks like the output got outdated, will rerun it.

palinatolmach:
> Looks like the output got outdated, will regenerate it.

Actually, I'm still getting the same output — @lisandrasilva can you please check what's the reason the tests are failing now? Are you still getting the same `contracts.k` and `foundry.k` files locally? ty!

lisandrasilva:
I just noticed that I am building with:
```
CXX=clang++-14 poetry run kdist --verbose build -j2 evm-semantics.haskell kontrol.foundry
```
That can be the reason why I get those changes in the output.

palinatolmach (approved):

PR #424, opened 2024-03-08
Fix failing case of expectEmit

anvacaru:
While running a foundry test of Solady with Kontrol, I discovered a failing case of `vm.expectEmit` cheat code.
I added a minimal example `testExpectEmitLessTopics` that fails on master, passes with `forge`, and passes with this branch.
The problem was for the case in which we mark `N` topics to check ( i.e., all 4 of them), but the emitted events have less than `N` (i.e. 3) topics (the topics are the same).

I also modified the `LOG(N)` rules in cheatcode.md a bit and added a description.


anvacaru:
In addition, seems that Foundry has updated the logic for `vm.expectEmit()` ([link](https://book.getfoundry.sh/cheatcodes/expect-emit?highlight=expectEmit#expectemit)).
The main differences that I see are:
    - the expected event is checked only for the next call.
    - you can now expect multiple events

src/tests/integration/test-data/foundry/test/EmitContractTest.t.sol
```diff
@@ -36,4 +36,11 @@ contract EmitContractTest is Test {
         // The event we get
         emitter.t();
     }
+
+    function testExpectEmitLessTopics() public {
+        ExpectEmit emitter = new ExpectEmit();
+        vm.expectEmit(true, true, true, false);
+        emit Transfer(address(this), address(1337), 1337);
+        emitter.t();
```
palinatolmach, line 44: Doesn't `t()` emit the event with the same number of topics? 
```
    function t() public {
        emit Transfer(msg.sender, address(1337), 1337);
    }
```

palinatolmach (commented):
Nice catch! I just left one question about the test, and would also wait for @PetarMax's approval.

src/tests/integration/test-data/foundry/test/EmitContractTest.t.sol
anvacaru: Yes, it does, but in `vm.expectEmit(true, true, true, false);` we are saying to compare all 4 topics. The `#checkTopics` returned false previously if the topic list was empty but we still wanted to check them.

PetarMax (commented):
Two relatively simple comments left to address. I think the replication of the `LOG` functionality rather than delegation is ok (and it does fix the issue). The conflation of the two rules, I think, should have been done regardless.

src/kontrol/kdist/cheatcodes.md
```diff
@@ -799,48 +799,47 @@ With address: Asserts the topics match and that the emitting address matches.
       requires SELECTOR ==Int selector ( "expectEmit(bool,bool,bool,bool,address)" )
 ```
 
+The `foundry.recordExpectedEvent` is used to record the event that is emitted following the `expectEmit` cheat code.
+When the `LOG(N)` OpCode is executed, a new `SubstateLogItem` will be stored in the `<log>` cell, including the `N` topics that are fetched from the `<wordStack>`.
+
 ```k
-    rule <k> LOG(N) _MEMSTART _MEMWIDTH ... </k>
+    rule [foundry.recordExpectedEvent]:
```
PetarMax, line 806: Could we please have a comment here that signals that should the `LOG` functionality change in KEVM, this logging should mirror that change?

PetarMax (approved):

PR #428, opened 2024-03-08
Fix `xml-test-report`: allow proofs to be in nested directories

JuanCoRo:
Currently the `--xml-test-report` flag will throw an error if the proofs are located in a nested directory. This is due to the line `_, contract_name = contract.split('%')` in the `foundry_to_xml` function. Replacing it with `_, contract_name = contract.rsplit('%', 1)` should solve the problem.


palinatolmach:
Thanks @JuanCoRo! It looks good, but do you think you can add a test for it? For example, you can add another folder with a test to [test-data/foundry-test](https://github.com/runtimeverification/kontrol/tree/master/src/tests/integration/test-data/foundry/test) and include it in [`test_foundry_xml_report`](https://github.com/runtimeverification/kontrol/blob/134b0a407385b769727bc55f291c3164782ee119/src/tests/integration/test_foundry_prove.py#L742).

src/tests/integration/test_foundry_prove.py
```diff
@@ -757,6 +757,7 @@ def test_foundry_xml_report(
         tests=[
             ('AssertTest.test_assert_true()', None),
             ('AssertTest.test_assert_false()', None),
+            ('AssertNestedTest.test_assert_true_nested()', None),
         ],
         prove_options=ProveOptions(
```
lisandrasilva, line 762: @palinatolmach I think we can remove the `reinit` flag now. We can do that in this PR.

src/kontrol/foundry.py
```diff
@@ -741,7 +741,7 @@ def foundry_to_xml(foundry: Foundry, proofs: list[APRProof]) -> None:
         tests += 1
         test, *_ = proof.id.split(':')
         contract, test_name = test.split('.')
-        _, contract_name = contract.split('%')
+        _, contract_name = contract.rsplit('%', 1)
```
lisandrasilva, line 744: Since processing the `proof.id` or `test.id` to get the contract name and the test name is being done more than once and probably will be needed in upcoming fixes, I would suggest refactoring this into a separate function:
```python
def get_contract_test_names(id: str):
    test, *_ = id.split(':')
    contract, test_name = test.split('.')
    _, contract_name = contract.split('%')
    return contract_name, test_name
```

lisandrasilva (commented):
LGTM! Thank you, Juan!
I just left a suggestion that can be addressed in this PR or an upcoming one.

src/kontrol/foundry.py
JuanCoRo: I'm not sure that would work since `contract` would be discarded by that function but is used in the following line with `foundry_contract = foundry.contracts[contract]`

src/tests/integration/test_foundry_prove.py
palinatolmach: Sure, I agree! The line to be removed is:
https://github.com/runtimeverification/kontrol/blob/134b0a407385b769727bc55f291c3164782ee119/src/tests/integration/test_foundry_prove.py#L763

anvacaru:
Looks like a  few modules are added for contracts
```
S2KsrcZModcseZModAdd-VERIFICATION
S2KsrcZModcseZModMultiply-VERIFICATION
S2KsrcZModcseZModSub-VERIFICATION
S2KtestZModCSETest-VERIFICATION
S2KlibZModforgeZSubstdZModsrcZModmocksZModMockERC20-VERIFICATION
S2KlibZModforgeZSubstdZModsrcZModmocksZModIERC721TokenReceiver-VERIFICATION
S2KlibZModforgeZSubstdZModsrcZModmocksZModMockERC721-VERIFICATION
S2KsrcZModcseZModUIntBinaryOp-VERIFICATION
S2KsrcZModcseZModUIntUnaryOp-VERIFICATION
S2KtestZModnestedZModAssertNestedTest-VERIFICATION
S2KsrcZModcseZModAddConst-VERIFICATION
S2KsrcZModcseZModIdentity-VERIFICATION
S2KsrcZModcseZModIterate-VERIFICATION
S2KsrcZModcseZModWETH9-VERIFICATION
```
formatting them:
```
./src/cse/Add
./src/cse/Multiply
./src/cse/Sub
./test/CSETest
./lib/forgeZSubstd/src/mocks/MockERC20
./lib/forgeZSubstd/src/mocks/IERC721TokenReceiver
./lib/forgeZSubstd/src/mocks/MockERC721
./src/cse/UIntBinaryOp
./src/cse/UIntUnaryOp
./test/nested/AssertNestedTest
./src/cse/AddConst
./src/cse/Identity
./src/cse/Iterate
./src/cse/WETH9
```
which does not reflect the Solidity file you added.

src/tests/integration/test_foundry_prove.py
JuanCoRo: Removed! :slightly_smiling_face:

PR #430, opened 2024-03-11
cheatcodes.md: use buf instead of bufStrict in freshUInt

anvacaru:
closes: https://github.com/runtimeverification/evm-semantics/issues/2038


palinatolmach (approved):
It looks good to me, but is it a better option than adding `[preserves-definedness]`, as [suggested](https://gist.github.com/jberthold/bb62e77a6b634df8ee0a179936d01a4e#analysis) as an alternative and @PetarMax did for a few other cheatcodes before?

PetarMax:
It does save us a simplification step, since that `#bufStrict` will reduce to `#buf`. I wonder if we could drop `[preserves-definedness]` now...

PR #432, opened 2024-03-11
Add a warning on non-test analysis w/o `foundry_success` checking

### LINKED ISSUE #386: Add a warning on non-test analysis w/o `foundry_success` checking
Related: https://github.com/runtimeverification/kontrol/issues/312

In Kontrol, we only check `foundry_success` predicate (and, therefore, report the test as failing) if the name of the test starts with `test`, `check`, or `prove` (https://github.com/runtimeverification/kontrol/pull/249). Once https://github.com/runtimeverification/kontrol/issues/312 is fixed, we'll be able to also check `hevm_success` predicate.  

However, right now, if the test doesn't start from one of these three prefixes, it is being symbolically explored, but does not get reported as failing even if the test reverts. As reported by @dwightguth, it is confusing and looks like a bug, so we should probably emit a warning that informs the user that the test is _not_ being checked for failures.

lisandrasilva:
If the test doesn't start with `test`, `check`, or `prove`, it is being symbolically explored but is not reported as failing even if the test reverts. This might confuse the user, so we emit a warning informing the user that the test is not being checked for failures.

Closes #386.


src/kontrol/prove.py
```diff
@@ -480,6 +480,10 @@ def _method_to_cfg(
         init_node = cfg.create_node(init_cterm)
         new_node_ids = [init_node.id]
         init_node_id = init_node.id
+        if not method.is_testfail and not method.is_test:
+            _LOGGER.warning(
```
palinatolmach, line 484: Let's make it an INFO-level log (as per this [comment](https://github.com/runtimeverification/kontrol/issues/386#issuecomment-1957198515)):
```suggestion
            _LOGGER.info(
```

src/kontrol/prove.py
```diff
@@ -480,6 +480,10 @@ def _method_to_cfg(
         init_node = cfg.create_node(init_cterm)
         new_node_ids = [init_node.id]
         init_node_id = init_node.id
+        if not method.is_testfail and not method.is_test:
```
palinatolmach, line 483: It should be sufficient to just check if the method starts with `test`, since `test` is a prefix of `testFail` and a function starting from  `testFail` returns `true` for both `is_testfail` and `is_test`:
```suggestion
        if not method.is_test:
```

src/kontrol/prove.py
```diff
@@ -480,6 +480,10 @@ def _method_to_cfg(
...
+            _LOGGER.warning(
+                f'{contract.name_with_path}.{method.signature} is not prefixed with test or testFail, therefore it is not being checked for failures.'
```
palinatolmach, line 485: I have two suggestions here: 
1. This happens if a function is not prefixed with `test`, `check`, or `prove`
2. I'm now not sure if it "not being checked for failures" is clear to the user (I understand it's a formulation from an issue, but I'd try to refine it a bit). Maybe we can say
- "`foundry` or `hevm` success predicate is not checked" (that's technically correct, but may not be clear to external users)
- "reverts and assertion violations are not reported"
- "it is not reported as failing in presence of reverts or assertion violations"

Or something along these lines.

src/kontrol/prove.py
lisandrasilva: But then it will only be printed with `--verbose`.

src/kontrol/prove.py
lisandrasilva: Makes sense! Addressed, thank you!

src/kontrol/prove.py
```diff
@@ -480,6 +480,10 @@ def _method_to_cfg(
         init_node = cfg.create_node(init_cterm)
         new_node_ids = [init_node.id]
         init_node_id = init_node.id
+        if not method.is_test:
+            _LOGGER.info(
+                f'{contract.name_with_path}.{method.signature} is not prefixed with test or testFail, therefore it is not reported as failing in the presence of reverts or assertion violations.'
```
palinatolmach, line 485: Thanks for updating! I'd also suggest adding other test prefixes, i.e., 
```suggestion
                f'{contract.name_with_path}.{method.signature} is not prefixed with 'test', 'prove', or 'check', therefore it is not reported as failing in the presence of reverts or assertion violations.'
```

src/kontrol/prove.py
palinatolmach: Oh right! I'm ok with it being a warning (or even a `print` statement) to make sure that people see it — there already were several occasions on which users were confused that such non-test functions weren't failing as expected. 

@ehildenb but if you find it too strong, we can leave it like that.

palinatolmach:
@lisandrasilva @ehildenb I just ran it locally with `--verbose`, and the info message gets buried in other execution logs, so I'm worried that our external users won't see it: 
```
INFO 2024-03-13 15:40:52,493 pyk.kore.rpc - Starting KoreServer: kore-rpc-booster ../timeout-test/out/kompiled/definition.kore --module FOUNDRY-MAIN --server-port 0 --smt-timeout 300 --smt-retry-limit 10 --llvm-backend-library ../timeout-test/out/kompiled/llvm-library/interpreter.dylib
[Info#proxy] Loading definition from ../timeout-test/out/kompiled/definition.kore, main module "FOUNDRY-MAIN"
[Info#kore] : Info (LogMessage):
    Reading the input file TimeSpec {sec = 0, nsec = 74923000} []
[Info#kore] : Info (LogMessage):
    Parsing the file TimeSpec {sec = 0, nsec = 0} []
[Info#kore] : Info (LogMessage):
    Verifying the definition TimeSpec {sec = 0, nsec = 0} []
[Info#kore] : Info (LogMessage):
    Executing TimeSpec {sec = 1, nsec = 88710000} []
[Info#kore] : Info (LogMessage):
    Reading the input file TimeSpec {sec = 0, nsec = 77683000} []
[Info#kore] : Info (LogMessage):
    Parsing the file TimeSpec {sec = 0, nsec = 0} []
[Info#kore] : Info (LogMessage):
    Verifying the definition TimeSpec {sec = 0, nsec = 1000} []
[Info#proxy] Starting RPC server
INFO 2024-03-13 15:41:02,840 pyk.kore.rpc - KoreServer started: :::57343, pid=11646
INFO 2024-03-13 15:41:02,840 pyk.kore.rpc - Connecting to host: localhost:57343
INFO 2024-03-13 15:41:02,842 pyk.kore.rpc - Connected to host: localhost:57343
INFO 2024-03-13 15:41:02,842 pyk.kast.outer - Loading JSON definition: ../timeout-test/out/kompiled/compiled.json
INFO 2024-03-13 15:41:03,159 pyk.kast.outer - Converting JSON definition to Kast: ../timeout-test/out/kompiled/compiled.json
INFO 2024-03-13 15:41:03,730 pyk.kore.kompiled - Reading JSON definition: ../timeout-test/out/kompiled/pyk-definition.kore.json
INFO 2024-03-13 15:41:03,983 kontrol.prove - Initializing KCFG for test: test%ConstructorTest.run_constructor():0
INFO 2024-03-13 15:41:04,054 kontrol.prove - test%ConstructorTest.run_constructor() is not prefixed with test or testFail, therefore it is not reported as failing in the presence of reverts or assertion violations.
INFO 2024-03-13 15:41:04,056 kontrol.prove - Expanding macros in node 1 for test: test%ConstructorTest.run_constructor()
INFO 2024-03-13 15:41:04,149 kontrol.prove - Computing definedness constraint for node 1 for test: test%ConstructorTest.run_constructor()
INFO 2024-03-13 15:41:04,158 pyk.kore.rpc - Sending request to localhost:57343: 1 - simplify
[Info] Process request 1 simplify
[Info#booster] Simplifying predicates
```
Can we maybe bring it closer to the start of the execution or, instead, print it next to the proof execution result (`PROOF PASSED ...`)?

lisandrasilva:
> Can we maybe bring it closer to the start of the execution or, instead, print it next to the proof execution result (`PROOF PASSED ...`)?
@palinatolmach Yes, I can take care of that. Which option do you prefer? At the beginning of execution or next to the execution result?

lisandrasilva:
Right now, the warning is as follows:
```
...
[Info] Process request 15 implies
[Info#proxy] ImpliesM (using kore)
[Info] Process request 16 implies
[Info#proxy] ImpliesM (using kore)
WARNING 2024-03-13 16:26:37,760 kontrol.prove - assertEq_true(uint256) is not prefixed with test or testFail, therefore it is not reported as failing in the presence of reverts or assertion violations.
[Info#proxy] Server shutting down
PROOF PASSED: test%HevmTests.assertEq_true(uint256):0
time: 55s
```
It's still not ideal since if there is more than one proof running it gets again buried in the execution log. I'll print it together with the proof status so it's more visible to the user.

lisandrasilva:
Right now, it is printed as follows:
```
...
[Info] Process request 15 implies
[Info#proxy] ImpliesM (using kore)
[Info] Process request 16 implies
[Info#proxy] ImpliesM (using kore)
[Info#proxy] Server shutting down
PROOF PASSED: test%HevmTests.assertEq_true(uint256):0
time: 55s
Info: test%HevmTests.assertEq_true(uint256):0 is not prefixed with test or testFail, therefore it is not reported as failing in the presence of reverts or assertion violations.
```
Please let me know what you think.

palinatolmach:
Thank you @lisandrasilva! It looks great. A few minor nits from my side:
- If Everett's ok with it, maybe we can leave it a warning, as you did originally, I think this format reads more naturally:
```
WARNING 2024-03-13 16:26:37,760 kontrol.prove - assertEq_true(uint256) is not prefixed with test or testFail, therefore it is not reported as failing in the presence of reverts or assertion violations.
```
Or maybe we can just make `INFO` all uppercase? If nor, I think it'd also be fine if we just unify `time` and `Info` to both start with a capital (or a small) letter.
- Another stylistic comment: I feel like it'd be more natural if we print this warning either before `PROOF PASSED` or between  `PROOF PASSED` and `time: ...`. Maybe we can print it only for passing proofs?
- I'd change `is not prefixed with test or testFail` to `if not prefixed with 'test', 'prove', or 'check'` to make sure we don't confuse users about what is considered a "test"

I don't have a strong opinion about any of these, please let me know what you'd prefer!

palinatolmach (approved):
Approved with some stylistic comments (please see https://github.com/runtimeverification/kontrol/pull/432#issuecomment-1996594888). It also requires updating expected output to include the new info message for failing tests.

lisandrasilva:
> If Everett's ok with it, maybe we can leave it a warning, as you did originally, I think this format reads more naturally

I agree!

> * Another stylistic comment: I feel like it'd be more natural if we print this warning either before `PROOF PASSED` or between  `PROOF PASSED` and `time: ...`. Maybe we can print it only for passing proofs?

I would prefer before `PROOF PASSED`.

lisandrasilva:
At the moment, the output is as follows: 
```
...
[Info#proxy] Server shutting down
WARNING 2024-03-14 09:48:29,943 kontrol.__main__ - assertEq_true(uint256) is not prefixed with 'test', 'prove', or 'check', therefore, it is not reported as failing in the presence of reverts or assertion violations.
PROOF PASSED: test%HevmTests.assertEq_true(uint256):0
time: 54s
```

PR #436, opened 2024-03-12
CI task for CSE 

PetarMax:
This PR adds a separate CI task for running CSE tests because of the complexity of said tests. It also moves the harness for running CSE tests into a separate file.


ehildenb (approved):

.github/workflows/test-pr.yml
```diff
@@ -116,14 +116,45 @@ jobs:
...
 
+  cse-tests:
+    needs: code-quality-checks
+    name: 'CSE Tests'
+    runs-on: [self-hosted, linux, normal, fast]
+    strategy:
+      fail-fast: false
+      matrix:
+        backend: ['booster']
+    timeout-minutes: 120
```
F-WRunTime, line 135: Will this be enough time, given the timeout we saw previously?

PetarMax: I think so, yes. If it's not, we will increase.

Issue #411, opened 2024-03-05
Adjust merge-nodes functionality for loop-invariant discovery


ehildenb:
Currently, the `merge-nodes` functionality will not work for two nodes that are not leaf nodes, because of the restriction on adding only a single successor per node in the KCFG. If a user wants to automatically help out with inferring a loop invariant, they may instead want to find two nodes that are the same loop-head in the KCFG, and merge them, to come up with a description of the state invariant to use for building the loop invariant.

We should test out the `merge-nodes` functionality on a piece of looping code where we unroll the loop a certain number of times, then call `merge-nodes` on two or three of the loop heads. This test will fail initially, because of the above restriction, so instead we'll have to prune the KCFG in a way to insert the merged node correctly.

This can be used for helping to infer the wordstack and localmem structure of the loop invariant LHS claim, to help speed up the process of updating loop invariants for engagements.

Ping @PetarMax and @palinatolmach

ehildenb:
Just to be clear, this issue is _just_ to fix the `merge-nodes` functionality in the case of merging two nodes that have successor states, so that we can use it to merge loop-heads after doing a fixed unrolling. The merged state _may_ be helpful in identifying a loop invariant, we don't know because we haven't tried, but I suspect it will be.

We are not trying to do loop invariant inference here, we are simply trying to make the `merge-nodes` functionality a bit more generic and see if it could _eventually_ help with that.

PR #444, opened 2024-03-13
Update `merge-nodes` to function for non-leaf nodes

qian-hu:
Closes: #411.

The `merge-nodes` functionality currently lacks support for merging two nodes that are not leaf nodes due to the restriction on adding only a single successor per node in the KCFG. This PR aims to modify the `merge-nodes` functionality to handle merging two nodes with successor states properly.


src/tests/integration/test_foundry_prove.py
```diff
@@ -334,6 +334,65 @@ def test_foundry_merge_nodes(
...
+
+    foundry_merge_nodes(foundry, test=test, node_ids=[15, 16], include_disjunct=True)
+
+    foundry_prove(
+        foundry,
+        tests=[(test, None)],
+        prove_options=ProveOptions(
+            max_iterations=10,
+            bug_report=bug_report,
+        ),
```
ehildenb, line 372:
```suggestion
        prove_options=ProveOptions(
            max_iterations=2,
            bug_report=bug_report,
        ),
```

Just to reduce how long this test runs for, and how much output is generated, can you restrict this to just 2 iterations after the node merge, and then update the expected output?

ehildenb (approved):
I've approved, but please address the review comment I have.

PR #445, opened 2024-03-13
Update master-push.yml

F-WRunTime:
Testing cachix failure fix related to nix-shell-action@v3.0.3


F-WRunTime:
Here is the set of consecutive runs. One of public then the next on internal intel runner. 

let me know what you think.

F-WRunTime:
Can't make heads or tails of why [this is failing ](https://github.com/runtimeverification/kontrol/actions/runs/8275194236/job/22713364756?pr=445)
 @goodlyrottenapple

.github/workflows/master-push.yml
```diff
@@ -3,40 +3,25 @@ on:
   push:
     branches:
       - 'master'
+      - 'cachix-failure'
```
F-WRunTime, line 6: This line will need to be removed before Merge.

F-WRunTime (commented):
I think this is a viable fix to fix the cachix publish issue occurring on the public Mac intel runner.

Baltoli:
Just to check I understand @F-WRunTime, the fix here is to run these workflows on a self-hosted intel mac runner with Nix and Cachix pre-installed - is that right?

F-WRunTime:
> Just to check I understand @F-WRunTime, the fix here is to run these workflows on a self-hosted intel mac runner with Nix and Cachix pre-installed - is that right?

@Baltoli That is correct. I think with the Public runners, we may likely be hitting a resource wall that is killing our builds, either cutting it off at the network bandwidth level or we're hitting a RAM related issue stopping uploads from proceeding. 
It is hard to tell with a limited window into the public runner, and how they shut down OOM limits. 
But with some testing it is visible to see we can successfully push these builds from our own intel runner

goodlyrottenapple (approved):

PR #467, opened 2024-03-19
Control reinitializing of the proof's `setUp` function

### LINKED ISSUE #282: Tests with a dedicated `setUp` function run slower than tests with inlined setup code.
I've observed a consistent performance discrepancy in our contract tests that utilize a dedicated `setUp` function compared to those where the setup code is inlined directly within the test function. This slower execution persists even when the `setUp` function has been previously executed, and its results are available from the file system cache. Of course, this situation flips when the `setUp` function is performing a lot of work, but for the average use case it seems to slow things down significantly. For the cases where the deployment logic becomes too complex, we now have the fast summarization feature. So, I think we should optimize the `setUp` logic for simple cases.

I expect this difference to become even more important once we enable compositional symbolic execution.

I'm wondering if there is some low-hanging fruit here to increase performance for contracts with a dedicated `setUp` function.

ovatman:
Resolves #282 
This PR adds the ability to skip the reinitalization of the `setUp()` function even though the `--reinit` flag was passed.
Users can pass `--skip-setup-reinit` flag to selectively skip the `setUp()`  proof being initialized. 

This ability can be handled in a couple of different ways. This PR implements it by directly using the test's version when creating setup proofs and create jumps in the versions of the setup proofs. When a proof with a missing setup-id proof is called the latest setup proof that is smaller than current proof's id is used.

An alternative approach may be to relate setup functions with test functions by adding test function name to proof name such as instead of `ContractName.setUp()` we may use `ContractName.testName.setUp()` which will increase the footprint of proofs.

A more relaxed implementation can break the id relation between test proofs and respected setup functions and always using the latest version of the `setUp()` proof.


src/kontrol/foundry.py
```diff
@@ -542,6 +576,14 @@ def latest_proof_version(
         versions = {int(pid.split(':')[1]) for pid in proof_ids}
         return max(versions, default=None)
 
+    def latest_proof_version_before(self, test: str, base_test_ver: int) -> int | None:
+        """
+        find the highest used proof ID, less than. Returns None if no version of this proof exists.
```
palinatolmach, line 581: Some nitpicking on [docstrings](https://peps.python.org/pep-0257/#multi-line-docstrings):
```suggestion
        Find the highest used proof ID, less than base_test_ver. 
        
        Returns None if no version of this proof exists.
```

palinatolmach:
This implementation seems reasonable to me. I wouldn't tie test names to their respective setups, because it'd be more efficient to allow for tests sharing the same `setUp()`. But I'm on the fence between this implementation and just using the latest `setUp()` proof — since `setUp` is shared between functions, maybe the user would expect the second command to use the latest version of setUp, since he knows it was reinitialized in the first command:
```
kontrol prove --match-test testA --reinit
kontrol prove --match-test testB --skip-setup-reinit --reinit
```
I'd also like to ask for opinion from @nwatson22, who implemented the logic for proof versioning, and @anvacaru.

src/kontrol/__main__.py
```diff
@@ -781,6 +783,13 @@ def parse(s: str) -> list[T]:
         action='store_true',
         help='Reinitialize CFGs even if they already exist.',
     )
+    prove_args.add_argument(
+        '--skip-setup-reinit',
+        dest='skip_setup_reinit',
+        default=False,
+        action='store_true',
+        help='Do not reinitialize test setup; reuse most recent setup proof instead',
```
palinatolmach, line 791: This reads like "the most recent version of `setUp` proof that was ever generated", but, if I understood correctly, it's test-specific?

palinatolmach: It might also be worth mentioning that it should be used together with `--reinit`.

nwatson22:
> Thanks for addressing this issue @ovatman! This implementation seems very reasonable to me. I wouldn't tie test names to their respective setups, because it'd be more efficient to allow for tests sharing the same `setUp()`. But I'm on the fence between this implementation and just using the latest `setUp()` proof — since `setUp` is shared between functions, maybe the user would expect the second command to use the latest version of setUp, since he knows it was reinitialized in the first command:
> 
> ```
> kontrol prove --match-test testA --reinit
> kontrol prove --match-test testB --skip-setup-reinit --reinit
> ```
> 
> I'd also like to ask for opinion from @nwatson22, who implemented the logic for proof versioning, and @anvacaru.

@ovatman @palinatolmach I agree using the latest proof version of the setup method before the version of the corresponding main method is maybe not ideal, since I don't really think of the version as being consistent across proofs. I would probably agree that using the latest version of setUp is the way to go, although I could still see situations where that would cause confusion. Do you you think would be better to, instead of having the `--skip-setup-reinit` option, to have an option that specifies the exact setup version to use?

palinatolmach:
@ovatman @nwatson22 Yes, I agree, having something like `--setup-version` sounds good to me.

src/kontrol/__main__.py
```diff
@@ -790,6 +792,13 @@ def parse(s: str) -> list[T]:
         action='store_true',
         help='Reinitialize CFGs even if they already exist.',
     )
+    prove_args.add_argument(
+        '--setup-version',
+        dest='setup_version',
+        default=None,
+        type=int,
+        help='Instead of reinitializing the test setup with together the test proof, select the setup version to be reused during the proof.',
```
palinatolmach, line 800:
```suggestion
        help='Instead of reinitializing the test setup together with the test proof, select the setup version to be reused during the proof.',
```

palinatolmach: Or maybe it's also worth highlighting that it should be used together with `--reinit`, like we say
```
  --break-on-basic-blocks
                        Store a node for every EVM basic block (implies --break-on-calls).
```
E.g., something like
```
"When reinitializing the proof with '--reinit', do not reinitialize 'setUp' and use the specified version instead"
```
Or you can just add `(implies --reinit)`, or something like that, but it looks good as it is too.

palinatolmach (approved):
Left some minor comments on help message, otherwise looks good!

PR #474, opened 2024-03-21
Adding fresh symbolic words

### LINKED ISSUE #292: Implement more `fresh*` cheatcodes
In addition to [`freshBytes`](https://github.com/runtimeverification/kontrol/pull/254), we should also support cheatcodes that allow creating fresh symbolic variables of other types, in particular `freshAddress()`.

lisandrasilva:
This PR adds the `freshAddress` cheatcode as well as installs `kontrol-cheatcodes` library in integration tests which allows access to more symbolic integers of various sizes. 

Closes #292.


src/kontrol/kdist/cheatcodes.md
```diff
@@ -440,7 +440,7 @@ This rule returns a symbolic boolean value being either 0 (false) or 1 (true).
        [preserves-definedness]
 ```
 
-#### `freshBytes` - Returns a single symbolic boolean.
+#### `freshBytes` - Returns a fully symbolic byte array value of the given length.
```
palinatolmach, line 443: Nice catch!

anvacaru (approved):

palinatolmach (approved):
It looks great, thanks @lisandrasilva! Can you please just add some more tests for new functions that are part of `SymbolicWords`, such as, e.g., `freshUInt32()`?

PetarMax:
I was wondering - should `freshAddress` be different from the existing addresses?

palinatolmach:
> I was wondering - should `freshAddress` be different from the existing addresses in `<accounts>`, and possibly some dedicated other addresses?

Oh, that's a great idea! I think it would be good for it to be different from Foundry cheatcode and test addresses, at least, similar to the constraints imposed on `CALLER_ID` and `ORIGIN_ID`.

PR #480, opened 2024-03-25
Corrections to how `setUp` and tests are connected

PetarMax:
This PR corrects the connection between the KCFG of the `setUp` function and the KCFG of the test function when running tests. In particular:
- it ensures that the `setUp` proof is passing, that is, that is has no pending/failing nodes;
- it minimizes the resulting `setUp` KCFG; and
- it propagates the constraints from the `setUp` final node to the test initial node.


src/tests/integration/test-data/foundry/test/SetUpTest.t.sol
```diff
@@ -38,4 +42,8 @@ contract SetUpTest is Test {
         // x is used and not thrown away by the optimizer
         assertEq(x, x);
     }
+
+    function testSetupData() public {
```
PetarMax, line 46: This test fails on `master` and passes on this branch. It effectively states that the constraint on a contract storage variable established in `setUp` is propagated to the main test.

src/tests/integration/test_foundry_prove.py
```diff
@@ -410,7 +410,7 @@ def test_foundry_remove_node(
     foundry_remove_node(
         foundry,
         test=test,
-        node=4,
+        node=8,
```
PetarMax, line 413: This needed to be changed because the minimization of the KCFG of the associated `setUp` function resulted in a change of node numbering.

src/kontrol/prove.py
```diff
@@ -468,9 +469,22 @@ def _method_to_cfg(
...
+                f'Initial state proof {setup_proof.id} for {contract.name_with_path}.{method.signature} still has failing branches.'
+            )
 
-        cfg = KCFG.from_dict(setup_proof.kcfg.to_dict())  # Copy KCFG
+        assert setup_proof.status == ProofStatus.PASSED
+
+        init_node_id = setup_proof.init
+        # Copy KCFG and minimize it
+        cfg = KCFG.from_dict(setup_proof.kcfg.to_dict())
+        cfg.minimize()
```
PetarMax, line 487: By minimizing the `setUp` KCFG, we reduce it to a single transition (unless `setUp` branches).

src/kontrol/prove.py
```diff
@@ -468,9 +469,22 @@ def _method_to_cfg(
                 f'Initial state proof {setup_proof.id} for {contract.name_with_path}.{method.signature} still has pending branches.'
             )
 
-        init_node_id = setup_proof.init
+        if setup_proof.bounded:
+            raise RuntimeError(
+                f'Initial state proof {setup_proof.id} for {contract.name_with_path}.{method.signature} still has bounded branches.'
```
palinatolmach, line 474: Does `bounded` indicates that at least of the nodes in the proof is bounded? Do we not allow those in `setUp`?

src/tests/integration/test-data/show/foundry.k.expected
```diff
@@ -117,7 +117,7 @@ module FOUNDRY-MAIN
     imports public S2KtestZModStoreTest-VERIFICATION
     imports public S2KtestZModSymbolicStorageTest-VERIFICATION
     imports public S2KtestZModSymbolicStore-VERIFICATION
-    imports public S2KlibZModforgeZSubstdZModsrcZModTest-VERIFICATION
+    imports public S2KlibZModforgeZSubstdZModsrcZModTest-VERIFICATION
```
palinatolmach, line 120: This extra space might cause `test_foundry_kompile` to fail.

palinatolmach (approved):
Approved with a single question. Let's bring up the `setUp` minimization in the standup and/or Slack channel, so everyone's aware of it.

src/kontrol/prove.py
PetarMax: I don't think we should, this means that `setUp` is looping and we haven't gotten out of the loop. I believe `setUp` must fully terminate successfully for us to proceed with the test?

palinatolmach: That makes sense! I'm just not sure what should happen if we're using BMC — if we've bounding all loops to, say, 3 iterations, that should probably include `setUp` loops and allow continuing the execution after that? Or would we not have `bounded` nodes in this case?

PetarMax: Ok, let's discuss this, because I'm not sure what kind of loop we could run into in `setUp`.

Issue #400, opened 2024-02-28
Use `--ast` in `forge build` in Kontrol


palinatolmach:
As discussed in https://github.com/runtimeverification/kontrol/pull/399, starting from https://github.com/foundry-rs/foundry/pull/7197, Foundry doesn't include AST in the output of `forge build` by default. If the AST is not present in `solc`-produced files, https://github.com/runtimeverification/kontrol/pull/399 added a descriptive error message suggesting that the user should add `ast = true` to `foundry.toml`.

However, for the AST to be generated automatically, we could also include `--ast` into the set of flags we are calling `forge build` with: https://github.com/runtimeverification/kontrol/blob/fc9a495e6628cbc046935f902e938184a080477e/src/kontrol/foundry.py#L252
The `--ast` flag was only added in https://github.com/foundry-rs/foundry/pull/7197, so it's not backwards compatible, and would require checking a version of Foundry being used first.

palinatolmach:
Alternatively, we can add a `--build-info` flag to `forge build`, which is backward compatible, but generates additional information (such as the build info file itself) — it doesn't increase the execution time for `forge build` on our integration test suite, but occupies more disk space: 
![image](https://github.com/runtimeverification/kontrol/assets/11912158/c47c4806-6d7f-4520-baca-e5191dc081c2)

PR #487, opened 2024-03-26
Add `build-info` to `forge build`

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/400.

After consideration with @anvacaru, it was decided to proceed with adding `--build-info` to `forge build`, which is backward compatible, but generates additional information, therefore occupying more disk space. 

An alternative is to add `--ast` (not backwards compatible) when a newer Foundry version is detected in the system, however, it appears that Foundry does not yet have a reliable and sufficiently fine-grained release and versioning system (I'll submit an issue to their repository).


anvacaru (approved):
LGTM 👍

Issue #448, opened 2024-03-13
Investigate inference of loop invariants using `merge-nodes` functionality


ehildenb:
Here https://github.com/runtimeverification/kontrol/pull/444, we update the `merge-nodes` functionality to allow specifying loop heads in looping code. Here, we have an example that uses a loop invariant to allow verifying a small piece of code with a simple loop: https://github.com/runtimeverification/kontrol/pull/369.

The `merge-nodes` functionality should help us in identifying the loop invariant needed in order to verify such simple loops. We probably can't infer the entire loop invariant, but at least we can make an initial guess ath the correct LHS of the loop to use.

Here:

-

ehildenb:
@qian-hu I don't quite see the nodes mentioned in the merges in the KCFG output there. Node 19 appears to be a target node, and node 13 and 14 are on either side of the same branch. Nodes 26 and 31 don't exist. Can we see the original KCFG output as well?

ehildenb:
Also, what is the process needed to go from the original merged node to the invariant there? What manual steps need to be taken?

ehildenb:
Another thought is this. The invariant has the form:

```
rule LOOP => LOOP_ENDED requires LOOP_INVARIANT
```

The `LOOP` node is the node we are currently trying to infer by merging a bunch of other nodes which show up in a repeating pattern for the loop. The `LOOP_ENDED` and `LOOP_INVARIANT` ones we probably can't infer as automatically.

But every loop should have structure `a -branch-> {b, c}`, where `b` is "stepping into the loop", and `c` is "stepping out of the loop" (because of the loop head condition). In this example, we've collected several `b_1, b_2, b_3, ...` and merged them to try and discover LHS of the invariant (the `LOOP` part). We could also consider collecting several `c_1, c_2, c_3, ...` and merge them to try and discover the RHS of the loop invariant (the `LOOP_ENDED` part).

palinatolmach:
Related: https://github.com/runtimeverification/k/issues/4425

PR #495, opened 2024-03-29
Update `merge-nodes` for merging nodes within the same loop

qian-hu:
This PR enhances the `merge-nodes` functionality to support merging nodes within the same loop but with different K cells. It enables improved inference of loop information and contributes to resolving #448.


src/kontrol/foundry.py
```diff
@@ -947,8 +956,12 @@ def check_cells_equal(cell: str, nodes: Iterable[KCFG.Node]) -> bool:
     nodes = [apr_proof.kcfg.node(int(node_id)) for node_id in node_ids]
     check_cells = ['K_CELL', 'PROGRAM_CELL', 'PC_CELL', 'CALLDEPTH_CELL']
     check_cells_ne = [check_cell for check_cell in check_cells if not check_cells_equal(check_cell, nodes)]
+
     if check_cells_ne:
-        raise ValueError(f'Nodes {node_ids} cannot be merged because they differ in: {check_cells_ne}')
+        if check_same_loop(nodes):
+            pass
+        else:
+            raise ValueError(f'Nodes {node_ids} cannot be merged because they differ in: {check_cells_ne}')
```
ehildenb, line 964:
```suggestion
        if not check_same_loop(nodes):
            raise ValueError(f'Nodes {node_ids} cannot be merged because they differ in: {check_cells_ne}')
```

ehildenb: We already check above that `len(nodes) >= 2`, so this more compact inline check should work.

src/kontrol/foundry.py
```diff
@@ -938,6 +938,15 @@ def check_cells_equal(cell: str, nodes: Iterable[KCFG.Node]) -> bool:
...
         return True
 
+    def check_same_loop(nodes: Iterable[KCFG.Node]) -> bool:
+        nodes = list(nodes)
+        if len(nodes) < 2:
+            return True
+        for node in nodes[1:]:
+            if not KEVMSemantics().same_loop(nodes[0].cterm, node.cterm):
+                return False
+        return True
```
ehildenb, line 948: With the more compact inline check below, we don't need this function anymore.

ehildenb:
Why do we make this restriction that the user can only merge nodes that are the same loop head? I don't understnad the benefit, it seems to limit how teh usre can use `merge-nodes`, and for example it can be useful when merging togethre paths right after a branch into a single node. So I think we're losing functionality here.

ehildenb:
@qian-hu thanks for the explanation! It makes sense now. It looks like it needs to be updated to latest master to fix merge conflicts and check that tests are passing.

ehildenb (approved):

PR #501, opened 2024-04-02
Add options classes

nwatson22:
Extend `Options` from pyk to centralize the location of default values for command line options. Adds option groups in `cli.py` and top-level command options groups in the foundry-related files.


src/kontrol/__main__.py
```diff
@@ -69,6 +80,51 @@ def _ignore_arg(args: dict[str, Any], arg: str, cli_option: str) -> None:
         args.pop(arg)
 
 
+def generate_options(args: dict[str, Any]) -> LoggingOptions:
+    command = args['command']
+    match command:
```
tothtamas28, line 85: Consider a `dict` lookup instead:

```python3
options = {
    'load-stat-diff': LoadStateDiffOptions(args),
    ...
}
try:
    return options[command]
except Keyerror as err:
    raise ValueError(f'Unrecognized command: {command}') from err
```

src/kontrol/__main__.py
```diff
@@ -89,12 +145,17 @@ def main() -> None:
 
     _check_k_version()
 
+    stripped_args = {
+        key: val for (key, val) in vars(args).items() if val is not None and not (isinstance(val, Iterable) and not val)
```
tothtamas28, line 149: What does the condition express?

src/kontrol/__main__.py
```diff
@@ -748,21 +462,21 @@ def parse(s: str) -> list[T]:
     build.add_argument(
         '--regen',
         dest='regen',
-        default=False,
+        default=None,
```
tothtamas28, line 465: These can just be removed.

src/kontrol/cli.py
```diff
@@ -15,6 +16,83 @@
     T = TypeVar('T')
 
 
+class FoundryOptions(Options):
+    foundry_root: Path
```
tothtamas28, line 20: Should these be `@dataclass`-es?

src/kontrol/__main__.py
nwatson22: Replaced this with a dict lookup.

src/kontrol/__main__.py
nwatson22: I thought so in a previous PR, but I guess omitting `default` does not always make the default `None`. At least in the case of `bool` it makes it false, although that may be caused by the `action='store_true'`, and this will overwrite the default specified in `Options.default`.

src/kontrol/__main__.py
nwatson22: argparse always sets some default for all options. This is attempting to strip those out so the default as determined by `Options.default` can be applied. I'm just trying to stick to a convention here where the default value specified for argparse for non-list parameters is `None` and for lists is `[]`, but I may be able to refactor the list to also be `None` which would simplify this condition.

src/kontrol/cli.py
nwatson22: Possibly, but I'm not sure. The base class in pyk provides a constructor that looks like this:
```
class Options:
    def __init__(self, args: dict[str, Any]) -> None:
        # Get defaults from this and all superclasses that define them, preferring the most specific class
        defaults: dict[str, Any] = {}
        for cl in reversed(type(self).mro()):
            if hasattr(cl, 'default'):
                defaults = defaults | cl.default()

        # Overwrite defaults with args from command line
        _args = defaults | args

        for attr, val in _args.items():
            self.__setattr__(attr, val)
```
which takes in the dict from argparse and combines that with the `default`s. Possibly we could still do this in a custom dataclass constructor like exists here: https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/options.py, but we'd have to specify all the options again in the signature to that, and the optional ones would have to be ` = None`

tothtamas28 (approved):

PR #503, opened 2024-04-02
Add option to pretty print in hex encoding

anvacaru:
blocked on : https://github.com/runtimeverification/kontrol/pull/502

Makes use of the newly overloaded `KEVM.pretty_print` function that can print tokens of sort `INT` and `BYTES` as hex, implemented in https://github.com/runtimeverification/evm-semantics/pull/2382.

In this PR:
    - modified `_load_foundry` function to take the `use_hex_encoding` arg to pass it to the KEVM constructor.
    - modified `exec_view_kcfg` to use hex encoding by default.
    - added a new `--use-hex-encoding` cli flag for `kontrol show`


ehildenb:
I would make using the hex encoding default, and then have `--no-use-hex-encoding` for turning it off if we need to print out configurations that should be parseable.

palinatolmach (approved):
lgtm, thank you @anvacaru! I agree with Everett's comment above.

anvacaru:
After discussing in the daily stand up, we decided to merge this as it is for now and set the hex encoding by default after I sort out the issue with the `alias` unparsing.

PR #506, opened 2024-04-05
Proof execution time is updated when proof is reloaded

### LINKED ISSUE #489: Proof execution time is updated when proof is reloaded
Related: https://github.com/runtimeverification/kontrol/pull/391

When the proof is reloaded, the time of this run (loading the proof and generating a model) is added to the total proof execution time. For example, for a proof executed once in a bit over 5 minutes, some of the following `kontrol prove` runs report the following execution time:
```
❯ kontrol prove --match-test ERC4626Test.testTransfer
Running functions: ['test%ERC4626Test.testTransfer(address,address,uint256)']
PROOF FAILED: test%ERC4626Test.testTransfer(address,address,uint256):0
time: 8m 15s
```
then, the next one is
```
❯ kontrol prove --match-test ERC4626Test.testTransfer
Running functions: ['test%ERC4626Test.testTransfer(address,address,uint256)']
PROOF FAILED: test%ERC4626Test.testTransfer(address,address,uint256):0
time: 9m 6s
```
etc.

lisandrasilva:
When a failing proof is reloaded from disk, the time to load the proof and generate a model is added to the proof's execution time. With this PR, the execution time of a proof loaded from disk is only updated if the proof status is `PENDING`. 

Fixes #489.


src/kontrol/prove.py
```diff
@@ -244,11 +244,12 @@ def _run_cfg_group(
...
     def init_and_run_proof(test: FoundryTest) -> APRFailureInfo | Exception | None:
+        proof = None
         if Proof.proof_data_exists(test.id, foundry.proofs_dir):
-            apr_proof = foundry.get_apr_proof(test.id)
-            if apr_proof.passed:
+            proof = foundry.get_apr_proof(test.id)
+            if proof.passed:
                 return None
-        start_time = time.time()
+        start_time = time.time() if proof is None or proof.status == ProofStatus.PENDING else 0.0
```
palinatolmach, line 252: Would it make sense to make it `None` instead of `0.0`, and have 
```
  if start_time is not None:
      end_time = time.time()
```
later?

palinatolmach (approved):
left one comment, thanks @lisandrasilva!

src/kontrol/prove.py
lisandrasilva: Yes, it is more readable. Addressed :)

PR #507, opened 2024-04-05
Native Assertions in Kontrol

anvacaru:
- adds `assert.md` with minimal native assertions supported.
- removing two unused/empty sol files: `Contract.sol` and `Constants.sol`
- in https://github.com/runtimeverification/kontrol/pull/507/commits/dbda29503b5949532860f8d451b4a83598e52f88: refactoring `kontrol init` command to take a `name` of a project as an argument, and extracting the functionality in foundry.py to use it in the new test group.
- added a new integration test file named `test_kontrol.py` with its own `foundry_project` fixture and `test_kontrol_end_to_end` that tests:
      -  a new project is created using the functionality of `kontrol init`, with the latest versions of `forge-std` and `kontrol-cheatcodes`.
      - a small batch of Solidity unit tests is copied from `src/tests/integration/test-data/src/` to the new project.
      - tests are executed using `kontrol build` and `kontrol prove`, checking only that the proofs pass.
- add a new `end-to-end` stage in CI that runs the test presented above.
- small refactorings to `test_foundry_prove.py` and `test_hevm_prove.py` to move duplicated `assert_pass` and `assert_fail` functions under `utils.py`.


anvacaru:
Seems like there is a slowdown introduced by this change:

Time difference for show tests:
```
| Test                                                                        |  master time   | native-assert time | (master/native-assert) time    
|-----------------------------------------------------------------------------|--------------- |--------------------|--------------------------------
| InitCodeTest.test_init()                                                    | 125.53         | 216.95             | 0.5786125835445955 
| AssumeTest.testFail_assume_false(uint256,uint256)                           | 66.49          | 107.35             | 0.6193758733115976 
| ConstructorTest.testFail_constructor()                                      | 61.06          | 96.01              | 0.6359754192271638 
| InitCodeTest.testFail_init()                                                | 49.31          | 72.3               | 0.6820193637621024 
| AddrTest.test_addr_true()-trace_options1                                    | 37.44          | 53.73              | 0.6968174204355109 
| src/tests/integration/test_foundry_prove.py::test_foundry_merge_loop_heads  | 164.03         | 228.27             | 0.7185788758925833 
| src/tests/integration/test_foundry_prove.py::test_foundry_refute_node       | 111.15         | 146.93             | 0.7564826788266521 
| BlockParamsTest.testWarp(uint256)-trace_options2                            | 37.98          | 48.63              | 0.7809993830968537 
| src/tests/integration/test_foundry_prove.py::test_foundry_xml_report        | 93.17          | 117.16             | 0.7952372823489245 
| AccountParamsTest.testDealConcrete()-trace_options0                         | 40.18          | 50.39              | 0.7973804326255209 
| AssertTest.testFail_expect_revert()                                         | 106.95         | 133.5              | 0.801123595505618  
| AssertTest.testFail_assert_true()                                           | 55.51          | 69.27              | 0.801357008806121  
| AssertTest.checkFail_assert_false()                                         | 42.31          | 52.69              | 0.8029986714746632 
| src/tests/integration/test_foundry_prove.py::test_foundry_auto_abstraction  | 51.7           | 63.57              | 0.8132767028472551 
| ConstructorTest.test_constructor()                                          | 165.77         | 201.67             | 0.8219864134477117 
| src/tests/integration/test_foundry_prove.py::test_foundry_merge_nodes       | 46.9           | 56.92              | 0.8239634574841883 
| AssertTest.test_assert_false()                                              | 90.95          | 110.18             | 0.8254674169540751 
| SetUpDeployTest.test_extcodesize()                                          | 93.26          | 112.67             | 0.82772699032573   
| AssertTest.test_assert_true()                                               | 76.44          | 92.28              | 0.8283485045513653 
| HevmTests.proveFail_all_branches                                            | 64.36          | 76.84              | 0.8375845913586674 
| AssumeTest.testFail_assume_true(uint256,uint256)                            | 82.99          | 96.93              | 0.8561848756834828 
| src/tests/integration/test_foundry_prove.py::test_foundry_resume_proof      | 79.68          | 93.05              | 0.8563138097796884 
| AssertTest.test_revert_branch(uint256,uint256)                              | 105.38         | 121.38             | 0.8681825671445048 
| src/tests/integration/test_foundry_prove.py::test_foundry_split_node        | 552.03         | 633.91             | 0.8708333990629584 
| src/tests/integration/test_foundry_prove.py::test_foundry_prove_skips_setup | 173.05         | 198.54             | 0.8716127732446863 
| AssumeTest.test_assume_false(uint256,uint256)                               | 87.53          | 97.24              | 0.9001439736733855 
| HevmTests.prove_require_assert_false                                        | 60.58          | 66.93              | 0.9051247572090242 
| HevmTests.proveFail_require_assert                                          | 55.53          | 58.64              | 0.9469645293315143 
| HevmTests.prove_revert                                                      | 20.83          | 21.8               | 0.9555045871559632 
| BMCLoopsTest.test_countdown_concrete()                                      | 26.27          | 25.02              | 1.0499600319744204 
| AssertTest.test_failing_branch(uint256)                                     | 147.6          | 130.54             | 1.1306879117511874 
| src/tests/integration/test_foundry_prove.py::test_foundry_remove_node       | 113.93         | 89.08              | 1.2789627301302202 
| TOTAL                                                                       | 3089.07        | 3745.3199999999997 | 0.8247813270962162 
```

Time difference for CSE and minimization tests
```
| Test                                                                     |  master time              | native-assert time | (master/native-assert) time    
|--------------------------------------------------------------------------|---------------------------|--------------------|----------------------------
| Identity.applyOp(uint256)                                                | 115.93                    | 136.24             | 0.8509248385202584 
| Identity.identity(uint256)                                               | 16.37                     | 19.18              | 0.8534932221063608 
| AddConst.applyOp(uint256)                                                | 73.95                     | 86.06              | 0.8592842203114106 
| AssertTest.test_revert_branch(uint256,uint256)                           | 126.12                    | 146.18             | 0.8627719250239431 
| AssertTest.test_failing_branch(uint256)                                  | 140.68                    | 162.59             | 0.865243864936343  
| CSETest.test_add_const(uint256,uint256)                                  | 255.67                    | 293.55             | 0.8709589507749956 
| AssertTest.test_assert_false()                                           | 88.05                     | 101.07             | 0.8711783912140101 
| ArithmeticContract.add(uint256,uint256)                                  | 73.6                      | 84.43              | 0.8717280587468907 
| AssertTest.testFail_expect_revert()                                      | 133.17                    | 151.88             | 0.8768106399789307 
| ArithmeticContract.add_sub_external(uint256,uint256,uint256)             | 171.91                    | 194.66             | 0.8831295592314805 
| ArithmeticCallTest.test_double_add_sub_external(uint256,uint256,uint256) | 374.52                    | 423.73             | 0.8838647251787695 
| ArithmeticCallTest.test_double_add(uint256,uint256)                      | 295.01                    | 327.36             | 0.9011791300097751 
| ArithmeticCallTest.test_double_add_double_sub(uint256,uint256)           | 430.26                    | 475.71             | 0.9044585987261147 
| CSETest.test_identity(uint256,uint256)                                   | 237.45                    | 260.96             | 0.9099095646842428 
| TOTAL                                                                    | 2532.6899999999996        | 2863.6             | 0.884442659589328  
```

PetarMax:
@anvacaru Could you please compare times between `master` and a version of this branch that just has the Foundry version changed, without the new assertions added? What is suspicious to me is that CSE tests such as `Identity.identity(uint256)` are also slower given that they do not use any cheatcodes.

This PR [has just timed out on 2h30 minutes](https://github.com/runtimeverification/kontrol/actions/runs/8784945298/job/24104750858?pr=507) for the integration tests on runner 25. [This run of another PR](https://github.com/runtimeverification/kontrol/actions/runs/8736499364/job/23971581764) from 4 days ago passes the same integration tests in just over an hour on the same runner. I've updated it to the latest version of Kontrol to see what will happen.

I think we need to get to the bottom of the slowdown instead of increasing CI time. The tests being run just aren't different enough to justify that change for me.

PetarMax:
Given how the expected outputs have changed, the obtained bytecode seems to be larger across the board.

anvacaru:
Yes, it's not optimal. I wanted to bump the `forge-std` to make sure that our current test suite passes with the new way of handling asserts. What I could do instead:
A. keep the previous `forge-std` version that we had, and write new tests for `vm.assertEq`s. 
B. keep the previous `forge-std` version for the `foundry` directory, and have a new project that uses the latest `forge-std` with some simple tests.

palinatolmach:
I find it so strange that there's a slowdown now — native assertions should take less time, I think, because now we're not executing an additional function call, and also the revert happens earlier if the assertion fails. Is it coming from just a new Foundry version? I haven't had enough time to look into it today (sorry about that), but I'll do a review tomorrow!

ehildenb:
We should drill into why the slowdown, I think a near 20% performance degradation is probably too much.
Maybe by creating before and after bug reports we can ask the HB team to investigate a bit?

.github/workflows/test-pr.yml
```diff
@@ -155,6 +155,37 @@ jobs:
...
           docker stop --time=0 kontrol-ci-integration-${GITHUB_SHA}
 
+  end-to-end-tests:
+    needs: code-quality-checks
+    name: 'End to End Sanity Tests'
+    runs-on: [self-hosted, linux, normal, fast]
+    strategy:
+      fail-fast: false
+      matrix:
+        backend: ['booster']
```
ehildenb, line 165:
```suggestion
```

It looks like this matrix isn't used at all, we just run with the same `TEST_ARGS` either way (only with the booster).

src/kontrol/__main__.py
```diff
@@ -856,9 +840,16 @@ def parse(s: str) -> list[T]:
...
-            kontrol_cli_args.foundry_args,
         ],
     )
+    init.add_argument(
+        '--name',
+        dest='project_root',
+        default=Path.cwd(),
+        type=Path,
+        help='Name of the project to be initialized. If missing, the current directory is used.',
+    )
```
ehildenb, line 851: Why `--name`? Does that match `forge init` behavior? I guess I would expect the same argument `--project-root` for this.

Also, I think defaults are supposed to be set in teh `*Options` classes, not here. @nwatson22 could say more.

ehildenb: It looks ilke `forge init DIR` is more the way. So maybe we just make it an optional positional argument? https://stackoverflow.com/questions/4480075/argparse-optional-positional-arguments

```suggestion
    init.add_argument(
        dest='project_root',
        nargs='?',
        default=None,
        type=Path,
        help='Name of the project to be initialized. If missing, the current directory is used.',
    )
```

anvacaru:
This morning I did an investigation on the slowdowns. It seems that they are not related to this PR, but with the difference between forge std versions.

I've used a fresh project from `forge init`, and using the example:
```solidity
    function test_Increment() public {
        counter.increment();
        assert(counter.number() == 1);
    }

```

#### With the kontrol `assert.md` branch:

- using `assert`:
  1. "version": "1.7.6": 55s
  2. "version": "1.6.0": 31s
- with `assertEq`:
  1. "version": "1.7.6": 56s
  2. "version": "1.6.0": 32s


#### With the kontrol `master` branch:

- using `assert`:
  1. "version": "1.7.6": 55s
  2. "version": "1.6.0": 32s
- using `assertEq`: NA as `master` does not support `assertEq` on version 1.7.6

.github/workflows/test-pr.yml
anvacaru: Addressed in https://github.com/runtimeverification/kontrol/pull/507/commits/3053c358d30f720fb1b8beddd9519e55fafc12e4.

src/kontrol/__main__.py
anvacaru: Addressed in https://github.com/runtimeverification/kontrol/pull/507/commits/b466e8af92712414db09cf3952c9fb993df90d54.

anvacaru:
Opened a new issue to investigate the slowdown here: https://github.com/runtimeverification/kontrol/issues/534

anvacaru:
>If slowdown is not related to this PR, then we should be able to make a version of this PR that doesn't contain the slowdown. This PR is about bumping Forge std version, so the slowdown is about this PR.

This PR is adding support to `assertEq` like cheat codes. The debate is about which is the best way to test them on CI. I tried bumping the `forge-std` version and then we noticed the CI timing out and the general slowdown of the new version. So I'd say that the support of native assertions and the slowdown identified are two different issues.

Currently this PR does **not** bump the version of the library used in our integration tests. Instead, I added the new stage `end-to-end` to run an entire process of creating a new project with the latest version of the library and test the new assert cheat codes. This way we can have a stage that will always run with the latest versions and we can identify failures/slowdowns faster. Also, this stage does not have to be a `required` stage for merging a pull request.

> Why a new test directory end_to_end? I think just adding a new test called test_end_to_end.py to test-integration should be fine?

 I needed a different fixture for foundry and this seemed the straightforward way.

Another way would be to keep adding tests  to our current integration test-suite. The test would have to encode the calldata manually, as the `vm` contract in the current version of forge-std does not have  signatures defined for asserts. I could write something like
```solidity
        (bool success, ) = vm.call(
            abi.encodeWithSignature("assertEq(uint256,uint256)", foo, baz)
        );
```

ehildenb:
Ok, makes sense that we need a different foundry fixture.

palinatolmach (approved):

src/kontrol/foundry.py
```diff
@@ -1301,3 +1302,28 @@ def foundry_node_printer(
     if type(proof) is APRProof:
         return FoundryAPRNodePrinter(foundry, contract_name, proof, omit_unstable_output=omit_unstable_output)
     raise ValueError(f'Cannot build NodePrinter for proof type: {type(proof)}')
+
+
+def init_project(skip_forge: bool, project_root: Path) -> None:
```
tothtamas28, line 1307:
```suggestion
def init_project(project_root: Path, *, skip_forge: bool) -> None:
```

src/tests/end-to-end/conftest.py
tothtamas28, line 1: Why not put these under `integration`?

src/tests/end-to-end/conftest.py
```diff
@@ -0,0 +1,64 @@
...
+        interim_simplification=None,
+        no_post_exec_simplify=None,
+    )
+
+
+TEST_DATA_DIR: Final = (Path(__file__).parent / 'test-data').resolve(strict=True)
+
+
+@pytest.fixture(scope='session')
+def foundry(foundry_root_dir: Path | None, tmp_path_factory: TempPathFactory, worker_id: str) -> Foundry:
```
tothtamas28, line 44: Is this copy-pasted code? If so, the duplication has to be factored out.

src/kontrol/foundry.py
anvacaru: Done in https://github.com/runtimeverification/kontrol/pull/507/commits/ebaa18e2adf5f1d10560cdc17a94e1071e96d817

src/tests/end-to-end/conftest.py
anvacaru: @tothtamas28 The two fixtures are a bit different.

- `end-to-end` initializes a new project without dependency management and using the latest versions available, while `integration`installs in the current `foundry` project specific versions of forge-std and kontrol-cheatcodes.
-  the `foundry_kompile` options are simplified.
- project directories and the specific files copied differ,because the setups are for different testing scopes.

It should be possible to merge these two fixtures together. But I do not know on which condition to branch the behaviours.

tothtamas28: The fixture could also return a function that initializes the project under some (by default, temporary) directory. Would that make it easier to remove some duplication?

.github/workflows/test-pr.yml
```diff
@@ -116,7 +116,7 @@ jobs:
             docker exec -u github-user kontrol-ci-integration-${GITHUB_SHA} /bin/bash -c 'CXX=clang++-14 poetry run kdist --verbose build -j`nproc` evm-semantics.haskell kontrol.foundry'
       - name: 'Run integration tests'
         run: |
-          TEST_ARGS='--numprocesses=8 -vv -k "not test_kontrol_cse and not test_foundry_minimize_proof"'
+          TEST_ARGS='--numprocesses=8 -vv -k "not test_kontrol_cse and not test_foundry_minimize_proof and not test_kontrol_end_to_end"'
```
tothtamas28, line 119:
```suggestion
          TEST_ARGS='--numprocesses=8 -vv -k "not (test_kontrol_cse or test_foundry_minimize_proof or test_kontrol_end_to_end)"'
```

.github/workflows/test-pr.yml
```diff
@@ -128,10 +128,6 @@ jobs:
     needs: code-quality-checks
     name: 'CSE Tests'
     runs-on: [self-hosted, linux, normal, fast]
-    strategy:
-      fail-fast: false
-      matrix:
-        backend: ['booster']
```
tothtamas28, line 134: Is this intentional?

src/kontrol/foundry.py
```diff
@@ -1301,3 +1302,28 @@ def foundry_node_printer(
...
+    cwd = Path.cwd()
+    chdir(root)
+    write_to_file(Path('lemmas.k'), empty_lemmas_file_contents())
+    write_to_file(Path('KONTROL.md'), kontrol_file_contents())
+
+    run_process(
+        ['forge', 'install', '--no-git', 'runtimeverification/kontrol-cheatcodes'],
+        logger=_LOGGER,
+    )
+    chdir(cwd)
```
tothtamas28, line 1329:
```suggestion
    write_to_file(root / 'lemmas.k', empty_lemmas_file_contents())
    write_to_file(root / 'KONTROL.md', kontrol_file_contents())
    run_process(
        ['forge', 'install', '--no-git', 'runtimeverification/kontrol-cheatcodes'],
        logger=_LOGGER,
        cwd=root,
    )
```

tothtamas28 (approved):

Issue #281, opened 2024-01-16
Extend the use of custom errors and exit codes


palinatolmach:
Related: https://github.com/runtimeverification/kontrol/pull/279

As @tothtamas28 suggested, we can start building a library of custom exceptions, and, for CLI usage, we can then catch them, print some error message, then exit with the corresponding exit code. We can also have a catch-all case for unhandled exceptions, for which we can do some extra logging and perhaps ask the user to submit a bug report. 

This would also help us capture errors more accurately. For example, we should throw an error instead of exiting with code `1` in [the following case](https://github.com/runtimeverification/kontrol/blob/fff05884eef5f61d92dffa2d0d7cb1116513dab3/src/kontrol/foundry.py#L251) (right now, if the caller wants to handle that exception (e.g. by translating to an error message on the output), they will be forced to catch `SystemExit`):
```
def build(self) -> None:
    try:
        run_process(['forge', 'build', '--root', str(self._root)], logger=_LOGGER)
    except FileNotFoundError:
        print("Error: 'forge' command not found. Please ensure that 'forge' is installed and added to your PATH.")
        sys.exit(1)
```

palinatolmach:
I don't think it's blocked by anything, other than us having to carefully consider how to do it properly. It would be good to have, but it's not urgent — I'd move it to backlog (cc @yale-vinson).

PR #510, opened 2024-04-09
Initial refactoring of exit codes

palinatolmach:
Addresses https://github.com/runtimeverification/kontrol/issues/281.

As discussed [on slack](https://runtimeverification.slack.com/archives/C03NP1UP2KZ/p1712427281409899) with @F-WRunTime and @tothtamas28, to facilitate the use of Kontrol on CI, this PR makes the following changes: 
- changes exit code for a `kontrol prove` run with failed tests from the number of failed tests to `1`
- throws a `RuntimeError` if a `forge` invocation fails, instead of returning `1` as before

### DIFF
--- a/src/kontrol/__main__.py
+++ b/src/kontrol/__main__.py
@@ -70,11 +70,10 @@ def _load_foundry(foundry_root: Path, bug_report: BugReport | None = None, use_h
     try:
         foundry = Foundry(foundry_root=foundry_root, bug_report=bug_report, use_hex_encoding=use_hex_encoding)
     except FileNotFoundError:
-        print(
+        raise RuntimeError(
             f'File foundry.toml not found in: {str(foundry_root)!r}. Are you running kontrol in a Foundry project?',
             file=sys.stderr,
         )
-        sys.exit(1)
     return foundry
 
 
@@ -321,7 +320,6 @@ def exec_prove(
         include_summaries=include_summaries,
         xml_test_report=xml_test_report,
     )
-    failed = 0
     for proof in results:
         _, test = proof.id.split('.')
         if not any(test.startswith(prefix) for prefix in ['test', 'check', 'prove']):
@@ -333,7 +331,6 @@ def exec_prove(
             print(f'PROOF PASSED: {proof.id}')
             print(f'time: {proof.formatted_exec_time()}')
         else:
-            failed += 1
             print(f'PROOF FAILED: {proof.id}')
             print(f'time: {proof.formatted_exec_time()}')
             failure_log = None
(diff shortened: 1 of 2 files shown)


src/kontrol/__main__.py
```diff
@@ -348,7 +344,7 @@ def exec_prove(
                 print(f'The proof cannot be completed while there are refuted nodes: {refuted_nodes}.')
                 print('Either unrefute the nodes or discharge the corresponding refutation subproofs.')
 
-    sys.exit(failed)
+    sys.exit(1)
```
tothtamas28, line 347:
```suggestion
    sys.exit(1 if failed else 0)
```

src/kontrol/__main__.py
```diff
@@ -69,12 +69,10 @@ def _ignore_arg(args: dict[str, Any], arg: str, cli_option: str) -> None:
...
     try:
         foundry = Foundry(foundry_root=foundry_root, bug_report=bug_report, use_hex_encoding=use_hex_encoding)
-    except FileNotFoundError:
-        print(
+    except FileNotFoundError as err:
+        raise RuntimeError(
             f'File foundry.toml not found in: {str(foundry_root)!r}. Are you running kontrol in a Foundry project?',
-            file=sys.stderr,
-        )
-        sys.exit(1)
```
tothtamas28, line 77: This was introduced in #204, which fixed #93. In this PR, I'd just change the exit code.

src/kontrol/foundry.py
```diff
@@ -260,9 +260,10 @@ def custom_view(self, contract_name: str, element: KCFGElem) -> Iterable[str]:
     def build(self) -> None:
         try:
             run_process(['forge', 'build', '--build-info', '--root', str(self._root)], logger=_LOGGER)
-        except FileNotFoundError:
-            print("Error: 'forge' command not found. Please ensure that 'forge' is installed and added to your PATH.")
-            sys.exit(1)
```
tothtamas28, line 265: On the other hand, I agree with this change: `sys.exit` should only be called from `__main__` (or the CLI layer in general).

src/kontrol/__main__.py
palinatolmach: Oh, right, thank you @tothtamas28!

src/kontrol/__main__.py
palinatolmach: Right, thank you! I've reverted this change and instead changed the exit code to `127` since it's a "file not found" issue: https://github.com/runtimeverification/kontrol/pull/510/commits/9d8e7400726f2a43fe2bfb70d24ce28b6cfd27a1.

tothtamas28 (approved):

PR #519, opened 2024-04-17
support for `expectRevert(bytes4)`

### LINKED ISSUE #466: Support `expectRevert(bytes4)` cheatcode
We should add support for `expectRevert(bytes4)` cheatcode.

ovatman:
Resolves #466 

This PR contains

- matching `expectRevert(bytes4)` selector in cheatcode support
- After revert happens, when checking to clear the revert previous version was checking the whole output with the expected reason except the initial 4 bytes. This version just checks the last 32 bytes which contain the whole selector information (regardless of the type of parameter)


anvacaru:
There are at least two tests that are skipped because we didn't have support for this cheatcode. Can you remove them from the `*-skip` and `*-skip-legacy` lists?
- `testFail_expectRevert_bytes4`
- `test_expectRevert_bytes4`

src/kontrol/kdist/cheatcodes.md
```diff
@@ -1171,7 +1172,7 @@ Will also return true if REASON is `.Bytes`.
     syntax Bool ::= "#matchReason" "(" Bytes "," Bytes ")" [function, klabel(foundry_matchReason)]
  // ----------------------------------------------------------------------------------------------
     rule #matchReason(REASON, _) => true requires REASON ==K .Bytes
-    rule #matchReason(REASON, OUT) => REASON ==K #range(OUT, 4, lengthBytes(OUT) -Int 4) requires REASON =/=K .Bytes
+    rule #matchReason(REASON, OUT) => #range(REASON, lengthBytes(REASON) -Int 32, 32) ==K #range(OUT, lengthBytes(OUT) -Int 32, 32) requires REASON =/=K .Bytes
```
anvacaru, line 1175: Before approving, could you double check that this the same behaviour that forge has for `expectRevert` ? [link to rust implementation](https://github.com/foundry-rs/foundry/blob/master/crates/cheatcodes/src/test/expect.rs#L494-L546)

src/kontrol/kdist/cheatcodes.md
```diff
@@ -1080,6 +1084,20 @@ Utils
          </expectedRevert>
 ```
 
+- `#setExpectRevertBytes4` sets the `<expectedRevert>` subconfiguration with the current call depth and the expected message from `expectRevert`. When `bytes4` type parameter is used some of the information is not directly included in the `EXPECTED` data since they become constant values, such as the size of the expected output. Here we set them manually to constant values.
```
anvacaru, line 1087: Formatting, one sentence per line.
```suggestion
- `#setExpectRevertBytes4` sets the `<expectedRevert>` subconfiguration with the current call depth and the expected message from `expectRevert`.
When `bytes4` type parameter is used some of the information is not directly included in the `EXPECTED` data since they become constant values, such as the size of the expected output.
Here we set them manually to constant values.
```

src/kontrol/kdist/cheatcodes.md
```diff
@@ -1080,6 +1084,20 @@ Utils
...
+
+```k
+    syntax KItem ::= "#setExpectRevertBytes4" Bytes [klabel(foundry_setExpectRevert)]
+ // ---------------------------------------------------------------------------
```
anvacaru, line 1091: Formatting.
```suggestion
 // ---------------------------------------------------------------------------------
```

anvacaru (approved):
Approved with some formatting comments.

PR #520, opened 2024-04-18
Use parallel prover branch

nwatson22:
- https://github.com/runtimeverification/evm-semantics/pull/2400
- https://github.com/runtimeverification/k/pull/4256

- Enables single-proof parallelism on foundry_prove, which makes `parallel_advance_proof` the default on the kontrol test suite and the `kontrol prove` CLI.
- To use, include the `--max-frontier-parallel N` option with `kontrol prove` to use up to N worker threads to advance independent proof branches in parallel.
- To streamline the creation of a `create_kcfg_explore` function to pass to KEVM's `run_prover`, uses `kore_server` in `_run_cfg_group` instead of `legacy_explore` (so we can get access to the server's port instead of having to copy the port of the client buried deep within the `KCFGExplore`.
- To clean up the logic of choosing between starting a new KoreServer and using the port of a KoreServer that was already started elsewhere, introduces `OptionalKoreServer` and 2 subclasses for each case that each yield the server port and handle closing the underlying server (if applicable).


src/kontrol/prove.py
```diff
@@ -316,7 +318,40 @@ def init_and_run_proof(test: FoundryTest) -> APRFailureInfo | Exception | None:
             start_server=start_server,
             port=options.port,
             maude_port=options.maude_port,
+            haskell_threads=options.max_frontier_parallel,
         ) as kcfg_explore:
+
+            def create_kcfg_explore() -> KCFGExplore:
+                if options.maude_port is None:
```
tothtamas28, line 325: Where does this logic come from? Is this an added feature, or something that got pushed up from `evm-semantics`?

src/kontrol/prove.py
```diff
@@ -316,7 +318,40 @@ def init_and_run_proof(test: FoundryTest) -> APRFailureInfo | Exception | None:
...
+                        'execute': [('localhost', options.maude_port, TransportType.HTTP)],
+                        'simplify': [('localhost', options.maude_port, TransportType.HTTP)],
+                        'add-module': [
+                            ('localhost', options.maude_port, TransportType.HTTP),
+                            ('localhost', kcfg_explore.cterm_symbolic._kore_client.port, TransportType.SINGLE_SOCKET),
+                        ],
+                    }
+                client = KoreClient(
+                    'localhost',
+                    kcfg_explore.cterm_symbolic._kore_client.port,
```
tothtamas28, line 338: This is a roundabout way of getting the server port. Can you think of a way to improve this?

src/kontrol/prove.py
nwatson22: Yes, the `dispatch` construction is pushed up up from `evm-semantics`, from `legacy_explore`.

src/kontrol/prove.py
nwatson22: It's not easy with `legacy_explore` from what I know, since that yields the `KCFGExplore` from inside the `with kore_server(...) as` block, and not the server itself, but I've refactored it so we start the server (if necessary) directly in `init_and_run_proof` and initialize the `KCFGExplore` used for initializing the proof with `create_kcfg_explore()` which I already had here.

ehildenb (approved):

Issue #531, opened 2024-04-25
Kompilation issues for EIP2612 lib


palinatolmach:
`kontrol build` throws errors when running on [permit2](https://github.com/Uniswap/permit2/tree/main) code base that indicate that we're not generating rules for functions based on signatures correctly:
```
[Error] Compiler: Found variable V1_amount_0 on right hand side of rule, not
bound on left hand side. Did you mean "?V1_amount_0"?
        Source(/out/kompiled/contracts.k)
        Location(19150,37,19150,48)
        19150 |        andBool ( #rangeUInt ( 256 , V1_amount_0 )
              .                                     ^~~~~~~~~~~
``` 

In addition, as reported by @paulslen, `kontrol prove` fails when running on a contract implementing EIP2612 due to the existence of symbolic `CHAIN_ID` in the contract bytecode (`block.chainId` is usually accessed in the constructor), leading to the constraints looking like this:
```
#And { 239 #Equals b"`\x80... , lengthBytes ( #padToWidth ( 32 , #asByteStack ( CHAINID_CELL:Int ) ) ) +Int 160 , 32 -Int minInt ( 32 , lengthBytes ( #padToWidth ( 32 , #asByteStack ( CHAINID_CELL:Int ) ) ) ) ) +Bytes ... [ 0 ] } 
```

palinatolmach:
It seems to be caused by the problem reported in https://github.com/orgs/runtimeverification/projects/23/views/7?pane=issue&itemId=56780327.

PR #542, opened 2024-05-03
Refactoring flattening for dynamic types in rules

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/531.

As described in the comments below in more detail, we need to refactor the way we process function inputs to use them during ABI calldata rule generation. This PR makes the following changes: 
- Introduces [`_flatten_tuple`](https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L140) and [`flatten_array`](https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L225) functions that recursively unwrap components of tuple and array inputs, respectively. 
- This allows refactoring of [`Input.flattened()`](https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L251) and [`Method.flat_inputs()`](https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L506) functions, which can now be used to correctly generate `arg_types` and `arg_names` without repetitively unwrapping elements of complex types in both functions:
https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L518-L519
This gets `arg_types` and `arg_names` back to their original implementation before we started introducing dynamic type support, which required unwrapping/flattening of arrays and tuples in https://github.com/runtimeverification/kontrol/pull/321. 
It also helps factor out the tuple flattening logic for rule generation in [`rule()`](https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L582):

https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L594-L596

- Updates the logic for assigning indexes (e.g., `V0`, `V1`, ...) for tuple and array elements in [`_unwrap_components`](https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L163) for tuple components, and function inputs in general in [inputs_to_abi() ](https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/kontrol/solc_to_k.py#L260). The update ensures that indexing accounts for possible structs and nested structs, e.g., for
```
test_dynamic_nested_struct_array(((uint256 id,bytes content)[],uint256 nonce))
```
the inputs would look like
```
V0_id_0 : uint256 , V1_content_0 : bytes , V0_id_1 : uint256 , V1_content_1 : bytes , V2_nonce : uint256
```
- Adds an integration test with a `tuple` containing a `tuple[]` component:
```solidity
    function test_dynamic_nested_struct_array(ComplexNestedType memory cntValues) public {
        require(cntValues.values.length == 2, "DynamicTypes: invalid default length for ComplexType[] in ComplexNestedType");
    }
```
- Adds a unit test for the updated `flattened()` function for Inputs: https://github.com/runtimeverification/kontrol/blob/16165c65eb8a2e8893970f84b53c99df1f538236/src/tests/unit/test_solc_to_k.py#L563-L568

One further improvement that could be done is factoring out common logic between tuple/array flattening and generating ABI-type objects for them, as described in https://github.com/runtimeverification/kontrol/pull/542#issuecomment-2095830729. I would prefer to leave that for the next PR to unblock the progress on a potential client demo, and I'd also like to check with Andrei before making a more significant change.


palinatolmach:
As manifested in https://github.com/runtimeverification/kontrol/issues/531, we need some refactoring of the way we handle function inputs and use them for ABI calldata rule generation. The rule for  `test_dynamic_nested_struct_array` defined as 
```
contract DynamicTypesTest is Test {
    struct ComplexType {
        uint256 id;
        bytes content;
    }

    struct ComplexNestedType {
        ComplexType[] values;
        uint256 nonce;
    }

    function test_dynamic_nested_struct_array(ComplexNestedType memory cntValues) public pure {
        require(
            cntValues.values.length == 2, "DynamicTypes: invalid default length for ComplexType[] in ComplexNestedType"
        );
    }
}
```
should be 
```
rule  ( S2KtestZModDynamicTypesTest . S2KtestZUnddynamicZUndnestedZUndstructZUndarray 
 ( V0_id_0 : uint256 , V1_content_0 : bytes , V0_id_1 : uint256 , V1_content_1 : bytes , V1_nonce : uint256 ) => #abiCallData ( "test_dynamic_nested_struct_array" , ( #tuple ( ( #array ( #tuple ( ( #uint256 ( V0_id_0 ) , ( #bytes ( V1_content_0 ) , .TypedArgs ) ) ) , 2 , ( #tuple ( ( #uint256 ( V0_id_0 ) , ( #bytes ( V1_content_0 ) , .TypedArgs ) ) ) , ( #tuple ( ( #uint256 ( V0_id_1 ) , ( #bytes ( V1_content_1 ) , .TypedArgs ) ) ) , .TypedArgs ) ) ) , ( #uint256 ( V1_nonce ) , .TypedArgs ) ) ) , .TypedArgs ) ) )
  ensures ( #rangeUInt ( 256 , V0_id_0 )
  andBool ( #rangeUInt ( 64 , lengthBytes ( V1_content_0 ) )
  andBool ( #rangeUInt ( 256 , V0_id_1 )
  andBool ( #rangeUInt ( 64 , lengthBytes ( V1_content_1 ) )
  andBool ( #rangeUInt ( 256 , V1_nonce )
 )))))

```
— since the default array length is 2, we want to have two instances of `ComplexType` tuples (`V0_id_0`, `V1_content_0`, `V0_id_1`, `V1_content_1`). While, at the moment, we generate 
```
    rule  ( S2KtestZModDynamicTypesTest . S2KtestZUnddynamicZUndnestedZUndstructZUndarray 
( V0_id : uint256 , V1_content : bytes , V1_nonce : uint256 ) => #abiCallData ( "test_dynamic_nested_struct_array" , ( #tuple ( ( #array ( #tuple ( ( #uint256 ( V0_id_0 ) , ( #bytes ( V1_content_0 ) , .TypedArgs ) ) ) , 2 , ( #tuple ( ( #uint256 ( V0_id_0 ) , ( #bytes ( V1_content_0 ) , .TypedArgs ) ) ) , ( #tuple ( ( #uint256 ( V0_id_1 ) , ( #bytes ( V1_content_1 ) , .TypedArgs ) ) ) , .TypedArgs ) ) ) , ( #uint256 ( V1_nonce ) , .TypedArgs ) ) ) , .TypedArgs ) ) )
       ensures ( #rangeUInt ( 256 , V0_id_0 )
       andBool ( #rangeUInt ( 64 , lengthBytes ( V1_content_0 ) )
       andBool ( #rangeUInt ( 256 , V0_id_1 )
       andBool ( #rangeUInt ( 64 , lengthBytes ( V1_content_1 ) )
       andBool ( #rangeUInt ( 256 , V1_nonce )
               )))))
```
— In other words, we don't unwrap array elements for `ComplexType[]` properly on the left hand side:
`( V0_id : uint256 , V1_content : bytes , V1_nonce : uint256 )`
vs
`( V0_id_0 : uint256 , V1_content_0 : bytes , V0_id_1 : uint256 , V1_content_1 : bytes , V1_nonce : uint256 )`)
which leads to the kompilation errors.

palinatolmach:
There are two problems: 
- `flattened()` function currently doesn't account for possible struct elements of an array type (hence, the issue https://github.com/runtimeverification/kontrol/issues/531):
https://github.com/runtimeverification/kontrol/blob/2bdfd3bc9b34a019c70d68055531f13b7ca53f32/src/kontrol/solc_to_k.py#L197
It is also not used meaningfully in the code.

- the same code "unwrapping" elements of dynamic types (i.e., transforming `array[]` into `array_0, array_1` or `tuple(uint a, address b)` into `uint a, address b` is repeated several times throughout the implementation:
  - when converting an Input into an ABI type: https://github.com/runtimeverification/kontrol/blob/2bdfd3bc9b34a019c70d68055531f13b7ca53f32/src/kontrol/solc_to_k.py#L170
  - when generating a set of argument (input) names: https://github.com/runtimeverification/kontrol/blob/2bdfd3bc9b34a019c70d68055531f13b7ca53f32/src/kontrol/solc_to_k.py#L452
  - when generating a list of argument (input) types: https://github.com/runtimeverification/kontrol/blob/2bdfd3bc9b34a019c70d68055531f13b7ca53f32/src/kontrol/solc_to_k.py#L467
  - when generating type range constraints for the ABI calldata rule: https://github.com/runtimeverification/kontrol/blob/2bdfd3bc9b34a019c70d68055531f13b7ca53f32/src/kontrol/solc_to_k.py#L550

That makes me think that the issue can be fixed together with factoring out the code for flattening (or unwrapping) the elements, and perhaps that can be incorporated into the `flattened()` function that can then be used to generate the unpacked list of inputs throughout `solc_to_k`.

palinatolmach:
I have pushed the draft implementation that fixes the issue in https://github.com/runtimeverification/kontrol/issues/531 in https://github.com/runtimeverification/kontrol/pull/542/commits/8621d5d81be5628a1c1cb0c1f0bac4aac77505ca. It, however, can be refactored further since there're two almost equivalent functions: one used to flatten array inputs for the rule generation:
```solidity
def flatten_array(self) -> list[Input]:
    """
    Recursively unwraps components of an array.
    
    The 'array_index' parameter is used to uniquely identify elements within an array of tuples.
    TODO: Add support for nested arrays and use the entire 'input.array_lengths' array
    instead of only the first value.
    """
    base_type = self.type.rstrip('[]')
    if self.array_lengths is None:
        raise ValueError(f'Array length bounds missing for {self.name}')
    array_length = self.array_lengths[0]
    array_elements: list[Input] = []

    if base_type == 'tuple':
        array_elements = [Input.flatten_tuple(self.components, index) for index in range(array_length)]
    else:
        array_elements = [
            Input(f'{self.name}_{n}', base_type, idx=self.idx).flattened() for n in range(array_length)
        ]
    return array_elements

def flattened(self, array_index: int | None = None) -> list[Input]:
    if self.type.endswith('[]'):
        return self.flatten_array()
    elif self.type == 'tuple':
        return Input.flatten_tuple(self.components, array_index)
    else:
        return [self]
```
and one for generating an ABI-type object for an array (and each of its elements recursively):
```solidity
def make_single_type(self) -> KApply:
    """
    Generates a KApply representation for a single type input.

    It handles arrays, nested arrays, and base types.
    For 'tuple[]', it calls '_make_tuple_type' for each element in the array.
    For arrays, it generates a list of Input objects numbered from 0 to `array_length`.
    TODO: Add support for nested arrays and use the entire 'input.array_lengths' array
    instead of only the first value.
    """
    if self.type.endswith('[]'):
        base_type = self.type.rstrip('[]')
        if self.array_lengths is None:
            raise ValueError(f'Array length bounds missing for {self.name}')
        array_length = self.array_lengths[0]
        array_elements: list[KInner] = []
        if base_type == 'tuple':
            array_elements = [Input._make_tuple_type(self.components, index) for index in range(array_length)]
        else:
            array_elements = [
                Input(f'{self.name}_{n}', base_type, idx=self.idx).make_single_type() for n in range(array_length)
            ]

        # THE PRIMARY DIFFERENCE IS HERE:

        return KEVM.abi_array(
            array_elements[0],
            intToken(array_length),
            array_elements,
        )

    else:
        return KEVM.abi_type(self.type, KVariable(self.arg_name))
```

The problem I'm facing when factoring the common code is that, when generating ABI objects in the latter function, 
```solidity
KEVM.abi_array(
    array_elements[0],
    intToken(array_length),
    array_elements,
)
```
should be called recursively on the elements of a nested array. When using the first, `flatten_array()` function, the information about an array element type is lost, so we need a way to keep it in the refactored version, so we can generate a rule with types being, e.g., `#tuple(#array(#array(...), ...), ...)`.

I spent the last couple of days thinking about the best solution, but didn't come up with one so far. Would it make sense to introduce some polymorphism for `Input`s, and introduce `Array`, `Tuple`, `Value` classes for different inputs, and define a factory that would generate these more specific input objects for the `solc`-provided ones? Right now, they are represented by the same [`Input`](https://github.com/runtimeverification/kontrol/blob/8621d5d81be5628a1c1cb0c1f0bac4aac77505ca/src/kontrol/solc_to_k.py#L74) class with the same set of fields. That'd make it easier to define `flattened()` and `to_abi()` functions for each of them separately, but it also gets less trivial with the `tuple[]` type involved.

Am I missing some design pattern that could be used here? Would appreciate any advice @tothtamas28 @nwatson22 @ovatman!

tothtamas28 (commented):
Could you add a few unit tests please? So basically a few input - expected output pairs that demonstrate where the existing implementation falls short, and how this PR fixes the problem.

palinatolmach:
Thanks @tothtamas28, will add some unit tests tomorrow! The primary change is in the behavior of `arg_names(self)`, which is a `Method`'s property, but I'll think of a way to test it:
https://github.com/runtimeverification/kontrol/blob/ad67aa8dbae31e88210dd69d3996f7d26fb1663d/src/kontrol/solc_to_k.py#L508-L509

palinatolmach:
I've pulled the minimized version to https://github.com/runtimeverification/kontrol/pull/546 to make it easier to review and merge in order to unblock me and @ovatman, who's also experiencing this issue. I'd leave further refactoring as a follow up after the demo and when Andrei is back to provide feedback, if that's ok.

palinatolmach:
Closing for now, will reopen when/if we get to refactor the input types.

PR #543, opened 2024-05-06
Don't run `BMCLoopsTest` twice

geo2a:
When working on the performance script for Kontrol proofs which we use in the haskell backend, I've noticed that we're [running](https://github.com/runtimeverification/haskell-backend/pull/3845#issuecomment-2093502414) `BMCLoopsTest` twice: once as part of `test_foundry_prove`, and then  again in `test_foundry_bmc`. This PR makes it so we only run it as part of `test_foundry_bmc`.


palinatolmach:
Thanks for catching and fixing @geo2a!

palinatolmach (approved):

Issue #556, opened 2024-05-13
Skipping native assertions in expectRevert


anvacaru:
According to the foundry book, [after calling expectRevert, calls to other cheatcodes before the reverting call are ignored.](https://book.getfoundry.sh/cheatcodes/expect-revert?highlight=expectRe)

However, when running a test like 
```Solidity
    function test_assertTrue_string() public {
        string memory err = "throw test";
        vm.expectRevert(bytes(err));
        assertTrue(false, err);
    }
```

the `vm.assertTrue` call is executed, while in kontrol it is skipped.
![image](https://github.com/runtimeverification/kontrol/assets/16517508/3a021b8e-4afb-4abe-93ca-5af6c0a86953)

jinxinglim:
I have "forwarded" this issue to `forge-std` repository: https://github.com/foundry-rs/forge-std/issues/565. If there is no reply from them in a few days, I will follow up and text them in their telegram support group.

jinxinglim:
### Update from Foundry team as of 10 June 2024

See comment: https://github.com/foundry-rs/foundry/issues/8049#issuecomment-2157102998

#### Reason for `expectRevert` to catch `assertTrue`

"""
`expectRevert` catches reverts on its depth or above. So, in your example it does not expect `assertTrue` call on depth 1 to revert, however, the revert of `assertTrue` is being propagated by Solidity to revert entire test contract call, thus causing a revert on depth 0 which `expectRevert` catches.
"""

#### Solution

"""
I see how this is confusing though, and this indeed will be fixed by v1 `expectRevert` changes
"""

palinatolmach:
Moving to "Blocked" until the fix is implemented in Foundry and we can confirm that the behavior is consistent with Kontrol.

jinxinglim:
I have updated the foundry version to https://github.com/foundry-rs/foundry/tree/nightly-7226aa06bbb473bfa7fabc0c92215c0938017d0c and tested this test function again

```solidity
    function test_assertTrue_string() public {
        string memory err = "throw test";
        vm.expectRevert(bytes(err));
        assertTrue(false, err);
    }
```

and it failed due to the following reason:
```
Ran 1 test for test/MinimalTest.t.sol:MinimalTest
[FAIL. Reason: call did not revert as expected] test_assertTrue_string() (gas: 14192)
Logs:
  Error: throw test
  Error: Assertion Failed

Traces:
  [14192] MinimalTest::test_assertTrue_string()
[...]

PR #558, opened 2024-05-14
Support more native assertions.

anvacaru:
- modify the `#assert` Kproduction to take a `bytes` of the error message from the call data.
 - add a skip file for end to end tests.
 - related issues:
 -  - #556  => skipping tests that use expectRevert and asserts in the end-to-end stage
 -  - #559  => postponing the implementation of `assertGt` `assertGe` `assertLt` `assertLe` with `int256` types.

Left to implement:
1. dynamic types `string` `bytes` `uint256[]` `int256[]` `bool[]` `bytes32[]` `address[]` for `assertEq` and `assertNotEq`.
2. `assertApproxEqAbs` and `assertApproxEqRel`: take an additional `maxDelta` argument.
3. `assert*Decimals` : take an additional `decimals` value used to format the values in error messages.


src/tests/integration/test-data/src/Unit.t.sol
```diff
@@ -4,44 +4,173 @@ pragma solidity ^0.8.13;
...
+    function test_assertEq(int256 a, int256 b) public pure {
+        vm.assume(a == b);
+        assertEq(a, b);
+    }
+
+    function test_assertEq_err() public {
+        string memory expected = "throw test: 11 != 121";
+        string memory err = "throw test";
+        vm.expectRevert(bytes(expected));
+        assertEq(uint256(11), 121, err);
```
palinatolmach, line 35: Is it worth adding a test that checks that a failing `assertEq` not taking an `err` argument reverts and returns `"assertion failed"`?

palinatolmach (approved):
Approved with a small comment!

src/tests/integration/test-data/src/Unit.t.sol
anvacaru: Good idea! I'll add the test now.

anvacaru: Added the test in commit https://github.com/runtimeverification/kontrol/pull/558/commits/2d15e71fad09a071a32a0833e92cef53a9903862. I also added it to the skipped list because it is blocked by https://github.com/runtimeverification/kontrol/issues/556.

src/kontrol/kdist/assert.md
```diff
@@ -11,101 +11,196 @@ module KONTROL-ASSERTIONS
...
+         <k> #cheatcode_call SELECTOR ARGS => #assert_not_eq #asWord(#range(ARGS, 0, 32)) #asWord(#range(ARGS, 32, 32)) String2Bytes("assertion failed") ... </k>
+      requires SELECTOR ==Int selector ( "assertNotEq(address,address)" )
+        orBool SELECTOR ==Int selector ( "assertNotEq(bool,bool)"       )
+        orBool SELECTOR ==Int selector ( "assertNotEq(bytes32,bytes32)" )
+        orBool SELECTOR ==Int selector ( "assertNotEq(int256,int256)"   )
+        orBool SELECTOR ==Int selector ( "assertNotEq(uint256,uint256)" )
+    [preserves-definedness]
+
+    rule [cheatcode.call.assertNotEq.err]:
+         <k> #cheatcode_call SELECTOR ARGS => #assert_not_eq #asWord(#range(ARGS, 0, 32)) #asWord(#range(ARGS, 32, 32)) #range(ARGS, 96, #asWord(#range(ARGS, 64, 32))) ... </k>
```
jinxinglim, line 83: Hi @anvacaru. Quick question: How is the `96` from `#range(ARGS, 96, #asWord(#range(ARGS, 64, 32)))` derived? Thank you.

anvacaru: For any of the `assertNotEq` selectors in the `requires` clause, the first two arguments are static (take `uint256`, for example). This means that each argument will take 32 bytes in the calldata. So `#range(ARGS, 0, 32)` and `#range(ARGS, 32, 32)` handles that.
The third argument, `string`, is dynamic.  To encode it in the calldata, you first have 32 bytes that represent the `length` of the string, and then `length` bytes representing the actual value.
Given, `#range(ARGS, 96, #asWord(#range(ARGS, 64, 32)))`, you first parse `#asWord(#range(ARGS, 64, 32))` to get the length value, and then read that many bytes starting from index `96` of the calldata.

So, 
```
calldata[0] : arg1, 32 bytes
calldata[32]: arg2, 32 bytes
calldata[64]: length of arg3: 32 bytes
calldata[96]: arg3, `length` bytes.

jinxinglim: Okay. Got it. Thank you!

PR #563, opened 2024-05-14
Executing constructor code  with CSE enabled

nwatson22:
### Additions
- Adds option `--config-type [SUMMARY_CONFIG | TEST_CONFIG]` to `kontrol prove` to control what mode to create the proof(s) in. `SUMMARY_CONFIG` mode is for running functions for CSE summarization and `TEST_CONFIG` mode is for running functions as kontrol tests.
- This mode affects the abstractness of the initial and target node configurations, as well as the behavior when using another proof as setup, as we do for constructor proofs and `setUp` functions.
- `TEST_CONFIG` is the default mode. When CSE is enabled, there are recursive calls to `foundry_prove` for called functions. These calls will always be made with `config_type=SUMMARY_CONFIG`, because the purpose of running proofs for called functions is to create CSE summaries for them.
- Proofs run with `config_type=SUMMARY_CONFIG` will always have KCFG minimization enabled.
- Constructors are no longer grafted in the same way as `setUp` functions. Certain fields are still copied from the constructor final state to the new initial state, but the nodes of the constructor KCFG don't appear in the new KCFG. As such, branching constructors are no longer allowed.
- The constructor bytecode is now only inserted into the `program` cell when creating a constructor proof, not the `code` cell of any account.
- Callvalue is now set for constructors in the same way it was for methods.
- `_update_cterm_from_node` previously assumed `Foundry.address_TEST_CONTRACT()` is in the accounts list, which is only the case under `TEST_CONFIG` mode.
- This accounts cell manipulation logic was also breaking with the presence of the `ACCOUNTS_REST` variable. To handle this, we split those members of the account list into cells and non-cells, with non-cells assumed to be variables.
- Abstracts storage of all accounts in summary mode. The use case for this is when we are executing a function in summary mode using its contract's constructor as a setup proof. We want to use the constructor as a setup proof to get access to immutable and constant variables set in the constructor, but not assume the contract state is as such as no other method has been called since construction. (More might need to be done here.)
- Remove useless constraints at the end of `_update_cterm_from_node`
- To prevent constructors that modify the storage from branching, always sets `static` cell to false for constructors, because constructors are always associated with contract creation, which is non-static.
- When CSE rules are generated and included for multiple functions of the same contract, when the LHS of the respective rules are the same except the `callData`, we needed an additional pair of lemmas to simplify the generated conditions like: `true #Equals b"\xe5\xc1\x9b-\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x158" ~> .K ==K b"\xe5\xc1\x9b-" +Bytes #buf ( 32 , chop ( VV0_x_114b9705:Int ) ) ~> .K` to prevent a branch where the wrong function is called.
- Adds a test in `ConstructorTest` which exercises using constructors alongside CSE. `ConstructorTest.test_contract_call` is run, using the constructor for `ConstructorTest` as setup proof, both in `TEST_CONFIG` mode, and supported by CSE rules generated by running `ImportedContract.add` and `ImportedContract.set`, each with `ImportedContract` constructor as setup proof, all in `SUMMARY_CONFIG` mode.
- On `test_foundry_dependency_automated`, enables `--run-constructor`. functions are either run with `SUMMARY_CONFIG` or `TEST_CONFIG` depending on whether the contract is a "Test" contract.
- Adds `test_contract_call` to `test_foundry_dependency_automated`. Re-enables `ConstructorTest.run_constructor()` in this test.

### Further issues
- We need to make sure this works with the test in https://github.com/runtimeverification/kontrol/pull/573.
- If you run without CSE, I think currently you can have constructors with arguments in called contracts, you just can't pass symbolic variables as those arguments. Having CSE tests now take into account the constructors means that with this PR, you can no longer have constructors with arguments in called contracts (when using `--cse`), for the same reason we already can't have constructors with arguments in top-level/test contracts.
- We might want to abstract other contract account fields, e.g. balance when using constructors as setup proofs.
- We might want to have the return value of the constructor used as the new bytecode rather than simply getting the (predicted) bytecode from the compiled json file, otherwise the immutables will not be set correctly.


nwatson22:
There is currently still a bug with this. When you graft a constructor KCFG to the beginning, this causes a bad circularity module (RHS too general). I discovered this because the circularity rule was conflicting with the CSE-generated dependency rules, which are the same priority, in some cases, causing nondeterministic branches.

src/tests/integration/test-data/show/AssertTest.checkFail_assert_false().expected
```diff
@@ -715,14 +715,13 @@ module SUMMARY-TEST%ASSERTTEST.CHECKFAIL-ASSERT-FALSE():0
        andBool ( 0 <=Int TIMESTAMP_CELL:Int
        andBool ( CALLER_ID:Int =/=Int 645326474426547203313410069153905908525362434349
        andBool ( ORIGIN_ID:Int =/=Int 645326474426547203313410069153905908525362434349
-       andBool ( _CONTRACT_ID =/=Int 645326474426547203313410069153905908525362434349
```
PetarMax, line 718: Do we know why these are disappearing?

src/tests/integration/test-data/show/ConstructorTest.test_contract_call().cse.expected
```diff
@@ -0,0 +1,288 @@
...
+│
+│  (1119 steps)
+└─ 6 (leaf, pending)
+    k: #execute ~> CONTINUATION:K
+    pc: 0
+    callDepth: 0
+    statusCode: STATUSCODE:StatusCode
+
+
+┌─ 7 (root, leaf, target)
```
PetarMax, line 16: Should this proof complete?

src/tests/integration/test-data/show/Identity.applyOp(uint256).cse.expected
```diff
@@ -55,7 +55,7 @@
 
 
 
-module SUMMARY-SRC%CSE%IDENTITY.APPLYOP(UINT256):0
+module SUMMARY-SRC%CSE%IDENTITY.APPLYOP(UINT256):1
```
PetarMax, line 58: This change is a bit worrying, why is there a change in version?

src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
```diff
@@ -1,52 +1,59 @@
 
-┌─ 1 (root, split, init)
+┌─ 1 (root, init)
```
PetarMax, line 2: The `split` should remain, it not being here indicates that the proof has not been minimized.

src/tests/integration/test-data/show/AssertTest.checkFail_assert_false().expected
nwatson22: I added the line
```
 new_init_cterm = new_init_cterm.remove_useless_constraints()
 ```
at the end of `update_node_from_cterm` to get rid of constraints that referred to variables that no longer exist after replacing the contents of certain cells as that function does. I'm pretty sure it comes from that.
I wanted to check with you to make sure that makes sense to do though.

File: src/kontrol/prove.py [hunk 5]
```diff
@@ -680,20 +720,37 @@ def _update_cterm_from_node(cterm: CTerm, node: KCFG.Node, contract_name: str) -
...
+        test_contract_account = new_accounts_map[Foundry.address_TEST_CONTRACT()]
+
+        new_accounts_map[Foundry.address_TEST_CONTRACT()] = CTerm(
+            set_cell(
+                test_contract_account.config,
+                'CODE_CELL',
+                KEVM.bin_runtime(KApply(f'contract_{contract_name}')),
+            ),
+            [],
+        )
```
nwatson22, line 741: I think maybe this whole block can be removed, since we now insert the deployed bytecode during constructor execution to begin with, and the deployed bytecode should be the same between `setUp` function and test function execution.

src/tests/integration/test-data/cse-lemmas.k
```diff
@@ -30,9 +30,22 @@ module CSE-LEMMAS
       requires notBool ( lengthBytes(B1) ==Int lengthBytes(B2) )
       [simplification]
 
+    rule (A ~> B) ==K (C ~> B) => (A ==K C) [simplification]
```
nwatson22, line 33: Check if this is still necessary.

src/kontrol/prove.py
```diff
@@ -716,6 +773,8 @@ def _update_cterm_from_node(cterm: CTerm, node: KCFG.Node, contract_name: str) -
         new_init_cterm = new_init_cterm.add_constraint(constraint)
     new_init_cterm = KEVM.add_invariant(new_init_cterm)
 
+    new_init_cterm = new_init_cterm.remove_useless_constraints()
```
nwatson22, line 776: Is this still needed?

src/kontrol/prove.py
```diff
@@ -680,20 +720,37 @@ def _update_cterm_from_node(cterm: CTerm, node: KCFG.Node, contract_name: str) -
...
+    if config_type == ConfigType.SUMMARY_CONFIG:
+        for account_id, account in new_accounts_map.items():
+            new_accounts_map[account_id] = CTerm(
+                set_cell(
+                    account.config,
+                    'STORAGE_CELL',
+                    abstract_term_safely(account.cell('STORAGE_CELL'), base_name='STORAGE', sort=KSort('Map')),
+                ),
+                [],
+            )
```
nwatson22, line 751: Are there other aspects of the state that can be set up in the constructor but potentially modified by later function calls that we also need to abstract?

src/tests/integration/test-data/show/ConstructorTest.test_contract_call().cse.expected
nwatson22: It is completing now.

src/tests/integration/test-data/show/Identity.applyOp(uint256).cse.expected
nwatson22: This is fixed now. I think it came from me trying to run/update dependencies on multiple test groups with one command.

src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
nwatson22: It's no longer splitting.

src/tests/integration/test-data/show/AssertTest.checkFail_assert_false().expected
PetarMax: Yes, this does make sense, and this constraint is not useful since this is not a CSE test, I believe.

src/tests/integration/test-data/cse-lemmas.k
```diff
@@ -30,9 +30,20 @@ module CSE-LEMMAS
...
       [simplification]
 
+    // Function selector does not match
+    rule { B:Bytes #Equals B1:Bytes +Bytes B2:Bytes } => #Bottom
+      requires #range(B, 0, 4) =/=K #range (B1, 0, 4)
+      [simplification(60), concrete(B, B1)]
+
+    // Function selector matches
+    rule { B:Bytes #Equals B1:Bytes +Bytes B2:Bytes } => {#range(B, 4, lengthBytes(B) -Int 4) #Equals B2}
+      requires #range(B, 0, 4) ==K B1
```
PetarMax, line 40:
```suggestion
      requires 4 <=Int lengthBytes(B) andBool #range(B, 0, 4) ==K B1
```

I have forgotten this constraint, otherwise `#range` will pad with zeros and equality is still possible.

src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
```diff
@@ -6,7 +6,7 @@
 │   statusCode: STATUSCODE:StatusCode
 ┃
 ┃ (branch)
-┣━━┓ constraint: VV0_x_114b9705:Int <=Int ( maxUInt256 -Int #lookup ( CONTRACT_STORAGE:Map , 0 ) )
+┣━━┓ constraint: VV0_x_114b9705:Int <=Int ( maxUInt256 -Int #lookup ( STORAGE_3b3c4496:Map , 0 ) )
```
PetarMax, line 9: How is the suffix `3b3c4496` created, do we know? Will it be stable?

src/tests/integration/test-data/show/ArithmeticCallTest.test_double_add(uint256,uint256).cse.expected
```diff
@@ -202,7 +202,9 @@ module SUMMARY-TEST%ARITHMETICCALLTEST.TEST-DOUBLE-ADD(UINT256,UINT256):0
                          0
                        </balance>
                        <storage>
-                         ( 27 |-> 491460923342184218035706888008750043977755113263 )
+                         ( ( 11 |-> 1 )
+                         ( ( 27 |-> 491460923342184218035706888008750043977755113263 )
+                         ( 7 |-> 1 ) ) )
```
PetarMax, line 207: What do these two new storage slots correspond to?

src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol
```diff
@@ -3,8 +3,34 @@ pragma solidity =0.8.13;
...
 import "forge-std/Test.sol";
 
+contract ImportedContract {
+    uint256 public count;
+
+    constructor() payable {
+        count = 5;
+    }
+
+    function set(uint256 x) public {
```
PetarMax, line 13: Since we have these functions in place, could we also test them, please, as part of `test_kontrol_cse`? `set`, specifically, revealed the issue we had with the resulting storage not being abstracted.

src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol
```diff
@@ -3,8 +3,34 @@ pragma solidity =0.8.13;
...
+        if(count < 3){
+            return;
+        }
+        count = x;
+    }
+
+    function add(uint256 x) public {
```
PetarMax, line 20: Same as for `set` above.

src/tests/integration/test-data/show/ConstructorTest.test_contract_call().cse.expected
```diff
@@ -0,0 +1,309 @@
+
+┌─ 1 (root, init)
+│   k: #execute ~> CONTINUATION:K
+│   pc: 0
+│   callDepth: 0
+│   statusCode: STATUSCODE:StatusCode
+│
+│  (2233 steps)
```
PetarMax, line 8: Do we know that the summaries of `set` and `add` are getting applied for sure?

src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol
```diff
@@ -3,8 +3,34 @@ pragma solidity =0.8.13;
 
 import "forge-std/Test.sol";
 
+contract ImportedContract {
+    uint256 public count;
+
+    constructor() payable {
```
PetarMax, line 9: I think that `payable` here will lead to the call value being symbolic?

src/kontrol/prove.py
PetarMax: Since the rules of the CSE-related outputs haven't changed, it would seem that everything is getting abstracted correctly. We should keep an eye on this.

src/kontrol/prove.py
```diff
@@ -852,13 +896,17 @@ def _create_initial_account_list(
...
     is_test: bool = True,
-    is_setup: bool = False,
     hevm: bool = False,
 ) -> CTerm:
-    final_term = _final_term(empty_config, program, is_test=is_test, is_setup=is_setup)
+    final_term = _final_term(
+        empty_config,
+        program,
+        config_type=config_type,
+    )
```
PetarMax, line 909: This is perhaps a bit silly, but why does this get broken down over four lines when the previous version didn't and this one has fewer characters?

src/kontrol/prove.py
```diff
@@ -891,7 +939,11 @@ def _final_cterm(
     return final_cterm
 
 
-def _final_term(empty_config: KInner, program: KInner, is_test: bool, is_setup: bool) -> KInner:
+def _final_term(
+    empty_config: KInner,
+    program: KInner,
+    config_type: ConfigType,
+) -> KInner:
```
PetarMax, line 946: Same as above for layout.

src/kontrol/prove.py
```diff
@@ -532,35 +545,40 @@ def _method_to_cfg(
...
     trace_options: TraceOptions | None = None,
 ) -> tuple[KCFG, list[int], int, int]:
     calldata = None
     callvalue = None
 
+    contract_code = KEVM.bin_runtime(KApply(f'contract_{contract.name_with_path}'))
     if isinstance(method, Contract.Constructor):
         program = KEVM.init_bytecode(KApply(f'contract_{contract.name_with_path}'))
+        callvalue = method.callvalue_cell
+        method.callvalue_cell
```
PetarMax, line 563: What does this line do?

PetarMax (commented):
Left a number of comments to resolve/address.

src/tests/integration/test-data/show/ArithmeticCallTest.test_double_add(uint256,uint256).cse.expected
nwatson22: They are introduced by running the (implicit) constructor since `test_foundry_dependency_automated` now uses `run_constructor=True`. This contract inherits from `Test`, which inherits from `DSTest`, which has a member
```
    bool public IS_TEST = true;
```
and `StdChains`, which has a member
```
    bool private fallbackToDefaultRpcUrls = true;
```
I'm pretty sure the 1s are these getting initialized.

src/kontrol/prove.py
nwatson22: I'm not sure. `black` seems to be happy with either way, so I've manually changed it to the one line version.

src/kontrol/prove.py
nwatson22: Changed to the one line version.

src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
nwatson22: It's created by `abstract_cell_safely()`. I'm not sure what you mean by stable.

src/kontrol/prove.py
nwatson22: Must have been a typo. Removed.

src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol
nwatson22: Added `set()`, `add()`, and `count()` to the test.

src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol
nwatson22: Added.

src/tests/integration/test-data/show/ConstructorTest.test_contract_call().cse.expected
nwatson22: I've verified this by running with `--break-on-calls` and finding the edges where they are applied. The rules show up as `UNKNOWN`, which is something it would be nice to fix, but there are the number I would expect (5) and the calldata in the state prior to these edges (where `#execute` is at the top of the k cell) looks right.

src/tests/integration/test-data/show/ArithmeticCallTest.test_double_add(uint256,uint256).cse.expected
PetarMax: Excellent, thank you.

src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
PetarMax: I meant that it looks like a hash of some kind, and if it changes in future the expected outputs will also change.

src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
PetarMax: Something feels not right here, this and other CSE proofs of non-tests should be splitting at the top node.

src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol
PetarMax: After some more consideration, I think that:
- if a function (constructors included) is `payable`, then the call value should be symbolic (and properly constrained) when starting CSE execution of that function;
- otherwise, the call value should be set to `0`.

If a non-payable function is called with a non-zero call value, the summary won't apply and the function will revert very quickly. I don't think we need a summary of that particular reverting behaviour.

From what I understand, in this test, the constructor should work with a symbolic call value, and the other functions should work with `0`. Could we please add:
- the `payable` attribute to one of the contract functions;
- the constructor itself to `test_kontrol_cse`?

nwatson22: @PetarMax  How you described is already how it works, as far as I can tell. I've added payable to `add` and the constructor for `ImportedContract` and added the constructors to the `test_kontrol_cse`.

src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
nwatson22: That function is implemented as a hash of the term that's being abstracted, and it will rehash if there is still a name collision. I've replaced it with a system that generates one from either the variable name of the contract's ID if it's a variable or the actual ID.

PetarMax (approved):

PR #567, opened 2024-05-16
Improve support for storage fields

anvacaru:
Previously, the `Contract` had the storage layout elements encoded as tuples of `label` and `slot`.
This PR implements the `StorageField` class to additionally store the `type` and `offset`. Now `Contract.fields` becomes a `tuple[StorageField, ...]`.

In addition to this, I'm removing the `Contract.field_sentences` property, which was used to generate unused K code like
```k
    syntax Field ::= S2KsrcZModCounterField

    syntax S2KsrcZModCounterField ::= "number" [symbol(""), klabel(field_src%Counter_number)]

    rule  ( #loc ( S2KsrcZModCounter . number ) => 0 )
```

With this change, building the `integration/test-data/foundry` project improves from 
```
Time:
  Elapsed:     3:16.26 (mm:ss)
  CPU Usage:   757%

Disk Usage:
  Size of kompiled:  921M
```
to
```
Time:
  Elapsed:     2:43.37 (mm:ss)
  CPU Usage:   769%

Disk Usage:
  Size of kompiled:  862M
```


palinatolmach (approved):
lgtm! But it'd be safer if Petar and Tamás could also take a look.

src/kontrol/solc_to_k.py
```diff
@@ -329,6 +329,20 @@ def parse_devdoc(tag: str, devdoc: dict | None) -> dict:
...
+    label: str
+    data_type: str
+    slot: int
+    offset: int
+
+    def __init__(self, label: str, data_type: str, slot: int, offset: int) -> None:
+        self.label = label
+        self.data_type = data_type
+        self.slot = slot
+        self.offset = offset
```
tothtamas28, line 343:
```suggestion
class StorageField(NamedTuple):
    label: str
    data_type: str
    slot: int
    offset: int
```

tothtamas28 (commented):
Integration tests are failing, might be related to these changes.

src/kontrol/solc_to_k.py
```diff
@@ -329,6 +329,13 @@ def parse_devdoc(tag: str, devdoc: dict | None) -> dict:
     return natspecs
 
 
+class StorageField(NamedTuple):
+    label: str
+    data_type: str
+    slot: int
+    offset: int
```
PetarMax, line 336: How expressive is this? Can it capture, for example, the case in which two consecutive slots share elements of a single `struct`? It's ok if not, this can be the next PR, I suppose.

PetarMax: For example, there's a real-world case when
```
      {
        "astId": 1371,
        "contract": "contracts/C.sol:C",
        "label": "_x",
        "offset": 0,
        "slot": "1",
        "type": "t_struct(SomeType)3669_storage"
      },
      {
        "astId": 1374,
        "contract": "contracts/C.sol:C",
        "label": "_y",
        "offset": 0,
        "slot": "9",
        "type": "SomeOtherType"
      },
```
and that struct is spread throughout slots 1 to 8.

PetarMax: Having seen the code below, I think that it should be expressive enough, but that the code currently doesn't account for it.

PetarMax (commented):
LGTM as a first step, should we have some sort of test that checks that this field calculation is correct?

src/kontrol/solc_to_k.py
anvacaru: Yes, I can add some unit tests. I haven't looked at `structs` examples yet.

flake.lock
```diff
@@ -681,7 +681,7 @@
     "solc-macos-amd64-list-json": {
       "flake": false,
       "locked": {
-        "narHash": "sha256-akcOahJE/QR0jCPvo4aBf0loJt9P8zvP2E7oCBCRIho=",
+        "narHash": "sha256-f5kzXsfnxgad+1FxaGOlflIUVt8RPNBGS5wkqb63DVA=",
```
tothtamas28, line 684: What made this change necessary?

src/kontrol/solc_to_k.py
```diff
@@ -1295,3 +1274,24 @@ def _find_function_calls(node: dict) -> None:
...
+            type_info = types.get(field['type'], {})
+            storage_field = StorageField(
+                label=field['label'],
+                data_type=type_info.get('label', field['type']),
+                slot=int(field['slot']),
+                offset=int(field['offset']),
+            )
+            fields_list.append(storage_field)
+        except (KeyError, ValueError) as e:
+            _LOGGER.error(f'Error processing field {field}: {e}')
```
tothtamas28, line 1295: Is it OK for a field to be not parsed in case of failure? If not, an exception should be raised here.

src/tests/unit/test_solc_to_k.py
```diff
@@ -432,3 +440,59 @@ def test_find_function_calls(test_id: str, ast: dict, expected: list[str]) -> No
...
+    ),
+]
+
+
+@pytest.mark.parametrize(
+    'test_id,storage_layout,expected',
+    LAYOUT_DATA,
+    ids=[test_id for test_id, *_ in LAYOUT_DATA],
+)
+def test_storage_layout_fields(test_id: str, storage_layout: dict, expected: tuple[StorageField, ...]) -> None:
```
tothtamas28, line 494: Nice!

src/kontrol/solc_to_k.py
anvacaru: I'm not sure at this moment what would cause a failure like that. I can raise an exception here instead in the next PR.

flake.lock
anvacaru: In previous runs, and on other PRs ([example](https://github.com/runtimeverification/kontrol/actions/runs/9107939716/job/25037927689?pr=567#step:5:705)), that hash was not matching anymore. I tried running `nix --extra-experimental-features nix-command --extra-experimental-features flakes flake update` but without any resulting changes. This was my next attempt.

PR #571, opened 2024-05-17
Contract storage fields in CSE

anvacaru:
~Blocked on : https://github.com/runtimeverification/kontrol/pull/570 to bring in the changes from https://github.com/runtimeverification/evm-semantics/pull/2433 to clean some of the code first.~

First PR for https://github.com/runtimeverification/kontrol/issues/362:

- extending the `StorageField` class to store the `numberOfBytes` property
- implementing `_create_initial_storage` in prove.py that creates the KAst representation of the storage layout for a given set of storage fields.
This representation is used to populate the storage of `CONTRACT_ID` symbolic account in CSE only.
For now supporting just `uint*`, `address` and `bool`


src/tests/integration/test-data/show/AddConst.applyOp(uint256).cse.expected
```diff
@@ -1,64 +1,32 @@
 
-┌─ 1 (root, split, init)
```
PetarMax, line 2: I think this should still be a `split`.

anvacaru: Yup. This happened because I stored the fields as `#buf`s, which are of sort `Bytes`, and the `#lookup(Map,Int)` rule that reads the storage expects the value to be an `Int`.

palinatolmach (approved):

PetarMax:
There is a problem with `test_add_const` in CSE, it would appear the summaries are not getting used. Investigating now.

src/tests/integration/test-data/show/CSETest.test_add_const(uint256,uint256).cse.expected
```diff
@@ -21,12 +21,12 @@
 ┃  │   callDepth: 0
 ┃  │   statusCode: STATUSCODE:StatusCode
 ┃  │
-┃  │  (464 steps)
-┃  └─ 12 (vacuous, leaf)
-┃      k: #execute ~> #return 128 0 ~> #pc [ CALL ] ~> #execute ~> CONTINUATION:K
+┃  │  (979 steps)
```
PetarMax, line 24: This increase is due to a problem with the specification of `setConst`.

src/tests/integration/test-data/show/CSETest.test_add_const(uint256,uint256).cse.expected
```diff
@@ -36,8 +36,8 @@
    │   callDepth: 0
    │   statusCode: STATUSCODE:StatusCode
    │
-   │  (1123 steps)
-   ├─ 13 (terminal)
+   │  (1417 steps)
```
PetarMax, line 39: This increase is due to a problem with the specification of `setConst`.

PetarMax:
We have a problem. By setting the storage in this way, we cannot use the generated specifications when a slot has not been explicitly initialised---this, for example, is the case for the `AddConst.setConst` function, where slot `0` holds an `uint256` and whose specification has
```
<storage>
  ( 0 |-> ( SF_C:Int => VV0_x_114b9705:Int ) )
</storage>
```
but when the `CSETest.test_add_const` initialises the `AddConst` contract, the storage is set to
```
<storage>
  .Map
</storage>
```
implicitly having all slots set to zero and using `#lookup` to leverage this implicitness. As a result, the specification of `setConst` is not used in `test_add_const`, and this is visible in the increase of steps in the expected outpute [here](https://github.com/runtimeverification/kontrol/pull/571/files#r1613308523) and [here](https://github.com/runtimeverification/kontrol/pull/571/files#r1613308714).

I have two solutions in mind, first more favourable than the second, I think:
1. We do not create an explicit storage layout, but instead have a fully symbolic storage and work in terms of `#lookup` and `&Int / >>Int` masks and shifts, just like the compiler does. For the `setConst` contract, this would mean just introducing the constraints `0 <=Int #lookup(STORAGE, 0) andBool #lookup(STORAGE, 0) <Int pow256`, which actually hold by construction, and the specification would have `<storage> STORAGE => STORAGE [ 0 <-| VV0_x_114b9705:Int ] </storage>`. I see the following pros and cons
    - Pros: resilient to presence or absence of explicit fields in storage, can handle all storage constructs;
    - Cons: more complex expressions in the path constraint, hindering readability and potentially slowing down execution.
2. We create an explicit storage layout as much as we can for tests as well. I am not sure how this would be done other than intercepting constructor calls, and it would not be feasible for mappings, which would have to remain implicit in the style of solution 1.

@anvacaru, @ehildenb, @palinatolmach, let's discuss this a bit, please. Other ideas welcome.

anvacaru:
Closing this since we won't merge it.

src/kontrol/solc_to_k.py
```diff
@@ -1289,9 +1291,25 @@ def process_storage_layout(storage_layout: dict) -> tuple[StorageField, ...]:
...
+    if data_type.startswith('uint'):
+        return mlEqualsTrue(KEVM.range_uint(int(data_type[4:]), symbolic_var))
+
+    match data_type:
+        case 'address':
+            return mlEqualsTrue(KEVM.range_address(symbolic_var))
+        case 'bool':
+            return mlEqualsTrue(KEVM.range_bool(symbolic_var))
+        case _:
+            raise ValueError(f'Unimplemented: {data_type}')
```
PetarMax, line 1315: Note: This will be useful for future.

PetarMax, line 1315: Note: this will be useful in future.

Issue #562, opened 2024-05-14
Support remaining assert methods


anvacaru:
After https://github.com/runtimeverification/kontrol/pull/558, here is a list with the remaining assertions that should be implemented:
1. dynamic types string bytes uint256[] int256[] bool[] bytes32[] address[] for assertEq and assertNotEq.
2. assertApproxEqAbs and assertApproxEqRel: take an additional maxDelta argument.
3. assert*Decimals : take an additional decimals value used to format the values in error messages.

[Vm.sol link here](https://github.com/foundry-rs/forge-std/blob/master/src/Vm.sol#L938)

Issue #559, opened 2024-05-14
Issues running signed integers


anvacaru:
Given the two test examples below
```solidity
    function test_assertLt_assertLe(int256 a, int256 b) public pure {
        vm.assume(a < b);
        assertLt(a, b);
        assertive(a, b);
    }
    function test_assertLt_assertLe(uint256 a, uint256 b) public pure {
        vm.assume(a < b);
        assertLt(a, b);
        assertLe(a, b);
    }
```
the assumption for the test that uses `uint256` values is simplified to 
```
#And ( { true #Equals VV0_a_114b9705:Int <Int VV1_b_114b9705:Int }
```

whereas for the one that uses `int256` it is
```
#And { true #Equals b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01" ~> .K ==K #range ( #padToWidth ( 32 , #asByteStack (
 ( (chop ( VV0_a_114b9705:Int )) s<Word (chop ( VV1_b_114b9705:Int )) ) ) ) , 0 , minInt ( 32 , lengthBytes ( #padToWidth ( 32 , #asByteStack ( ( (chop ( VV0_a_114b9705:Int )) s<Word (chop ( VV1_b_114b970
5:Int )) ) ) ) ) ) ) +Bytes #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\
x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\
[...]

PR #584, opened 2024-05-29
`assertApproxEqAbs` cheatcode (partial) support

jinxinglim:
Partially addresses https://github.com/runtimeverification/kontrol/issues/562.

### What's new

This PR attempts to introduce support of `assertApproxEqAbs` cheatcode in Kontrol. The PR is broken down into smaller commits for easy review.

1. 91ce6248dd5653ed07cea82a6cae07396bc2b82a - Introduce `assertApproxEqAbs` rules in `assert.md`
2. aefaa8ecdf9150962ee914131d16db285712448c - Added `assertApproxEqAbs` related tests

Just for context, `assertApproxEqAbs(uint/int a, uint/int b, uint maxDelta)` checks if the difference between `a` and `b` is within `maxDelta`.

### Comments

1. There are 4 tests added in `Unit.t.sol`. Two of them, `test_assertApproxEqAbs_uint` and `test_assertApproxEqAbs_int_same_sign` passed but not the other two, i.e., `test_assertApproxEqAbs_int_opp_sign` and `test_assertApproxEqAbs_int_zero_cases`. The two tests that failed involve calling `assertApproxEqAbs` for `int256`. The following are the supporting documents for the proofs ran:

    [Proof results for the 4 new tests](https://github.com/runtimeverification/kontrol/files/15477921/assertApproxEqAbs_tests.txt)
    [kontrol show test_assertApproxEqAbs_int_opp_sign](https://github.com/runtimeverification/kontrol/files/15477923/kontrol_show_test_assertApproxEqAbs_int_opp_sign.txt)
    [kontrol show test_assertApproxEqAbs_int_zero_cases](https://github.com/runtimeverification/kontrol/files/15477926/kontrol_show_test_assertApproxEqAbs_int_zero_cases.txt)

2. @palinatolmach and @anvacaru As this is my first PR for Kontrol, do let me know how I can go about addressing point 1 above and/or if I can improve on so that I  can make better PR in the future. Thank you.


src/kontrol/kdist/assert.md
```diff
@@ -33,19 +33,22 @@ We define macros for different assert methods.
...
-                   | "#assert_lt" Int Int Bytes     [label(assert_lt),     macro]
-                   | "#assert_not_eq" Int Int Bytes [label(assert_not_eq), macro]
+    syntax KItem ::= "#assert_eq" Int Int Bytes                [label(assert_eq),     macro]
+                   | "#assert_ge" Int Int Bytes                [label(assert_ge),     macro]
+                   | "#assert_gt" Int Int Bytes                [label(assert_gt),     macro]
+                   | "#assert_le" Int Int Bytes                [label(assert_le),     macro]
+                   | "#assert_lt" Int Int Bytes                [label(assert_lt),     macro]
+                   | "#assert_not_eq" Int Int Bytes            [label(assert_not_eq), macro]
+                   | "#assert_approx_eq_abs" Int Int Int Bytes [label(assert_approx_eq_abs), macro]
  // -----------------------------------------------------------------------------
```
anvacaru, line 43: Formatting.
```suggestion
                   | "#assert_approx_eq_abs" Int Int Int Bytes [label(assert_approx_eq_abs), macro]
 // -----------------------------------------------------------------------------------------------
```

anvacaru (approved):
Approved with a small comment! 

@jinxinglim I've seen a similar failure for `int256` when implementing the other assert methods. It looks like the prover cannot simplify further a `chop (X:Int)` if `X` is a `int256`.
I've opened an issue about this here https://github.com/runtimeverification/kontrol/issues/559. 

For now I'd say to add the tests that use the `int256` to the corresponding skip file.

jinxinglim:
> Approved with a small comment!
> 
> @jinxinglim I've seen a similar failure for `int256` when implementing the other assert methods. It looks like the prover cannot simplify further a `chop (X:Int)` if `X` is a `int256`. I've opened an issue about this here #559.
> 
> For now I'd say to add the tests that use the `int256` to the corresponding skip file.

Okay, thank you @anvacaru for the review!

palinatolmach (approved):
lgtm too, thanks @jinxinglim and @anvacaru!

palinatolmach:
Oh one moment, sorry: maybe in `test_assertApproxEqAbs_int_zero_cases ` we're missing an explicit conversion of the result to `int256`? I just tried that in Solidity shell, and here's what I get for adding two `int256`s one of which is zero:

```
❯ chisel
Welcome to Chisel! Type `!help` to show available commands.
➜ int256 x = int256(14)
➜ x = -x
Type: int256
├ Hex: 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff2
├ Hex (full word): 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff2
└ Decimal: -14
➜ x
Type: int256
├ Hex: 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff2
├ Hex (full word): 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff2
└ Decimal: -14
➜ int256 y = int256(0)
➜ y
Type: int256
├ Hex: 0x0
├ Hex (full word): 0x0
└ Decimal: 0
➜ y + x
Type: uint256
├ Hex: 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff2
├ Hex (full word): 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff2
└ Decimal: 115792089237316195423570985008687907853269984665640564039457584007913129639922
```
The result of `y + x` would be greater than `maxDelta >= x`, and that might be what we're seeing in `{ false #Equals chop ( 0 -Int VV0_a_114b9705:Int ) <=Int VV1_maxDelta_114b9705:Int }`?

palinatolmach:
Ok, so all these four tests pass with Foundry. But [here's](https://github.com/foundry-rs/forge-std/blob/master/src/StdMath.sol) how they calculate delta for signed ints:
```solidity
  function delta(uint256 a, uint256 b) internal pure returns (uint256) {
      return a > b ? a - b : b - a;
  }

  function delta(int256 a, int256 b) internal pure returns (uint256) {
      // a and b are of the same sign
      // this works thanks to two's complement, the left-most bit is the sign bit
      if ((a ^ b) > -1) {
          return delta(abs(a), abs(b));
      }

      // a and b are of opposite signs
      return abs(a) + abs(b);
  }

  function abs(int256 a) internal pure returns (uint256) {
      // Required or it will fail when `a = type(int256).min`
      if (a == INT256_MIN) {
          return 57896044618658097711785492504343953926634992332820282019728792003956564819968;
      }

      return uint256(a > 0 ? a : -a);
  }
```
Maybe we should do something similar? Not necessarily in this PR.

jinxinglim:
Thank you @palinatolmach for the review.

> I don't think we change the sign bit in symbolic variables when casting them to `int`, so that'd probably has to be fixed first?

Yes, I think so. I tested with the following simple unit test and it passed on Foundry but failed in Kontrol:

```solidity
    function test_unit() public pure {
        int256 a = -14;
        int256 zero = 0;
        uint256 delta = 15;
        assertApproxEqAbs(a, zero, delta);
    }
```

> Maybe we should do something similar? Not necessarily in this PR.

I have tried several ways for the rewrite rules for `#assert_approx_eq_abs`, e.g., using `absInt(W1) + absInt(W2)` when `W1` and `W2` are opposite signs instead of using `maxInt(W1, W2) - minInt(W1, W2)` but it still do not work.

palinatolmach:
Thanks @jinxinglim! Let's merge this PR as is, as Andrei also suggested, and get this addressed if there's a need to signed `int` support. I've labeled it as `automerge`, this way it'll get merged once the CI is passing.

Issue #586, opened 2024-05-29
Add a `kontrol.toml` to the files that get added using `kontrol init`.


anvacaru:
To do for a follow up PR: we should add a `kontrol.toml` to the files that get added in `kontrol init`.

_Originally posted by @anvacaru in https://github.com/runtimeverification/kontrol/issues/580#issuecomment-2136505538_

Issue #442, opened 2024-03-13
Lemmas imported in one module are applied in others


palinatolmach:
For further investigation. 

While working on https://github.com/runtimeverification/kontrol/pull/369, @PetarMax and I noticed that Optimism-specific lemmas that are imported in the `PortalTest` contract via
```
kontrol build --require lemmas.k --module-import PortalTest:PAUSABILITY-LEMMAS
```
are only being imported into `PortalTest-VERIFICATION` module:
```
module S2KtestZModPortalTest-VERIFICATION
    imports public S2KtestZModPortalTest-CONTRACT
    imports public PAUSABILITY-LEMMAS
endmodule
```
but it looks like they’re being used across all tests in Kontrol test suite. For example, `AssumeTest.test_assume_false(uint256,uint256)` is affected, with
```
<k>
   ( #addr [ ISZERO ]
   ~> #exec [ ISZERO ]
   ~> #pc [ ISZERO ] => CALL 0 645326474426547203313410069153905908525362434349 0 388 100 388 0
   ~> #pc [ CALL ] )
   ~> #execute
   ~> _CONTINUATION
</k>
```
getting rewritten to
```
<k>
   ( .K => CALL 0 645326474426547203313410069153905908525362434349 0 388 100 388 0
   ~> #pc [ CALL ] )
   ~> #execute
   ~> _CONTINUATION
</k>
```

PR #587, opened 2024-05-30
Insert kontrol.toml with kontrol init

anvacaru:
Closes: #586


src/kontrol/utils.py
```diff
@@ -112,3 +104,40 @@ def kontrol_file_contents() -> str:
...
+
+def kontrol_toml_file_contents() -> str:
+    return """[build.default]
+foundry-project-root       = '.'
+regen                      = true
+rekompile                  = true
+verbose                    = true
+debug                      = false
+require                    = 'lemmas.k'
+module-import              = 'TestBase:KONTROL-LEMMAS'
```
palinatolmach, line 117: That'll fail on kompilation if `TestBase` is not present, right? Maybe we can use `CounterTest` that get included as part of the Foundry template?

src/kontrol/utils.py
```diff
@@ -112,3 +104,40 @@ def kontrol_file_contents() -> str:
...
+
+[prove.default]
+foundry-project-root       = '.'
+verbose                    = true
+debug                      = false
+max-depth                  = 25000
+reinit                     = false
+cse                        = false
+workers                    = 4
+failure-information        = false
```
palinatolmach, line 127: I think we might want `failure-information` to be true by default, as it prints the failing node with the path condition? Also, IIRC, counterexample might not get printed if `failure-information` is off even with `counterexample-information` being true (but not 100% sure).

src/kontrol/utils.py
```diff
@@ -112,3 +104,40 @@ def kontrol_file_contents() -> str:
...
+minimize-proofs            = false
+fail-fast                  = true
+smt-timeout                = 1000
+break-every-step           = false
+break-on-jumpi             = false
+break-on-calls             = false
+break-on-storage           = false
+break-on-basic-blocks      = false
+break-on-cheatcodes        = false
+
```
palinatolmach, line 138: I understand it's not an an exhaustive list, but maybe we can add
```
run-constructor = false
```
too? Maybe that'll also serve as a reminder that they're not getting executed by default.

palinatolmach (approved):
Looks good, thanks! Just left a few suggestions on the options to be included.

src/kontrol/utils.py
anvacaru: `CounterTest` is present only if you run `kontrol init`. If you run `kontrol init --skip-forge` which installs the kontrol dependencies on an already existing foundry project, then it might be missing.
`TestBase` instead is inherited by the forge-std `Test` contract, which all test contracts should inherit.
https://github.com/foundry-rs/forge-std/blob/52715a217dc51d0de15877878ab8213f6cbbbab5/src/Test.sol#L30

Additionally, we could try something like `KontrolCheats`, which is the contract of the Kontrol cheatcodes and is always added with `kontrol init`.

src/kontrol/utils.py
anvacaru: Thanks! I'll add these!

src/kontrol/utils.py
palinatolmach: Ah right, fair enough, thanks! I forgot `TestBase` is a real contract, I think it's a good choice then. Frankly speaking, because of https://github.com/runtimeverification/kontrol/issues/442, these lemmas are probably applied to all other modules too :).

src/kontrol/utils.py
```diff
@@ -112,3 +104,40 @@ def kontrol_file_contents() -> str:
...
+debug                      = false
+require                    = 'lemmas.k'
+module-import              = 'TestBase:KONTROL-LEMMAS'
+
+[prove.default]
+foundry-project-root       = '.'
+verbose                    = true
+debug                      = false
+max-depth                  = 25000
+reinit                     = false
```
PetarMax, line 124: Just to say that we have a little bit of an issue with `--reinit`, in that (as far as I know) `setUp` is not re-run unless we say `--reinit` explicitly even if the code has changed. I don't think this affects its default being `false`.

src/kontrol/utils.py
PetarMax: We also don't have `--max-iterations`, what does it default to internally?

PR #593, opened 2024-05-31
Support of `assertApproxEqRel` cheatcode

jinxinglim:
### What's new

This PR attempts to support the `assertApproxEqRel` cheatcode in Kontrol (see Appendix section below for the implementation of this cheatcode).

I added 4 unit tests in `src/tests/integration/test-data/src/Unit.t.sol`:
1. `test_assertApproxEqRel_uint_unit()` - passed in <1min, and hence, under `src/tests/integration/test-data/end-to-end-prove-all`
2. `test_assertApproxEqRel_int_same_sign_unit()` - passed in <1min, and hence, under `src/tests/integration/test-data/end-to-end-prove-all`
3. `test_assertApproxEqRel_int_zero_cases_unit()` - passed in <1min, and hence, under `src/tests/integration/test-data/end-to-end-prove-all`
4. `test_assertApproxEqRel_int_opp_sign_unit()` - failed in <1min, and hence, under `src/tests/integration/test-data/end-to-end-prove-skip`

Here is the proofs' output: [test_units.txt](https://github.com/user-attachments/files/15508048/test_units.txt)

### Appendix

Here is the older version of the `assertApproxEqRel(uint256, uint256, uint256 maxPercentDelta)` cheatcode (but I believe the same implementation still holds in the current version with onyl the difference in log outputs) and the relevant functions required:

```solidity
...
    int256 private constant INT256_MIN = -57896044618658097711785492504343953926634992332820282019728792003956564819968;

    function abs(int256 a) internal pure returns (uint256) {
        // Required or it will fail when `a = type(int256).min`
        if (a == INT256_MIN) {
            return 57896044618658097711785492504343953926634992332820282019728792003956564819968;
        }

        return uint256(a > 0 ? a : -a);
    }

    function delta(uint256 a, uint256 b) internal pure returns (uint256) {
        return a > b ? a - b : b - a;
    }

    function delta(int256 a, int256 b) internal pure returns (uint256) {
        // a and b are of the same sign
        // this works thanks to two's complement, the left-most bit is the sign bit
        if ((a ^ b) > -1) {
            return delta(abs(a), abs(b));
        }

        // a and b are of opposite signs
        return abs(a) + abs(b);
    }

    function percentDelta(uint256 a, uint256 b) internal pure returns (uint256) {
        uint256 absDelta = delta(a, b);

        return absDelta * 1e18 / b;
    }

    function percentDelta(int256 a, int256 b) internal pure returns (uint256) {
        uint256 absDelta = delta(a, b);
        uint256 absB = abs(b);

        return absDelta * 1e18 / absB;
    }
...
    function percentDelta(uint256 a, uint256 b) internal pure returns (uint256) {
        uint256 absDelta = delta(a, b);

        return absDelta * 1e18 / b;
    }

...

    // Same implementation for `assertApproxEqRel(int256 a, int256 b, uint256 maxPercentDelta)`
    function assertApproxEqRel(
        uint256 a,
        uint256 b,
        uint256 maxPercentDelta // An 18 decimal fixed point number, where 1e18 == 100%
    ) internal virtual {
        if (b == 0) return assertEq(a, b); // If the expected is 0, actual must be too.

        uint256 percentDelta = stdMath.percentDelta(a, b);

        if (percentDelta > maxPercentDelta) {
            emit log("Error: a ~= b not satisfied [uint]");
            emit log_named_uint("    Expected", b);
            emit log_named_uint("      Actual", a);
            emit log_named_decimal_uint(" Max % Delta", maxPercentDelta * 100, 18);
            emit log_named_decimal_uint("     % Delta", percentDelta * 100, 18);
            fail();
        }
    }
```

### Question/Help needed

I tried running the generalized version of the 4 tests but it took too long to run (some took >30min). The following is two examples where I generalized `test_assertApproxEqRel_uint_unit()`:
```solidity
    function test_assertApproxEqRel_uint_v1(uint256 a, uint256 b, uint256 maxPercentDelta) public pure 
    {
        if (b == 0) {
            a = 0;
        }
        else {
            uint256 absDelta = max(a, b) - min(a, b);
            vm.assume(absDelta <= type(uint256).max / 1e18); // prevent overflow of absDelta * 1e18
            uint256 percentDelta = absDelta * 1e18 / b;
            vm.assume(maxPercentDelta >= percentDelta);
        }
        string memory err = "throw test";
        assertApproxEqRel(a, b, maxPercentDelta);
        assertApproxEqRel(a, b, maxPercentDelta, err); 
    }

    function test_assertApproxEqRel_uint_v2(uint256 a, uint256 b, uint256 maxPercentDelta) public pure 
    {
        vm.assume(b > 0);
        uint256 absDelta = max(a, b) - min(a, b);
        vm.assume(absDelta <= type(uint256).max / 1e18); // prevent overflow of absDelta * 1e18
        uint256 percentDelta = absDelta * 1e18 / b;
        vm.assume(maxPercentDelta >= percentDelta);
        string memory err = "throw test";
        assertApproxEqRel(a, b, maxPercentDelta);
        assertApproxEqRel(a, b, maxPercentDelta, err); 
    }
```

`test_assertApproxEqRel_uint_v1` took >10min to complete. here is his `kontrol show`: [kshow_test_assertApproxEqRel_uint_vA1.txt](https://github.com/user-attachments/files/15508128/kshow_test_assertApproxEqRel_uint_vA1.txt).

On the other hand, `test_assertApproxEqRel_uint_v2` took >30min and yet to complete, which I have to shut it down.

@anvacaru and @palinatolmach I suspect its the additional assumptions added in these tests. Do you think it is an issue with the tests themselves or its the rewrite rules for `assertApproxEqRel` that I have introduced? I have looked at the `kcfg`,the branching looks pretty normal so not too sure where the proofs have taken so long to execute. One thing I noticed is that it stuck at `[Info#proxy] Simplifying execution state` for very long at some point during the proofs.


src/kontrol/kdist/assert.md
```diff
@@ -49,6 +50,14 @@ These macros expand into the `#assert` production defined earlier, providing an
     rule #assert_not_eq W1 W2 ERR => #assert (W1 =/=Int W2) ERR +Bytes String2Bytes(": " +String Int2String(W1) +String " == " +String Int2String(W2))
     rule #assert_approx_eq_abs W1 W2 W3 ERR => #assert ((maxInt(W1, W2) -Int minInt(W1, W2)) <=Int W3) ERR 
       +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) +String " (max delta: " +String Int2String(W3) +String ", real delta: " +String Int2String(maxInt(W1, W2) -Int minInt(W1, W2)) +String ")")
+    rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert (W1 ==Int W2) ERR 
+      +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
+      +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: undefined)") // displayed percentage in Kontrol does not include any decimal values
+      requires W2 ==Int 0
+    rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert ((((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) <=Int W3) ERR
```
palinatolmach, line 57: The implementation looks accurate to me, but I believe that it's just the multiplication and division that makes this particularly slow and difficult for the SMT solver. We've encountered that within the engagement, but I'm not sure if we've got any solutions for it now; let's check in on Monday.

I think, we're also doing this evaluation twice — once for the assert, and once for the resulting error message?

src/tests/integration/test-data/src/Unit.t.sol
```diff
@@ -242,6 +242,59 @@ contract UnitTest is Test {
         assertApproxEqAbs(int_zero, neg_a, maxDelta, err);
     }
 
+    function test_assertApproxEqRel_uint_unit() public pure {
+        uint256 zero = 0;
+        uint256 a = 8;
+        uint256 b = 10;
+        uint256 percentDelta = 1e18;
+        string memory err = "throw test";
```
palinatolmach, line 250: Maybe it'd be useful to add tests checking that the error `err` is returned correctly, e.g., similar to https://github.com/runtimeverification/kontrol/blob/a551f742db727cd23c989c4c761b37a0e0821a7f/src/tests/integration/test-data/src/Unit.t.sol#L60-L63

palinatolmach:
Here's the current implementation for the `uint` version of the cheatcode: https://github.com/foundry-rs/foundry/blob/5494c33bc7977b3537bd296e375431d938d44ca3/crates/cheatcodes/src/test/assert.rs#L1120, it does look very similar to the older Solidity version to me too.

src/kontrol/kdist/assert.md
jinxinglim: Okay. Perhaps I will remove the resulting error message assertions from all the tests and put them in separate tests.

src/tests/integration/test-data/src/Unit.t.sol
jinxinglim: Yes, I agree. Will remove them from the tests and put them in separate tests.

jinxinglim: Hi @palinatolmach. I have updated the tests in https://github.com/runtimeverification/kontrol/pull/593/commits/01dda9d25ab4b0a9247e0e7c18d895bd9651d099. If the changes are okay, please feel free to resolve this particular issue.

palinatolmach: Thanks @jinxinglim! Do some of these new tests pass as expected, e.g., `test_assertApproxEqAbs_uint_err`? Right now, they'll all be skipped as they're all placed in the `...-skip` file.

jinxinglim: I ran `forge test` on them locally and they passed. But have not tried them with `kontrol prove`. There are `test_*_err` tests that are placed in both files. So I am unsure what's the criteria to place tests in those files. Are the tests in `*-all` supposed to pass `forge test` or `kontrol prove`?

palinatolmach: All tests in `*-all` are supposed to pass with `kontrol prove`, these are tests that are executed with `kontrol prove` on CI to make sure all new changes are implemented correctly and don't break any of the existing functionality. For features that are available in both Foundry and Kontrol, they should pass with both commands :). But, generally, all tests are concerned with Kontrol. 

You can add tests to `*-skip` if they tests functionality that, e.g., has not been implemented yet, so we can't include those in CI runs. But all tests that exercise new functionality that works are to only be placed in `*-all`, this way they'll be run by CI on both booster and legacy backends. Ideally, they should take not more than several minutes to run, and, if they're time consuming, please add them to `*-skip-legacy` so they're only run on the booster backend, and the CI run time doesn't increase significantly. There's also come explanation here: https://github.com/runtimeverification/kontrol/blob/master/CONTRIBUTING.md. Hope that clarifies!

jinxinglim: Okay. Got it. Thank you! Let me test the new `test_*_err` tests with `kontrol prove` and place them accordingly.

jinxinglim: Hi @palinatolmach. I just ran a few of the new `test_*_err` tests and all of them returned the same error as the one seen in issue https://github.com/runtimeverification/kontrol/issues/581, which is yet to resolve. The new `test_*_err` tests only passed at `forge test` but not `kontrol prove`. So I will place them in `end-to-end-skip`. Is this okay?

palinatolmach: I don't think you've missed anything! Looks like it's an SMT error caused by CI flakiness, I triggered a re-run, so let's see. I see about the https://github.com/runtimeverification/kontrol/issues/581 failure, thanks! I agree with adding tests to `end-to-end-skip`, but it'd be good to figure out why that happens, let's check with the backend tomorrow.

src/tests/integration/test-data/src/Unit.t.sol
```diff
@@ -181,9 +181,17 @@ contract UnitTest is Test {
...
-        string memory err = "throw test";
         assertApproxEqAbs(a, b, maxDelta);
-        assertApproxEqAbs(a, b, maxDelta, err); 
+    }
+
+    function test_assertApproxEqAbs_uint_err() public {
+        string memory err = "throw test";
+        string memory err_lt = "throw test: 11 !~= 121 (max delta: 100, real delta: 110)";
+        string memory err_gt = "throw test: 121 !~= 11 (max delta: 100, real delta: 110)";
+        vm.expectRevert(bytes(err_lt));
```
palinatolmach, line 191: That's not directly related to this PR, but I'm not sure how well Kontrol handles tests with several `expectRevert`s, might be worth checking.

src/kontrol/kdist/assert.md
```diff
@@ -49,6 +50,14 @@ These macros expand into the `#assert` production defined earlier, providing an
...
       +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) +String " (max delta: " +String Int2String(W3) +String ", real delta: " +String Int2String(maxInt(W1, W2) -Int minInt(W1, W2)) +String ")")
+    rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert (W1 ==Int W2) ERR 
+      +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
+      +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: undefined)") // displayed percentage in Kontrol does not include any decimal values
+      requires W2 ==Int 0
+    rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert ((((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) <=Int W3) ERR 
+      +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
+      +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: " +String Int2String((((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) divInt (10 ^Int 16)) +String "%)") // displayed percentage in Kontrol does not include any decimal values
```
palinatolmach, line 59: Do you think there might be a way to compute and cache the delta once, to avoid computing it twice — once when checking if the assert holds, and another — to generate an error string?

src/tests/integration/test-data/src/Unit.t.sol
jinxinglim: Okay. Will check on this.

src/kontrol/kdist/assert.md
jinxinglim: Oh. This is a good point! Let me think about how to do it. If you have examples that somehow caches values, that would be good for me to learn from it.

palinatolmach: I think it should be possible to introduce a variable to use within the rule, but I'm not sure how — maybe you can ask in K or Kontrol channel? ty!

anvacaru: You could try using the `#as` pattern ([link here](https://github.com/runtimeverification/k/blob/9100689e357a7e5820362869a73684468ef9c91b/docs/user_manual.md?plain=1#L2086))

anvacaru: Alternatively, you can rewrite the `#assert_approx_eq_rel W1 W2 W3 ERR` rule into an aux rule that has the delta already computed.
Something like
```k
<k> #assert_approx_eq_rel W1 W2 W3 ERR => #assert_approx_eq_rel_aux W1 W2 W3 divInt (10 ^Int 16) ... </k>
```

Or you could compute the max-delta in the `#cheatcode_call` rule like
```k
    rule [cheatcode.call.assertApproxEqRel]:
         <k> #cheatcode_call SELECTOR ARGS => #assert_approx_eq_rel _ _ (#asWord(#range(ARGS, 64, 32)) divInt (10 ^Int 16) ... </k>
      requires ...

```

jinxinglim: Thanks for all the suggestions. I will take all these into considerations and update the PR.

src/tests/integration/test-data/src/Unit.t.sol
jinxinglim: I have limited all `test_*_err` tests to have 1 `expectRevert` in commit https://github.com/runtimeverification/kontrol/pull/593/commits/b5064de4b17f7383531ffe138ffe964e6e1842ed.

palinatolmach: Thanks! Does it still fail with the same issue? Do you think we support several subsequent `expectRevert`s in general?

src/kontrol/kdist/assert.md
jinxinglim: Hi @palinatolmach and @anvacaru. After experimenting with different methods for the `#assert_approx_eq_rel`, here are my findings:

I ran the following test:
```solidity
    function test_assertApproxEqRel_uint_vA1(uint256 a, uint256 b, uint256 maxPercentDelta) public pure 
    {
        if (b == 0) {
            a = 0;
        }
        else {
            uint256 absDelta = max(a, b) - min(a, b);
            vm.assume(absDelta <= type(uint256).max / 1e18); // prevent overflow of absDelta * 1e18
            uint256 percentDelta = absDelta * 1e18 / b;
            vm.assume(maxPercentDelta >= percentDelta);
        }
        assertApproxEqRel(a, b, maxPercentDelta);
    }
```

with the following 3 versions of 

1. Current PR version:
    ```k
        rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert (W1 ==Int W2) ERR 
          +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
          +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: undefined)") // displayed percentage in Kontrol does not include any decimal values
          requires W2 ==Int 0
        rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert ((((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) <=Int W3) ERR 
          +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
          +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: " +String Int2String((((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) divInt (10 ^Int 16)) +String "%)") // displayed percentage in Kontrol does not include any decimal values
          requires W2 =/=Int 0   
    ```

2. Using `#assert_approx_eq_rel_aux`:
    ```k
        rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert (W1 ==Int W2) ERR 
          +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
          +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: undefined)") // displayed percentage in Kontrol does not include any decimal values
          requires W2 ==Int 0
        rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert_approx_eq_rel_aux W1 W2 (((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) W3 ERR 
          requires W2 =/=Int 0   
        rule #assert_approx_eq_rel_aux W1 W2 W W3 ERR => #assert (W <=Int W3) ERR 
          +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
          +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: " +String Int2String(W divInt (10 ^Int 16)) +String "%)") // displayed percentage in Kontrol does not include any decimal values
    ```

3. Using `#let ...#in`:
    ```k
        rule #assert_approx_eq_rel W1 W2 W3 ERR => 
          #let W3_D = W3 divInt (10 ^Int 16) #in
          #assert (W1 ==Int W2) ERR 
          +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
          +String " (max delta: " +String Int2String(W3_D) +String "% , real delta: undefined)") // displayed percentage in Kontrol does not include any decimal values
          requires W2 ==Int 0
        rule #assert_approx_eq_rel W1 W2 W3 ERR => 
          #let W = ((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2) #in
          #let W3_D = W3 divInt (10 ^Int 16) #in
          #let W_D = W divInt (10 ^Int 16) #in
          #assert (W <=Int W3) ERR 
          +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2) 
          +String " (max delta: " +String Int2String(W3_D) +String "% , real delta: " +String Int2String(W_D) +String "%)") // displayed percentage in Kontrol does not include any decimal values
          requires W2 =/=Int 0   
    ```

It seems like the current version is the fastest (average of 4m45s) surprisingly, while using `#assert_approx_eq_rel_aux` and `#let ... #in` took around on average 5m15s and 5m30s respectively. I believe the other 2 might be fastest if we can test for the log string output if the assertion fails.

Should I stick to the current version or change to any of the other 2 versions (as the other 2 versions seem more presentable)?

palinatolmach (approved):

src/kontrol/kdist/assert.md
anvacaru: @jinxinglim 
One more comment, I would suggest formatting the rules like this to increase readability. In addition, the comments can go outside the ```k block.
```
    rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert (W1 ==Int W2) ERR 
                                        +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2)
                                                                 +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16))
                                                                 +String "% , real delta: undefined)")
      requires W2 ==Int 0

    rule #assert_approx_eq_rel W1 W2 W3 ERR => #assert ((((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) <=Int W3) ERR
                                        +Bytes String2Bytes(": " +String Int2String(W1) +String " !~= " +String Int2String(W2)
                                                                 +String " (max delta: " +String Int2String(W3 divInt (10 ^Int 16)) +String "% , real delta: " 
                                                                 +String Int2String((((maxInt(W1, W2) -Int minInt(W1, W2)) *Int (10 ^Int 18)) divInt absInt(W2)) divInt (10 ^Int 16)) +String "%)")
      requires W2 =/=Int 0
```

jinxinglim: Okay. Formatted in this commit https://github.com/runtimeverification/kontrol/pull/593/commits/34ebaa28f2c313f9315b6eb56eedaa4d198f21e5. Thank you!

PR #599, opened 2024-06-04
CSE: refactoring  `config_type` for top-level proofs

palinatolmach:
Follow up to https://github.com/runtimeverification/kontrol/pull/563/files.

### DIFF
--- a/src/kontrol/prove.py
+++ b/src/kontrol/prove.py
@@ -124,7 +124,6 @@ def foundry_prove(
             options=options,
             summary_ids=(summary_ids if include_summaries else []),
             deployment_state_entries=deployment_state_entries,
-            config_type=options.config_type,
         )
 
     if options.run_constructor:
@@ -272,7 +271,6 @@ def _run_cfg_group(
     foundry: Foundry,
     options: ProveOptions,
     summary_ids: Iterable[str],
-    config_type: ConfigType,
     deployment_state_entries: Iterable[DeploymentStateEntry] | None = None,
 ) -> list[APRProof]:
     def init_and_run_proof(test: FoundryTest) -> APRFailureInfo | Exception | None:
@@ -337,6 +335,10 @@ def _run_cfg_group(
                 )
 
             if proof is None:
+                # With CSE, top-level proof should be a summary if it's not a test function
+                if options.cse and options.config_type == ConfigType.TEST_CONFIG and not test.method.is_test:
+                    options.config_type = ConfigType.SUMMARY_CONFIG
+
                 proof = method_to_apr_proof(
                     test=test,
                     foundry=foundry,
@@ -348,7 +350,7 @@ def _run_cfg_group(
                     summary_ids=summary_ids,
                     active_symbolik=options.with_non_general_state,
                     hevm=options.hevm,
-                    config_type=config_type,
+                    config_type=options.config_type,
                     trace_options=TraceOptions(
                         {
                             'active_tracing': options.active_tracing,


src/kontrol/prove.py
```diff
@@ -81,7 +81,10 @@ def foundry_prove(
     )
 
     if options.cse:
-        test_suite = collect_tests(foundry, options.tests, reinit=options.reinit, return_empty=True)
+        exact_match = options.config_type == ConfigType.SUMMARY_CONFIG
```
PetarMax, line 84: Could you please add to the PR description the issue that this solves? If I'm remembering correctly, not having an exact match when running the underlying CSE functions would result in unrelated functions from various libraries being analysed?

palinatolmach: Yes yes, sure, I'm in the process of writing it :) thanks!

palinatolmach: Done!

palinatolmach:
The end-to-end failure seems to be flakiness, but some of the CSE ones get stuck, I'll investigate that.

src/kontrol/prove.py
```diff
@@ -337,6 +344,14 @@ def create_kcfg_explore() -> KCFGExplore:
                 )
 
             if proof is None:
+                # With CSE, top-level proof should be a summary if it's not a test or setUp function
+                if (
+                    options.cse
+                    and options.config_type == ConfigType.TEST_CONFIG
+                    and not test.method.contract_name.endswith('Test')
```
anvacaru, line 351: This might not hold for all test contracts. I don't think it is enforced in any way.

anvacaru: My suggestion would be to create a static function in the `Contract` class, that checks the AST of the contract to see if a contract inherits the `Test` contract, or if it has the `IS_TEST` variable in the storage. https://github.com/foundry-rs/forge-std/blob/master/src/Test.sol#L32

palinatolmach: Thanks @anvacaru! That's my concern too, but we need a way to determine if a constructor belongs to a test contract, and I couldn't think of a better one. It seems to be the case with CSE tests, because in https://github.com/runtimeverification/kontrol/pull/563 Noah used it as a way to decide on the configuration to be used for each test in `test_kontrol_cse`, and it's passing: https://github.com/runtimeverification/kontrol/blob/0b1e62950172f9101b589dd029374859456e38c2/src/tests/integration/test_kontrol_cse.py#L42-L43
But it's also getting stuck somewhere now, so I'm trying to isolate what it is.

palinatolmach: Oh wait, sorry, I guess it doesn't? https://github.com/runtimeverification/kontrol/actions/runs/9384264356/job/25839912502?pr=599

palinatolmach: > My suggestion would be to create a static function in the Contract class, that checks the AST of the contract to see if a contract inherits the Test contract, or if it has the IS_TEST variable in the storage. https://github.com/foundry-rs/forge-std/blob/master/src/Test.sol#L32

Yes, agreed, that sounds great! Will do it today.

palinatolmach: @anvacaru done in https://github.com/runtimeverification/kontrol/pull/599/commits/6351549a4c358edcc3b7627c311e5a7ea8b3a33e. Thanks for the suggestion!

I've decided to go with `IS_TEST` and leverage the `fields` you've added for it, since some Kontrol test contracts e.g., for [Optimism](https://github.com/ethereum-optimism/optimism/blob/c38ce096def52e4acdaecb7ffd2d396f464693dd/packages/contracts-bedrock/test/kontrol/proofs/OptimismPortal.k.sol#L24), don't inherit `Test` and it's easier to add an `IS_TEST` variable to them.

palinatolmach: And all tests are passing locally now.

PetarMax (approved):

anvacaru (approved):

PR #608, opened 2024-06-07
Just-In-Time jump destination computation in Kontrol

PetarMax:
This PR introduces an optimisation layer for Kontrol, in which jump destinations are always computed just-in-time and the `<jumpDest>` cell is always set to `.Set`, removing intensive set reasoning. 

Specifrically, the jump destinations are computed by checking only that the opcode is correct. This means that we are working under the assumption that the Solidity compiler will not create EVM code that jumps into `PUSH` data. Since such an execution would immediately throw an error and has not been observed so far, this appears as a safe assumption to make.


palinatolmach:
Closed in favor of https://github.com/runtimeverification/evm-semantics/pull/2471, which showed equally good performance improvements.

palinatolmach:
Closed in favor of https://github.com/runtimeverification/evm-semantics/pull/2471, which showed equally good performance improvements.

PR #600, opened 2024-06-05
CSE: Include symbolic contracts corresponding to `contract` fields into accounts

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/597.

This PR 
- introduces the `_create_cse_accounts` function that recursively ensures that code and storage are in place for addresses/accounts corresponding to `contract` fields during CSE;
- introduces lemmas needed for symbolic bytes reasoning, in particular, when handling constructors with symbolic parameters;
- adds `ContractFieldTest.testEscrowToken` CSE test that exercises new functionality by creating summaries for contracts with `contract` fields.

Follow-up issues to this and [related](https://github.com/runtimeverification/kontrol/pull/612) PRs:
- https://github.com/runtimeverification/kontrol/issues/597
- https://github.com/runtimeverification/kontrol/issues/623
- https://github.com/runtimeverification/kontrol/issues/624
- https://github.com/runtimeverification/evm-semantics/issues/2476

PR #612, opened 2024-06-10
Append constructor args to runtime bytecode

palinatolmach:
As identified by @PetarMax, we were not appending arguments to init bytecode when executing constructors taking inputs as [we should](https://docs.soliditylang.org/en/latest/contracts.html#creating-contracts). This PR
- adds a new property, [`encoded_args`](https://github.com/runtimeverification/kontrol/blob/9d2aebe6bcd9d406c4fda12e1270e5e51fd01dd8/src/kontrol/solc_to_k.py#L414), to `Constructor` class, which returns concatenated ABI-encoded arguments based on constructor's inputs 
- if the function being executed is a constructor, [appends](https://github.com/runtimeverification/kontrol/blob/9d2aebe6bcd9d406c4fda12e1270e5e51fd01dd8/src/kontrol/prove.py#L868) those arguments to the init bytecode and insert the result into the `<program>` cell in the initial CTerm and adds the corresponding type constraints
- adds a test, [`ConstructorArgsTest.test_constructor_args` ](https://github.com/runtimeverification/kontrol/blob/9d2aebe6bcd9d406c4fda12e1270e5e51fd01dd8/src/tests/integration/test-data/foundry/test/ConstructorTest.t.sol) that exercises a constructor with two symbolic arguments

There're a few follow-ups to this PR: 
- we need to support NatSpec comments for specifying length of dynamically sized types
- further refactoring of `init_cterm` arguments will be included in https://github.com/runtimeverification/kontrol/pull/600
- there's some repetition in `Method` and `Constructor` structure and input processing, with the opportunity for factoring out common logic (I've made an attempt here https://github.com/runtimeverification/kontrol/pull/542, but didn't have enough time to update this PR or do it properly yet)


src/kontrol/solc_to_k.py
```diff
@@ -411,6 +410,47 @@ def callvalue_cell(self) -> KInner:
...
+                            for i in range(input.array_lengths[0])
+                            for _c in components
+                        ]
+                        components = tuple(tuple_array_components)
+
+                    for sub_input in components:
+                        _abi_type = sub_input.to_abi()
+                        rps.extend(_range_predicates(_abi_type, sub_input.dynamic_type_length))
+                else:
+                    rps = _range_predicates(abi_type, input.dynamic_type_length)
```
PetarMax, line 446: Could this be a place to bring in the string constraints that we need? They'd need to be written using `#lookup`, so we need the storage and slot information...

palinatolmach: Yes, I think so! We should have access to the storage and slot information from [`Contract.fields`](https://github.com/runtimeverification/kontrol/blob/2fc32a5938beff42730dccc0548d30ba6969426d/src/kontrol/solc_to_k.py#L788), for example — we can pass `contract: Contract` parameter to this function; or just add this info to the `Input` object directly maybe (cc @nwatson22)

PetarMax (approved):

PR #627, opened 2024-06-13
String storage fix

nwatson22:
Adds structuring assumptions to strings in contract storage when running proofs in summary mode, allowing these proofs to finish.
Specifically, assumes for all storage slots that hold a `string`, that the least significant bit is set, which signals the length is <32 bytes, and that the length, which is stored in this case in the next 7 bits, actually <32, and is positive.


src/kontrol/prove.py
```diff
@@ -869,9 +869,10 @@ def _init_cterm(
             # Symbolic account for the contract being executed
             accounts.append(Foundry.symbolic_account(Foundry.symbolic_contract_prefix(), contract_code))
         else:
+            storage_map = KVariable(Foundry.symbolic_contract_prefix() + '_STORAGE', sort=KSort('Map'))
```
palinatolmach, line 872: Is this variable declaration needed here and as a separate argument to `_create_cse_accounts`? It gets initialized in the same manner within `_create_cse_accounts`: https://github.com/runtimeverification/kontrol/blob/fe577e524655da12a6313c3baa4ed671890b23ed/src/kontrol/prove.py#L977
On the first (non-recursive) invocation of `_create_cse_accounts`, `contract_name` is `Foundry.symbolic_contract_prefix()`, so it gets initialized with the exact same values, I think.

src/tests/integration/test_kontrol_cse.py
```diff
@@ -48,6 +48,10 @@ def test_foundry_dependency_automated(
     if bug_report is not None:
         server._populate_bug_report(bug_report)
 
+    run_constructor = True
+    if test_id in ['CallableStorageTest.test_str()', 'CallableStorageContract.str()']:
```
palinatolmach, line 52: Would `CallableStorageContract.init` fail if we run it with `--run-constructor`? Would it be sufficient to handle the engagement contracts @PetarMax?

src/kontrol/prove.py
```diff
@@ -972,6 +978,63 @@ def _create_cse_accounts(
     new_accounts.append(Foundry.symbolic_account(contract_name, contract_code, storage_map))
 
     for field in storage_fields:
+        if field.data_type == 'string':
+            lookup = KApply('lookup', [storage_map, intToken(field.slot)])
+            length_byte_lt32 = KApply(
+                '_<Int_',
```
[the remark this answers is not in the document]

src/kontrol/prove.py
```diff
@@ -972,6 +978,63 @@ def _create_cse_accounts(
...
+                                    intToken(1),
+                                ],
+                            ),
+                        ],
+                    ),
+                    intToken(0),
+                ],
+            )
+            lowest_bit_not_set = KApply(
+                '_==Int_',
```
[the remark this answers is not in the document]

src/kontrol/prove.py

src/tests/integration/test_kontrol_cse.py
PetarMax: We don't want to handle this. We need the constraints in constructors as well, either as part of this PR or the next one.

src/tests/integration/test-data/foundry/test/CallableStorageTest.t.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.13;
+
+import "forge-std/Test.sol";
+import "kontrol-cheatcodes/KontrolCheats.sol";
+
+contract CallableStorageContract {
+    uint public num;
+    string public str;
+    uint immutable imm;
```
PetarMax, line 10: Is this needed?

src/tests/integration/test-data/foundry/test/CallableStorageTest.t.sol
```diff
@@ -0,0 +1,32 @@
...
+
+    constructor(string memory a) payable {
+        str = a;
+        imm = 2;
+    }
+
+    function doNothing() public {
+        return;
+    }
```
PetarMax, line 19: This, too...

src/tests/integration/test-data/foundry/test/CallableStorageTest.t.sol
nwatson22: Removed.

src/tests/integration/test-data/foundry/test/CallableStorageTest.t.sol
nwatson22: Removed.

nwatson22:
> Let's remove the immutable and the `doNothing` function. Putting the string conditions in for when a constructor is run can be done as part of this PR as as part of next PR.

@PetarMax It's removed. I will save the string conditions for constructors for another PR. Is there something with constructors specifically with string arguments, because I thought it was the same issue for constructors and functions both (that it creates a copy loop).

PetarMax:
No, there isn't, but I am guessing the constraints aren't put in at the moment at constructor-time?

PetarMax (approved):

PR #635, opened 2024-06-21
Refining cse storage generation

PetarMax:
The storage generation mechanism for CSE was proving to be too costly in terms of performance, as it involved lots of `&Int` expressions that were slowing down Z3. This PR removes these expressions by:
- directly placing in storage address fields and contract fields, under assumption that the slot is entirely held by the address 
- directly placing in storage strings, assuming short form, reformulating the string-related constraints accordingly, and adding related simplifications.

In addition:
- the variable naming mechanism for contracts and fields has been modified to prevent name clasnes.


src/tests/integration/test-data/symbolic-bytes-lemmas.k
```diff
@@ -46,4 +51,36 @@ module SYMBOLIC-BYTES-LEMMAS
...
+
+    rule [string-is-short]:
+      1 &Int ( 2 *Int X:Int ) => 0
+      requires 0 <=Int X
+      [simplification]
+
+    //
+    // Arithmetic
+    //
+    rule ( A:Int *Int B:Int ) >>Int C:Int => ( A /Int ( 2 ^Int C ) ) *Int B
```
nwatson22, line 82: Do we want nonnegative constraints for any of `A`/`B`/`C`?

nwatson22 (approved):
Looks good, as long as the test in [src/tests/integration/test-data/show/CallableStorageContract.str().cse.expected](https://github.com/runtimeverification/kontrol/pull/627/files#diff-29a196ef734968de543cf0d0e71f105c8c7b7f935c253e01a078d2c43449da49) still only has 2 branches i.e. the different approach to structuring the storage map works the same

src/tests/integration/test-data/symbolic-bytes-lemmas.k
PetarMax: Yes, we certainly do.

src/kontrol/foundry.py
```diff
@@ -400,11 +400,11 @@ def loc_FOUNDRY_FAILED() -> KApply:  # noqa: N802
...
     @staticmethod
     def symbolic_contract_prefix() -> str:
-        return 'CONTRACT'
+        return 'C'
 
     @staticmethod
-    def symbolic_contract_id() -> str:
-        return Foundry.symbolic_contract_prefix() + '_ID'
+    def symbolic_contract_id(contract_name: str) -> str:
+        return Foundry.symbolic_contract_prefix() + '_' + contract_name.upper() + '_ID'
```
PetarMax, line 407: This allows us to have more informative variable names and a better debugging experience, while also preventing variable capture.

src/kontrol/prove.py
```diff
@@ -345,7 +346,11 @@ def create_kcfg_explore() -> KCFGExplore:
 
             if proof is None:
                 # With CSE, top-level proof should be a summary if it's not a test or setUp function
-                if options.cse and options.config_type == ConfigType.TEST_CONFIG and not test.contract.is_test_contract:
+                if (
+                    (options.cse or options.include_summaries)
+                    and options.config_type == ConfigType.TEST_CONFIG
+                    and not test.contract.is_test_contract
+                ):
```
PetarMax, line 353: Accounts should be structured whenever either `--cse` or `--include-summary` is used.

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
     new_accounts: list[KInner] = []
     new_account_constraints: list[KApply] = []
 
-    storage_map = KVariable(contract_name + '_STORAGE', sort=KSort('Map'))
-    new_accounts.append(Foundry.symbolic_account(contract_name, contract_code, storage_map))
+    storage_map: KInner = KVariable(contract_name + '_STORAGE', sort=KSort('Map'))
+
+    singly_occupied_slots = [
+        slot for (slot, count) in Counter([field.slot for field in storage_fields]).items() if count == 1
+    ]
```
PetarMax, line 994: Currently, we are limited to handling only the cases in which addresses and contracts in storage occupy an entire slot. This should be one of the immedate next PRs.

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
...
+                # The offset must equal zero
+                assert field.offset == 0
+                # Create appropriate symbolic variable
+                field_variable = KVariable(field_name + '_ID', sort=KSort('Int'))
+                # Update storage map accordingly: ( field.slot |-> contract_account_variable ) STORAGE_MAP
+                storage_map = KApply('_Map_', [map_item(intToken(field.slot), field_variable), storage_map])
+                address_range_lb = leInt(intToken(0), field_variable)
+                address_range_ub = ltInt(field_variable, intToken(1461501637330902918203684832716283019655932542976))
+                new_account_constraints.append(mlEqualsTrue(address_range_lb))
+                new_account_constraints.append(mlEqualsTrue(address_range_ub))
```
PetarMax, line 1040: Addresses are assumed to be initialised by the constructor and are placed in storage explicitly.

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
...
+                        # TODO: Support shared slots for contract variables
+                        raise (
+                            ValueError(
+                                'Unsupported: CSE for contracts with contract or interface fields in shared slots.'
+                            )
+                        )
 
                     contract_accounts, contract_constraints = _create_cse_accounts(
-                        foundry, contract_obj.fields, contract_account_name, contract_account_code
+                        foundry, contract_obj.fields, field_name, contract_account_code
```
PetarMax, line 1080: This prevents variable capture.

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
...
-                KApply(
-                    '_&Int_',
-                    [
-                        intToken(1),
-                        lookup,
-                    ],
-                ),
+            string_contents = KVariable(field_name + '_S_CONTENTS', sort=KSort('Bytes'))
+            string_length = KVariable(field_name + '_S_LENGTH', sort=KSort('Int'))
+            string_structure = KApply(
```
PetarMax, line 1002: Strings are put in storage directly, and we are assuming that they are in short form. We should handle their long form only when a reasonable use case comes up.

palinatolmach (approved):

src/kontrol/prove.py
```diff
@@ -864,14 +872,15 @@ def _init_cterm(
         init_subst.update(init_subst_test)
     else:
         accounts: list[KInner] = []
+        contract_account_id = Foundry.symbolic_contract_prefix() + '_' + contract_name
```
anvacaru, line 875: why not `Foundry.symbolic_contract_id(contract_name)` instead?
```suggestion
        contract_account_id = Foundry.symbolic_contract_id(contract_name)
```

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
     new_accounts: list[KInner] = []
     new_account_constraints: list[KApply] = []
 
-    storage_map = KVariable(contract_name + '_STORAGE', sort=KSort('Map'))
-    new_accounts.append(Foundry.symbolic_account(contract_name, contract_code, storage_map))
+    storage_map: KInner = KVariable(contract_name + '_STORAGE', sort=KSort('Map'))
+
+    singly_occupied_slots = [
+        slot for (slot, count) in Counter([field.slot for field in storage_fields]).items() if count == 1
```
anvacaru, line 993: I don't think you need the intermediate list.

```suggestion
        slot for (slot, count) in Counter(field.slot for field in storage_fields).items() if count == 1
```

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
...
-                    '_&Int_',
-                    [
-                        intToken(1),
-                        lookup,
-                    ],
-                ),
+            string_contents = KVariable(field_name + '_S_CONTENTS', sort=KSort('Bytes'))
+            string_length = KVariable(field_name + '_S_LENGTH', sort=KSort('Int'))
+            string_structure = KApply(
+                'asWord',
```
anvacaru, line 1003: Here you can use `as_word`, imported `from kevm_pyk.kevm`.
```suggestion
            string_structure = as_word (
```

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
...
+                [
+                    KApply(
+                        '_+Bytes__BYTES-HOOKED_Bytes_Bytes_Bytes',
+                        [
+                            string_contents,
+                            KApply(
+                                'buf',
```
anvacaru, line 1010: In a similar way, there is a `buf` method.

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
...
+            new_account_constraints.append(mlEqualsTrue(string_length_ub))
+        # Processing of addresses
+        if field.data_type == 'address':
+            if field.slot in singly_occupied_slots:
+                # The offset must equal zero
+                assert field.offset == 0
+                # Create appropriate symbolic variable
+                field_variable = KVariable(field_name + '_ID', sort=KSort('Int'))
+                # Update storage map accordingly: ( field.slot |-> contract_account_variable ) STORAGE_MAP
+                storage_map = KApply('_Map_', [map_item(intToken(field.slot), field_variable), storage_map])
```
anvacaru, line 1036: There is also the `map_of` function from `pyk.prelude.collections` that can be used to generate a Map. It uses `build_assoc` instead.

src/kontrol/prove.py
```diff
@@ -978,101 +987,104 @@ def _create_cse_accounts(
...
                         )
                     )
+                    # The contract address must occupy the entire slot
+                    if field.slot in singly_occupied_slots:
+                        # The offset must equal zero
+                        assert field.offset == 0
+                        # Update storage map accordingly: ( field.slot |-> contract_account_variable ) STORAGE_MAP
+                        storage_map = KApply(
+                            '_Map_', [map_item(intToken(field.slot), contract_account_variable), storage_map]
+                        )
```
anvacaru, line 1070: This part seems duplicated, could you extract it in an auxiliary function?

src/kontrol/prove.py
PetarMax: Yes, I'm not a fan of that one because of `.Map` - will it get simplified? I could try.

src/kontrol/prove.py
PetarMax: Because that will already have `_ID`, and then the `symbolic_account` would add `_ID`, `_BAL`, etc. to that. That, however, indicates that the `symbolic_contract_id` is not defined well.

PetarMax: I've now differentiated between `symbolic_account_name` and `symbolic_account_id`.

src/kontrol/prove.py
PetarMax: Correct!

src/kontrol/prove.py
PetarMax: Yes.

src/kontrol/prove.py
PetarMax: Yes.

src/kontrol/prove.py
PetarMax: Actually, that doesn't work, because that works for key-value pairs, I have a key-value pair and the rest.

src/kontrol/prove.py
PetarMax: Done.

anvacaru (approved):
LGTM :+1:

PR #638, opened 2024-06-24
Add `--mt` as a `--match-test` alias

JuanCoRo:
Foundry has `--mt` as an alias for `--match-test`, and we currently don't support that argument. This PR makes `--mt` available to be able to execute proofs by doing `kontrol prove --mt $proof`


palinatolmach (approved):
Thanks for this PR @JuanCoRo! It's unironically a big UX improvement :)

PR #642, opened 2024-06-24
Add pretty feedback from Kontrol

JuanCoRo:
Kontrol's feedback, albeit informative, is a bit dry. This PR aims to increase the dopamine peak users feel when executing `kontrol build` or `kontrol prove`.

As of the opening of this PR, this is what is enhanced (apologies for the flakiness of the screenshots):
- Successful `kontrol build` message:  
image
- Failing `kontrol build` message:
image1
- Proofs that are being run:
image2
- Successful `kontrol prove` summary:
image3
- Unsuccessful `kontrol prove` summary:
Image4


ehildenb:
DOPAMINE!!!!

ehildenb:
Perhaps another thing to address: currently users run `kontrol prove ...`, and then they think it's hanging because it just isn't outputting anything. We then say "add `--verbose` to the `kontrol prove ...`, and you'll see intermediate output!". Perhaps the prover could output information about progress as it goes? eg. it could output th ecurrent number of frontier nodes, etc...

Basically, it could output what `kontrol list ...` outputs, each time there is an update to the proof, and in a more compact format? If it could even just output it to the same line, so it's not a big scrolling line of text, that would be cool too. Maybe for another issue.

JuanCoRo:
Yeah definitely!! I took a stab at this, and from what I've seen, it is definitely possible. It could be laid out in a very cool format.
I just stopped at what I've shared in the screenshots because I most pressingly wanted the pretty success summaries. But having a real-time progress message shouldn't be too difficult!

palinatolmach:
Foundry has just added some progress output for invariant tests, maybe we can use it for some inspiration: https://x.com/lucasmanuel_eth/status/1801353643410530393. 

On the prover output side, generally, I meant to check with the backend team tomorrow if we can get some of the progress messages back :). But in the meantime suggesting to use `--verbose` is a good workaround too, @JuanCoRo do you think it makes sense to incorporate it in this PR?

JuanCoRo:
@palinatolmach you mean adding a message when not providing the `--verbose` flag to suggest using it?

JuanCoRo:
Thinking about it, it might be that the `kontrol prove` function requires more work than what this PR is intended to do, which is to give a glow-up to the outputs.
For the `kontrol build` command I've changed it so that there is the following message with spinning dots while Kontrol is building:
image
And it disappears once the project has been built:
image

Note that the transient "Building Kontrol project" is in RV blue! And the rotating dots in RV yellow! Let me know if you'd like to keep this or if we should just revert to regular black and green. It might be the case that for light-themed terminals like mine, RV yellow is not the most legible. Likewise, if we want a transient "Building Kontrol project" or we'd like for it to be recorded as terminal output.

Also also, feel free to discuss emojis :slightly_smiling_face:

palinatolmach:
Makes sense, thanks @JuanCoRo! Let me try running it in my terminal to see how it looks in a terminal with darker background, but it looks great on screenshots!

palinatolmach (approved):
It looks great to me, thanks @JuanCoRo! 

One minor note is that, in my terminal, it looks like a whitespace might be missing between the checkmark and "Success!", but it looks good in your screenshots, so maybe it's my issue:
image
Upd: the same is true for all messages starting with emojis, actually. But some of the logs from the backends are back now :) 
image

anvacaru (approved):
Nice!!

JuanCoRo:
Thanks a lot @palinatolmach for trying it out!!
Let me add some whitespace and confirm that it looks good on your machine! I think a lot of people are using macs, so better get it right!
Also, I just realized that the "Add `--verbose` to `kontrol build`" message will show up even if `--verbose` is already present :sweat_smile: I'll address this as well!

JuanCoRo:
So it seems we can do very cool stuff with [Rich](https://rich.readthedocs.io/en/stable/). But for the moment, we get some flaky behavior when it interacts with the current logger:
image
I will leave this PR as is now, without any progress message for `kontrol prove`, and open an issue to improve the coolness of Kontrol's feedback even further.
Notice that, as of now, with the `--verbose` option in  Kontrol build, there is a bit of the above flakiness. But I don't think it is that bad. Let me know what you think!

PR #646, opened 2024-06-25
change rekompilation conditions and fix inclusion of setup function into proof

ovatman:
This PR performs two changes on how rekompilation and inclusion of setup in the proof is handled:
- Before the PR, a rekompilation is triggered only when there's a change in .k files generated from the foundry sources. This results in a rekompilation only when the interface of the contract changes since these generated files are based only on the function signatures. After the PR, any change in the foundry source files will trigger a rekompilation.
- Before the PR, without the reinit parameter the `setUp()` function in the test would not have been included in the proof even when the function contents change. After the PR, independent of the reinit flag, a change in the `setUp()` function would always result in the inclusion of it in the proof.


.vscode/launch.json
```diff
@@ -0,0 +1,75 @@
+{
```
anvacaru, line 1: Are these commited on purpose? If not, you could add a new rule to `.gitignore`.

src/kontrol/foundry.py
```diff
@@ -518,32 +518,30 @@ def resolve_setup_proof_version(
...
 
+        if not method.up_to_date(self.digest_file):
+            _LOGGER.info(f'Creating a new version of {test} setup proof because it was updated.')
+            return self.free_proof_version(test)
+
         if reinit:
             if user_specified_setup_version is None:
                 _LOGGER.info(
-                    f'Creating a new version of {test} because --reinit was specified and --setup-version is not specified.'
+                    f'Creating a new version of {test} setup proof because --reinit was specified and --setup-version is not specified.'
```
palinatolmach, line 528: Maybe we can make it more specific using something like this?
```
f'... a new version of `{test.contract.contract_name}.setUp'
```

palinatolmach:
@ovatman can I just check if my understanding on the second item in the PR description is correct: we'll now re-run `setUp` even if there's no `--reinit` flag provided?

src/kontrol/foundry.py
```diff
@@ -518,32 +518,30 @@ def resolve_setup_proof_version(
         _, method = self.get_contract_and_method(test)
         effective_test_version = 0 if test_version is None else self.free_proof_version(test)
 
+        if not method.up_to_date(self.digest_file):
+            _LOGGER.info(f'Creating a new version of {test} setup proof because it was updated.')
```
palinatolmach, line 522: Would it create a new version only for test setup, or for the test itself too?

src/kontrol/foundry.py
palinatolmach: Ah, thanks! Do we need a separate `"setup proof"` then? But it looks fine to me either way.

palinatolmach (approved):
Looks good, thanks! As a potential follow-up, @ovatman if you get a chance to, can you please take a look at the documentation (https://github.com/runtimeverification/gitbook-kontrol) to see how we can clarify it there too?

anvacaru (approved):
LGTM! :+1:

PR #661, opened 2024-07-02
fixes for the issues on toml related argument parsing

ovatman:
The folowing fixes are done
- The defualt value for the config file is set to `kontrol.toml` under the foundry project root
- toml arguments are also taken into consideration when setting the logging level
- `--cse` option's default value is set to `None` for the `argparse`


JuanCoRo:
Could we use this PR to default `verbose` to `false` in both `prove` and `build`? And the same for `regen` and `rekompile`?

src/kontrol/__main__.py
```diff
@@ -375,15 +376,30 @@ def exec_init(options: InitOptions) -> None:
...
+    elif hasattr(args, 'debug') and args.debug:
         return logging.DEBUG
 
-    if hasattr(args, 'verbose') and args.verbose:
+    if (not hasattr(args, 'verbose') or args.verbose is None) and ('verbose' in toml_args and toml_args['verbose']):
+        return logging.INFO
+    elif hasattr(args, 'verbose') and args.verbose:
         return logging.INFO
 
     return logging.WARNING
```
anvacaru, line 390: You can extract this logic into an auxiliary function to improve readability. Also, maybe using `getattr` instead of `hasattr`. Something like
```py
def _loglevel(args: Namespace, toml_args: dict) -> int:
    def is_attr_used(attr_name: str) -> bool:
        return getattr(args, attr_name, None) is not None or toml_args.get(attr_name)

    if is_attr_used('debug'):
        return logging.DEBUG

    if is_attr_used('verbose'):
        return logging.INFO

    return logging.WARNING
```

src/kontrol/__main__.py
```diff
@@ -375,15 +376,30 @@ def exec_init(options: InitOptions) -> None:
...
 
 
+def _config_file_path(args: Namespace) -> Path:
+    return (
+        Path.joinpath(
+            Path('.') if not hasattr(args, 'foundry_root') or args.foundry_root is None else args.foundry_root,
+            'kontrol.toml',
+        )
+        if not hasattr(args, 'config_file') or args.config_file is None
```
anvacaru, line 399:
```suggestion
        if not hasattr(args, 'config_file') or args.config_file is None
        if getattr(args, 'config_file', None) is not None:
```

JuanCoRo:
Thanks @ovatman! I think I was getting them set to true via `kontrol init`, but perhaps that's another issue

anvacaru (approved):

PR #663, opened 2024-07-02
Support symbolic contract storage slots with nonzero offsets

nwatson22:
Handles the case for CSE where contracts in storage share a storage slot with other variables, with a nonzero offset. Since an address is 20 bytes, and a storage slot is 32, only 1 address may ever be in a single storage slot, so we represent that storage slot as something like: `([slot idx] |-> #asWord(AFTER +Bytes #buf(20, ACCOUNT) +Bytes BEFORE)` with the appropriate lengthBytes constraints on `BEFORE` and `AFTER`.


palinatolmach (commented):
lgtm, @PetarMax can you please take a look at the lemma?

src/tests/integration/test-data/cse-lemmas.k
```diff
@@ -55,6 +55,12 @@ module CSE-LEMMAS
        andBool X +Int 1 <=Int A
        [concrete(X), simplification, preserves-definedness]
 
+    rule { #asWord( B1 +Bytes B2 ) #Equals #asWord( B3 +Bytes B4 ) } 
+      => { B1 #Equals B3 } #And { B2 #Equals B4 }
+    requires lengthBytes(B1 +Bytes B2) <=Int 32
+    andBool lengthBytes(B1) ==Int lengthBytes(B3) 
+    andBool lengthBytes(B2) ==Int lengthBytes(B4) [simplification]
```
PetarMax, line 62: You only need one of these two, I think.

PetarMax: Actually, note [this KEVM simplification](https://github.com/runtimeverification/evm-semantics/blob/bd4e443f80d827355236fb2cd4d3a4aaec75d075/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k#L21). I think that it would be better not to repeat its logic, but to instead use
```
rule { #asWord ( X ) #Equals #asWord ( Y ) } => #Top
  requires X ==K Y
  [simplification]
```
which would then piggyback onto the existing ones for `Bytes`.

src/kontrol/prove.py
```diff
@@ -1147,10 +1147,55 @@ def extend_storage(map: KInner, slot: int, value: KInner) -> KInner:
...
+                                    '_+Bytes__BYTES-HOOKED_Bytes_Bytes_Bytes',
+                                    [
+                                        KApply(
+                                            '_+Bytes__BYTES-HOOKED_Bytes_Bytes_Bytes',
+                                            [
+                                                slot_var_after,
+                                                KEVM.buf(intToken(20), contract_account_variable),
+                                            ],
+                                        ),
+                                        slot_var_before,
```
PetarMax, line 1165: Please have `before` on the LHS and `after` on the RHS since the indexing of `+Bytes` is left-to-right, and also have the `+Bytes` be right-associative so that no further simplifications are needed to transform this into the canonical form.

src/kontrol/prove.py
```diff
@@ -1147,10 +1147,55 @@ def extend_storage(map: KInner, slot: int, value: KInner) -> KInner:
...
+                            '_Map_', [map_item(intToken(field.slot), masked_contract_account_var), storage_map]
+                        )
+                        new_account_constraints.append(
+                            mlEqualsTrue(
+                                KApply(
+                                    '_==K_',
+                                    [
+                                        KApply(
+                                            'lengthBytes(_)_BYTES-HOOKED_Int_Bytes',
+                                            [slot_var_before],
```
PetarMax, line 1180: `slot_var_after`

src/kontrol/prove.py
```diff
@@ -1147,10 +1147,55 @@ def extend_storage(map: KInner, slot: int, value: KInner) -> KInner:
...
+                            )
+                        )
+                        new_account_constraints.append(
+                            mlEqualsTrue(
+                                KApply(
+                                    '_==K_',
+                                    [
+                                        KApply(
+                                            'lengthBytes(_)_BYTES-HOOKED_Int_Bytes',
+                                            [slot_var_after],
```
PetarMax, line 1194: `slot_var_before`

src/tests/integration/test-data/show/ContractFieldTest.testEscrowToken().cse.expected
```diff
@@ -5,7 +5,7 @@
 │   callDepth: 0
 │   statusCode: STATUSCODE:StatusCode
 │
-│  (2219 steps)
+│  (2231 steps)
```
PetarMax, line 8: Why would this change, do we know?

src/tests/integration/test-data/show/ContractFieldTest.testEscrowToken().cse.expected
```diff
@@ -130,7 +130,7 @@ module SUMMARY-TEST%CONTRACTFIELDTEST.TESTESCROWTOKEN():0
                        0
                      </balance>
                      <storage>
-                       ( .Map => ( 0 |-> 491460923342184218035706888008750043977755113263 ) )
+                       ( .Map => ( 0 |-> 125813996375599159817140963330240011258305308995328 ) )
```
PetarMax, line 133: Or this?

src/tests/integration/test-data/show/ContractFieldTest.testEscrowToken().cse.expected
PetarMax: Ah, because of `spacer`.

src/tests/integration/test-data/show/ContractFieldTest.testEscrowToken().cse.expected
PetarMax: Same reason as above.

PetarMax (changes requested):
Minor changes required, left in comments.

src/kontrol/prove.py
nwatson22: The reason I have `before` on the right and `after` on the left is because I was trying to refer to the order in which the storage fields are be declared in the contract, so `before` would be those fields in the same slot declared before the contract and `after` those declared after. Does it not do that or do you not like that convention?

src/tests/integration/test-data/cse-lemmas.k
nwatson22: Replaced with the new lemma.

src/kontrol/prove.py
nwatson22: I changed it so the `+Bytes` term is right-grouped to start.

PetarMax: Since `+Bytes` indexing is always left-to-right, for me 'before' in `Bytes` means 'to the left' (not sure about others), that's all, and since we will be debugging code like this likely often, I'm thinking that it's better to not break that intuition.

src/kontrol/prove.py

src/kontrol/prove.py

src/kontrol/prove.py
nwatson22: Ok, thanks. I've changed the ordering.

src/tests/integration/test-data/cse-lemmas.k
nwatson22: The combination of the two lemmas
```
    rule { #asWord ( X ) #Equals #asWord ( Y ) } => #Top
      requires X ==K Y
      [simplification]
```
and
```
    rule [bytes-equal-concat-split-ml]:
      A:Bytes +Bytes B:Bytes ==K C:Bytes +Bytes D:Bytes => (A ==K C) andBool (B ==K D)
      requires lengthBytes(A) ==Int lengthBytes(C)
        orBool lengthBytes(B) ==Int lengthBytes(D)
      [simplification]
```
Does not eliminate the branching we were trying to eliminate because what my original lemma was doing was getting byte equality out of word equality.
Would this one be acceptable instead to work in tandem with `bytes-equal-concat-split-ml`?
```
    rule { #asWord ( X ) #Equals #asWord ( Y ) } => { X #Equals Y }
      requires lengthBytes(X) ==Int lengthBytes(Y) andBool lengthBytes(X) <=Int 32
      [simplification]
```

PetarMax: Oh that's interesting. Yes, that should work.

palinatolmach (approved):

PetarMax (approved):

Issue #665, opened 2024-07-04
CSE: events are identified as function calls


anvacaru:
Given a test such as:

```Solidity
    function testTransferSuccess_0(address alice, uint256 amount)
      public {
        _notBuiltinAddress(alice);
        vm.assume(alice != address(0));
        uint256 balanceAlice = gld.balanceOf(alice);
        vm.assume(balanceAlice >= amount);
        vm.expectEmit(true, true, false, true);
        emit IERC20.Transfer(alice, alice, amount);
        vm.startPrank(alice);
        gld.transfer(alice, amount);
        assertEq(gld.balanceOf(alice), balanceAlice);
    }
```

the `IERC20.Transfer(address,address,uint256)` event is identified as a dependency.
```
For test test%GLDTokenTest.testTransferSuccess_0(address,uint256), found external calls: [('GLDToken.balanceOf(address)', None), ('IERC20).Transfer(address,address,uint256)', None), ('GLDToken.transfer(address,uint256)', None)]       
```

Leading to an error, because the function identifier does not exist.
```
ValueError: Test identifiers not found: {'(^|%)(IERC20\\).Transfer\\(address,address,uint256\\))$'}                                                                                                                
```

PR #666, opened 2024-07-04
CSE: Check if external function is an event

anvacaru:
Fixes: #665

Adding a check in `_find_function_calls` to ignore events emitted.


palinatolmach (approved):
thanks for identifying and fixing it @anvacaru!

PR #675, opened 2024-07-07
Support expectRevert and expectEmit in CSE

anvacaru:
Summaries would not be applied when an `expectRevert` or an `expectEmit` is active because, in CSE's initial state, these are initialized with concrete values.

Setting them to be symbolic in the context of CSE.


JuanCoRo (approved):
LGTM!

PetarMax (approved):
The expected output changes are the expected ones. Ok to merge when CSE tests pass.

PR #680, opened 2024-07-08
Mock function cheatcode

nwatson22:
Adds `mockFunction` cheat code, which makes all future calls to a specific calldata match behave as if they are calling a different, specified, contract with the same call data. Adds tests for the cases of mocking a function with no arguments, mocking with one argument but no argument specified, and mocking with argument fully specified.


src/kontrol/kdist/cheatcodes.md
```diff
@@ -962,6 +968,60 @@ We use `#next[OP]` to identify OpCodes that represent function calls. If there i
...
+           ( .Bag
+            => <mockFunction>
+                  <mockFunctionAddress> MOCKADDRESS </mockFunctionAddress>
+                  <mockFunctionValues> .Map [ MOCKCALLDATA <- MOCKTARGET ] </mockFunctionValues>
+               </mockFunction>
+           )
+           ...
+         </mockFunctions>
+```
+
```
palinatolmach, line 991: Is it intentional that the definition of `mockFunction`-related rules is split into three `k` blocks separated by several empty lines? They could probably be combined into one?

palinatolmach (commented):
lgtm other than one formatting comment, but I'm wondering if we should also clean `mockFunction` after a call is completed, e.g., similar to `clearPrank`, so it can be used twice?

@lucasmt would be able to give this PR a try, if you have time?

src/kontrol/kdist/cheatcodes.md
```diff
@@ -962,6 +968,45 @@ We use `#next[OP]` to identify OpCodes that represent function calls. If there i
       [priority(30)]
 ```
 
+
+Mock functions
+----------
```
nwatson22, line 973:
```suggestion
Mock functions
--------------
```

src/kontrol/kdist/cheatcodes.md
nwatson22: I've moved the `setMockFunction` section down after the `setMockCall` stuff in the utils section and combined `[cheatcode.call.mockFunction]` and `[foundry.set.mockFunction]` under their own header.

I'm not sure what you mean about the cleaning after a call is completed. It is intended to work similar to the `mockCall` cheatcode which I thought was supposed to mock all subsequent matching calls.

palinatolmach: Thanks! You're right about not having to clean `mockFunctions`, I didn't realise we want to mock subsequent matching calls.

palinatolmach:
I'll merge it now for the sake of the Friday demo/workshop, let's update it later if @lucasmt identifies any changes needed for the engagement.

palinatolmach (approved):

PR #682, opened 2024-07-08
Skip debug JSON files in from the output directory

anvacaru:
With the latest 0.8.26 version of the solc compiler, solc-to-k has struggles with json files like 
```json
{
  "_format": "hh-sol-dbg-1",
  "buildInfo": "../../../../build-info/2eec03df821d5a4eb672f3c8173e1d07.json"
}
```

throwing errors like `An error occurred while building your Kontrol project: 'id'`. 

Most likely this is not the only error with solc 0.8.26.


src/kontrol/foundry.py
```diff
@@ -187,7 +187,11 @@ def find_enums(dct: dict) -> None:
             contract_name = json_path.split('/')[-1]
             contract_json = json.loads(Path(json_path).read_text())
             contract_name = contract_name[0:-5] if contract_name.endswith('.json') else contract_name
-            contract = Contract(contract_name, contract_json, foundry=True)
+            try:
+                contract = Contract(contract_name, contract_json, foundry=True)
+            except KeyError:
+                _LOGGER.warning(f'Skipping non-compatible JSON file for contract: {contract_name} at {json_path}.')
```
palinatolmach, line 193: Would it make sense to suggest downgrading Solidity version to 0.8.25?

palinatolmach (approved):

src/kontrol/foundry.py
palinatolmach: Ok, I can't confirm that it's caused by this specific version; let's merge as is and pinpoint the issue later based on user feedback.

PR #687, opened 2024-07-10
Add option to omit enum constraints

nwatson22:
Since sometimes we don't care about adding constraints based on enum information from the compiled solidity output, and the user may want to avoid having to change the names of enums in their codebase to avoid the naming conflict error, adds `--omit-enum-constraints` option which disables the feature. When you get the naming conflict error, suggests using this option.


lucasmt:
This is working, but it seems the option needs to be passed to every call of `kontrol`, including `kontrol show`, which is inconvenient. Is there any way to make it so it only needs to be passed once to `kontrol build`?

nwatson22:
> This is working, but it seems the option needs to be passed to every call of `kontrol`, including `kontrol show`, which is inconvenient. Is there any way to make it so it only needs to be passed once to `kontrol build`?

@lucasmt There would be, but I'd have to just dump the option somewhere in the `kontrol build` output, since the `Foundry` object is reinitialized every time a new command is run. There's no existing options that work like this, as far as I know. Maybe it would be useful for other options to be determined at build time as well? Or might it limit flexibility?

We support putting options in a config file `kontrol.toml` under the foundry project dir, so that is one way you can avoid having to add the option every time. E.g., to have the option be always on for commands `prove` and `show` you could have
`kontrol.toml`
```
prove = { omit-enum-constraints = false }
show = { omit-enum-constraints = false }
```

lucasmt:
@nwatson22 That's fair, the `kontrol.toml` option might be a good way to do it then.

nwatson22:
@lucasmt @palinatolmach We decided it would be best to just invert the option, so it is off by default. That way for most commands where it's not actually relevant it can be omitted safely and we can just turn it on when we need to eliminate specific branches on `prove`s.

src/kontrol/__main__.py
```diff
@@ -156,7 +166,7 @@ def _compare_versions(ver1: KVersion, ver2: KVersion) -> bool:
 def exec_load_state(options: LoadStateOptions) -> None:
     foundry_state_load(
         options=options,
-        foundry=_load_foundry(options.foundry_root),
+        foundry=_load_foundry(options.foundry_root, add_enum_constraints=not options.enum_constraints),
```
palinatolmach, line 169: Should `not` be removed now that the option is inverted? Or is it intentional? Same for other `_load_foundry` calls below.

```suggestion
        foundry=_load_foundry(options.foundry_root, add_enum_constraints=options.enum_constraints),
```

src/kontrol/cli.py
```diff
@@ -168,6 +168,13 @@ def foundry_args(self) -> ArgumentParser:
             type=dir_path,
             help='Path to Foundry project root directory.',
         )
+        args.add_argument(
+            '--enum-constraints',
+            dest='enum_constraints',
+            default=None,
+            action='store_true',
+            help='Skip adding constraints for enum function arguments and storage slots.',
```
palinatolmach, line 176: Should this change too?
```suggestion
            help='Add constraints for enum function arguments and storage slots.',
```

src/kontrol/__main__.py
```diff
@@ -90,9 +90,19 @@ def _ignore_arg(args: dict[str, Any], arg: str, cli_option: str) -> None:
         args.pop(arg)
 
 
-def _load_foundry(foundry_root: Path, bug_report: BugReport | None = None, use_hex_encoding: bool = False) -> Foundry:
+def _load_foundry(
+    foundry_root: Path,
+    bug_report: BugReport | None = None,
+    use_hex_encoding: bool = False,
+    add_enum_constraints: bool = True,
```
palinatolmach, line 97: Is there a reason for making it `True` by default here, while the `--enum-constraints`' default is None?
```suggestion
    add_enum_constraints: bool = True,
```

nwatson22: Fixed.

src/kontrol/cli.py
nwatson22: Fixed.

palinatolmach (approved):
Left one more comment, otherwise lgtm, thanks!

Issue #505, opened 2024-04-03
Improve hex encoding, use it by default in kontrol show


anvacaru:
With, https://github.com/runtimeverification/kontrol/pull/503 we add support for the hex encoding in the `view-kcfg` TUI. In addition to this, there is also the `--use-hex-encoding` flag that will enable the hex encoding in `kontrol show`.

Now the next step is to set the hex encoding active by default, replacing `--use-hex-encoding` with `--no-use-hex-encoding`.

In addition to this, I can confirm that the use of the hex encoding in `pretty_print` does not interfere with the `kontrol show --to-module` feature, used in CSE.

anvacaru:
output should be encoded as strings?

(with `--use-hex-encoding`))
<img width="491" alt="image" src="https://github.com/runtimeverification/kontrol/assets/16517508/ce1f01a1-6a8e-4626-8ecf-538e7e2f3409">

(without `--use-hex-encoding`)
<img width="491" alt="image" src="https://github.com/runtimeverification/kontrol/assets/16517508/c609501c-909a-415d-a232-0ffa7fc27087">

palinatolmach:
Initial implementation was done in https://github.com/runtimeverification/kontrol/pull/688, however, it is still missing a few things:
- leaving integer constants as they are, e.g., 32 instead of 0x1000000000000000000000000;
- preserving aliases such as maxUint256
- changing default values in functions

PR #688, opened 2024-07-11
Improve hex encoding, use it in `kontrol show` by default

jinxinglim:
Merging of this PR will close issue https://github.com/runtimeverification/kontrol/issues/505.

This PR changes the default setting for `kontrol show` to use and display hex encoding and if the user does not want to use hex encoding, he/she can add the option `--no-use-hex-encoding`.

### Questions

1. Is the help message in this argument okay? Please feel free to suggest a better one. 
    ```python
        show_args.add_argument(
            '--no-use-hex-encoding',
            dest='use_hex_encoding',
            default=None,
            action='store_false',
            help='Print elements in non-hexadecimal encoding.',
        )
    ```

2. I changed the default `use_hex_encoding` in some of the function, e.g., `__init__` from `src/kontrol/foundry.py`, from `False` to `True` and I am not sure if it will affect its dependencies. From the CI failed tests, it seems like it is affecting its dependencies? Should I change them back to the default `False`?

Added by Palina on 25/07:

We're missing a few things: 
- leaving integer constants as they are, e.g., `32` instead of `0x1000000000000000000000000`;
- preserving aliases such as `maxUint256` 
- changing default values in functions back


anvacaru:
One limitation of our current hex encoding is that aliases such as `maxUInt256` are no longer displayed in the CTerm. Only the hex value is displayed.

jinxinglim:
> One limitation of our current hex encoding is that aliases such as `maxUInt256` are no longer displayed in the CTerm. Only the hex value is displayed.

@anvacaru Do you think this limitation should be fix in this PR? I am not too sure how easy/difficult this fix will be. If yes, could I trouble you to point me to files that I can work on? Thank you!

anvacaru:
>@anvacaru Do you think this limitation should be fix in this PR?

No, that's out of the scope of this PR. I wanted to bring it up, as it could make the kcfg less readable in a way. @PetarMax, what do you think?

palinatolmach:
We'll need to bring `True` as default value for `use_hex_encoding` and update the output before merging, but I agree that `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` is less readable than `maxUint256`, and I'm worried it'll interfere with the engagement. Good call @anvacaru, thanks! Let's probably postpone merging it for a few weeks or until we figure out how to continue using aliases?

palinatolmach:
We're missing a few things: 
- leaving integer constants as they are, e.g., `32` instead of `0x1000000000000000000000000`;
- preserving aliases such as `maxUint256` 
- changing default values in functions back

palinatolmach:
Closing for now, since we won't make hex encoding the default in `kontrol show` until it's improved. The list of improvements needed is available in https://github.com/runtimeverification/kontrol/issues/505.

PR #700, opened 2024-07-16
Update warning on missing ast flag

drinkius:
When running kontrol in a previously used repo with Foundry project that doesn't have **ast = true** flag in foundry.toml - just adding a flag doesn't work, the user also needs to re-build the project so **forge clean** is necessary. We need to update the error message to reflect that


palinatolmach:
Thanks @drinkius! We're working on some changes to allow CI runs on PRs authored by external contributors, will get it merged asap.

palinatolmach (approved):

anvacaru (approved):

Issue #698, opened 2024-07-16
To add a new `_notExternalAddress` function in `src/kontrol/state_record.py`


jinxinglim:
While working on using [external computation](https://github.com/runtimeverification/kontrol-workshop-ethcc-7/tree/main/4-external-computation) together with `fork-url $FORK_URL` on https://github.com/runtimeverification/ercx-kontrol, using `vm.prank(caller)` or `vm.startPrank(caller)`, where `caller` is a symbolic address, will cause the proof to branch on each account found in the `state-dump/StateDump.json`. 

This will cause the test to branch unnecessarily especially when there are many accounts retrieved from `fork-url $FORK_URL`.

It would be good to add a new Solidity function, say `_notExternalAddress` that works similarly to our usual `_notBuiltinAddress`, which will `vm.assume(caller != <address retrieved>)` for all `<address retrieved>`.

Code to work on: https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/state_record.py#L122

@JuanCoRo Let me know if my explanation is clear and if this is a good idea. Thank you.

PR #701, opened 2024-07-17
Generate `_notExternalAddress` function in RecordedState

jinxinglim:
As mentioned in issue https://github.com/runtimeverification/kontrol/issues/698, this PR adds a new `_notExternalAddress` function in the generated main contract via `generate_main_contract` function in `src/kontrol/state_record.py`. The goal is to make this function available for user to use it whenever he/she wants to make sure that the new caller does not clash with any of the addresses retrieved from external computation.

Here is an example of the generated function: https://github.com/runtimeverification/ercx-kontrol/blob/fork-testing/ERCx_test_suites/kontrol/RecordedState.sol#L186.


palinatolmach (approved):
lgtm, thanks @jinxinglim! @JuanCoRo shall we merge it now and use in the demo?

anvacaru (approved):

PR #720, opened 2024-07-23
KCFG Readability improvements

nwatson22:
- By default, `<program>` and `<code>` cells are printed instead of the literal byte string as the name of the contract to which they correspond, if the lookup is successful, and the `<jumpDests>` cell is abbreviated.
- Adds `--expand-config` option to show the real configuration without these simplifications.
- Displays currently executing function, based on matching first 4 bytes of calldata to function ID and matching code to known contract, for each node in the KCFG.


src/kontrol/foundry.py
```diff
@@ -1229,6 +1292,31 @@ def print_node(self, kcfg: KCFG, node: KCFG.Node) -> list[str]:
...
+            selector_bytes = None
+            if type(calldata_cell) is KToken:
+                selector_bytes = ast.literal_eval(calldata_cell.token)
+                selector_bytes = selector_bytes[:32]
+            elif (
+                type(calldata_cell) is KApply and calldata_cell.label.name == '_+Bytes__BYTES-HOOKED_Bytes_Bytes_Bytes'
+            ):
+                first_bytes = flatten_label(label='_+Bytes__BYTES-HOOKED_Bytes_Bytes_Bytes', kast=calldata_cell)[0]
+                if type(first_bytes) is KToken:
+                    selector_bytes = ast.literal_eval(first_bytes.token)
```
PetarMax, line 1309: I think that we need the first four bytes only? It is possible for the first element of symbolic calldata`+Bytes` to be longer than 4 bytes, think of a function where the first parameter passed in is concrete and the second id symbolic.

nwatson22: Fixed.

PetarMax:
> Do we want `--show-contract-names` turned on by default, and turn it off with `--no-show-contract-names` or `--show-full-bytecode`?

I think so, yes. How would this work with a contract that has immutables, whose deployed bytecode depends on what comes out of the constructor? Is this orthogonal?

(Semi-)Separately, could we turn the jumpdests off completely, please? :)

nwatson22:
> Do we want `--show-contract-names` turned on by default, and turn it off with `--no-show-contract-names` or `--show-full-bytecode`?

@ehildenb @PetarMax I have no strong opinion either way, but I leaned toward off by default because showing the contract name could be misleading since it's modifying the actual configuration cell contents for purely cosmetic purposes right before printing in a way that doesn't make sense given the kontrol semantics.

nwatson22:
> I think so, yes. How would this work with a contract that has immutables, whose deployed bytecode depends on what comes out of the constructor? Is this orthogonal?

@PetarMax If there are immutables in the program, we grab the immutables range information from the compiled json file and compare modulo that. If the program in the configuration is not fully concrete, we don't attempt a lookup and instead just leave the cell as is. Theoretically we could do something more advanced to cover the case where the program is fully concrete excepting variables in the immutable ranges, but this isn't handled now.

> (Semi-)Separately, could we turn the jumpdests off completely, please? :)

So it sounds like we want contract names susbtituted by default, jumpdests hidden by default, and then two options `--show-full-bytecode` and `--show-jumpdests`?

PetarMax:
I am not sure how fine-grained we'd like the showing options to be. It could be that we just want the debugging-friendly one (with program and jumpdests hidden) and then the full one. @ehildenb?

ehildenb:
I think we want the most compact/friendly view by default, and then way to expand it with options. Let's move forward! I'm not a Kontrol user, so I think that Petar's opinion here is more imporntant.

ehildenb (approved):
Approving to unblock. Please make the default view the nicest one.

PR #724, opened 2024-07-26
Fix code lookup not accounting for linkReference placeholder

nwatson22:
Fixes a crash caused by there being non hexadecimal characters in the form of placeholders for linked library addresses in the `deployedBytecode`. When handling this bytecode, replaces these placeholders with zeroes, and then also tracks the `linkReferences` as emitted by the solc compiler along with the bytecode to remove these ranges from the comparison just as the immutables are removed.

Also hides the `interimStates` cell if `--expand-config` is not passed, as is already done for the `jumpDests` cell.


src/kontrol/solc_to_k.py
```diff
@@ -776,7 +777,19 @@ def __init__(self, contract_name: str, contract_json: dict, foundry: bool = Fals
...
+                for rng_grp in ref.values():
+                    for rng in rng_grp:
+                        self.link_ranges.append((rng['start'], rng['length']))
+
         self.deployed_bytecode = deployed_bytecode['object'].replace('0x', '')
+        try:
+            bytearray.fromhex(self.deployed_bytecode)
+        except ValueError:
+            print(f'non hex value for deployed_bytecode found in {self.name_with_path}')
+            print(self.deployed_bytecode)
```
anvacaru, line 792: Just a small comment here. Perhaps it would be better to use `_LOGGER.error` instead of `print` for consistency?

src/kontrol/solc_to_k.py
```diff
@@ -776,7 +777,19 @@ def __init__(self, contract_name: str, contract_json: dict, foundry: bool = Fals
                 for rng in ref:
                     self.immutable_ranges.append((rng['start'], rng['length']))
 
+        self.link_ranges = []
+        if 'linkReferences' in deployed_bytecode:
+            for ref in deployed_bytecode['linkReferences'].values():
+                for rng_grp in ref.values():
+                    for rng in rng_grp:
+                        self.link_ranges.append((rng['start'], rng['length']))
```
anvacaru, line 785: Another suggestion would be to use list comprehension instead of nested loops for better readability. Maybe something in the lines of:
```suggestion
        self.link_ranges = [
            (rng['start'], rng['length'])
            for ref in deployed_bytecode.get('linkReferences', {}).values()
            for rng_grp in ref.values()
            for rng in rng_grp
        ]
```

src/kontrol/foundry.py
```diff
@@ -268,12 +268,19 @@ def method_digest(self, contract_name: str, method_sig: str) -> str:
     def contract_name_from_bytecode(self, bytecode: bytes) -> str:
         for contract_name, contract_obj in self.contracts.items():
             zeroed_bytecode = bytearray(bytecode)
-            for start, length in contract_obj.immutable_ranges:
+            deployed_bytecode_str = re.sub(
+                pattern='__\\$(.){34}\\$__',
+                repl='0000000000000000000000000000000000000000',
+                string=contract_obj.deployed_bytecode,
+            )
+            deployed_bytecode = bytearray.fromhex(deployed_bytecode_str)
```
anvacaru, line 276: It would be nice to have a unit test for this function in [src/tests/unit/test_solc_to_k.py](https://github.com/runtimeverification/kontrol/blob/master/src/tests/unit/test_solc_to_k.py)

src/kontrol/solc_to_k.py
PetarMax: Yes, I am getting some prints while testing that probably shouldn't be there.

nwatson22: Fixed.

src/kontrol/solc_to_k.py
nwatson22: Fixed.

src/kontrol/foundry.py
nwatson22: Added.

PetarMax (approved):
LGTM.

PetarMax:
I'm assuming the failures are due to flakiness, retrying CI.

PR #726, opened 2024-07-29
Don't attempt to hex encode contract names

nwatson22:
Fixes a crash that occurs when viewing proof in `kcfg-view` where the code -> contract name replacement was getting applied where it shouldn't. Would also occur when `--use-hex-encoding` was added. Now the  use_hex_encoding parameter which is always used in `kcfg-view` and optional in `show` will override the simplification behavior.

I tried finding a way to test the `kcfg-view` but the built-in testing capabilities of `textual` seem to rely on being executed in an `async` function (see https://textual.textualize.io/guide/testing/) so I don't see a good way to test the rendering unless we want to add new capabilities to our test suite.

I did add a regression test which catches the crash with foundry show with hex encoding enabled.

### DIFF
--- a/src/kontrol/foundry.py
+++ b/src/kontrol/foundry.py
@@ -122,7 +122,7 @@ class FoundryKEVM(KEVM):
                     return KApply(_term.label, [DOTS])
             return _term
 
-        if not self.foundry._expand_config and isinstance(kast, KInner):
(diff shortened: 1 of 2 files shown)


palinatolmach (approved):
Looks good, thanks @nwatson22!

Issue #725, opened 2024-07-29
Use `kontrol version` to determine if the proof is up to date


palinatolmach:
Related: https://github.com/runtimeverification/kontrol/issues/601

As @PetarMax identified, the proof now restarts after adding lemmas and running `kontrol build --rekompile --regen`, which didn't happen before, and which is not desirable. To address this, we can probably
- record `kontrol version` in the digest file, which we'd like to have for KaaS anyway
- incorporate the check on Kontrol version when determining if the proof is up to date: for example, we can _not_ re-initialize the proof if the Foundry hash is up to date, Kontrol version is the same, but the K kompilation hash has changed. At the same time, if Kontrol version changed, we might want to reinitialize the proof. 

Also, https://github.com/runtimeverification/kontrol/pull/467 added the ability to specify the version of `setUp` function to be used via `--setup-version`. Similarly, we can probably specify the version of a proof itself, which would help, e.g., to complete the proof even if it's sources got outdated. That'll require adding the ability to specify and propagate `user_specified_version` that is not only available for `setUp`. Might be a separate PR.

PR #731, opened 2024-07-30
Refactor digest checks.

anvacaru:
- Modified `update_kompilation_digest` to store the kontrol version in the digest file.
- Added a new `kontrol_up_to_date` function that checks if the current version is the same as the one in the digest file.
- Added a new `should_rekompile` function in `foundry_kompile` that: 
     - checks if there's an explicit request to recompile.
     - checks if the compilation timestamp file does not exist.
     - checks if any of the components (definitions, contracts, or tool) are not up to date.
- refactored all the `*up_to_date` and `update_digest` methods and extracted the common logic in a single  `_read_digest_file` function.
- Modified `resolve_proof_version` and `resolve_setup_proof_version` to check the kontrol verison using `kontrol_up_to_date`, emitting a warning if it is out of date.


F-WRunTime:
Will the `kontrol_up_to_date` be a hard check for enforcing the version of kontrol defined in toml is what is used for the execution process? 

Will there be an option to make it a soft enforcement and just provides a warning instead of an error?

F-WRunTime:
The above question may also be getting addressed in #725 ?

anvacaru:
@F-WRunTime currently it isn't set to enforce the version of kontrol. For `kontrol build`, it will trigger a rekompilation. For `prove` it will only emit a warning, because the tool version might contain changes to the semantics and so the definition might become outdated.

palinatolmach (approved):
Looks really good to me, thanks @anvacaru!

I opened an issue about documenting the rekompilation/reinitialization logic for us to address whenever someone has time (https://github.com/runtimeverification/gitbook-kontrol/issues/54).

PR #735, opened 2024-08-01
Add `--remove-old-proofs` command

JuanCoRo:
This PR adds the `kontrol prove` flag `--remove-old-proofs` that will check if the digest of the contracts is up to date, and if it's not, it will remove the `out/proofs` directory.

This is important in the context of KaaS, since we don't want to be producing new versions for the proofs.


src/kontrol/cli.py
```diff
@@ -524,6 +524,13 @@ def parse(s: str) -> list[T]:
         action='store_false',
         help='If tracing is active, avoid storing memory information.',
     )
+    prove_args.add_argument(
+        '--remove-old-proofs',
+        dest='remove_old_proofs',
+        default=None,
+        action='store_true',
+        help='Remove all existing KCFGs if they need to be regenerated.',
```
anvacaru, line 532: Hmm, maybe `remove all outdated KCFGs`?

src/kontrol/foundry.py
```diff
@@ -722,6 +725,18 @@ def free_proof_version(
...
+    def remove_old_proofs(self, force_remove: bool = False) -> bool:
+        if force_remove or any(
+                not method.contract_up_to_date(Path(method.contract_digest))
+                for contract in self.contracts.values()
+                for method in contract.methods
+            ):
+                shutil.rmtree(self.proofs_dir.absolute())
+                return True
+        else:
+            return False
```
anvacaru, line 737: Wouldn't it make more sense for this function to delete only the outdated proof dirs?

Also, I'd say it is redundant to check the `contract_up_to_date` on each method of a contract. You could either:
1. move the `contract_up_to_date` and `contract_digest` methods in the `Contract` class instead of the `Method` and then just do
```python
not contract.contract_up_to_date(Path(contract_digest))
for contract in self.contracts.values()
```
2. (might be easier to implement now) Check `contract_up_to_date` only for the first function of a contract.
```py
  not contract.methods[0].contract_up_to_date(Path(contract.methods[0].contract_digest))
for contract in self.contracts.values()
```

JuanCoRo: Thank you for the recommendation!

Isn't the digest based on all functions of the contract? Namely, if a function changes or a new one is added, the digest changes? Asking to know how exactly would targeting the individual outdated proof dirs work!

JuanCoRo: A problem with option 2 might be that if a contract doesn't have any methods it'll throw an error when trying to access `contract.methods[0]` I think

anvacaru: > Isn't the digest based on all functions of the contract? Namely, if a function changes or a new one is added, the digest changes?

That's my understanding as well. In addition, there is a `method` digest, and a `contract` digest. AFAIK, the `contract` digest is the same for any method of that contract.

edit:
here's a digest example
```json
{
    "methods": {
        "test%CounterTest.test_Increment()": {
            "method": "4c32041fa5a7c64776f36dab5ae8cc61588efc3e3fb09b3c95ed0917587687d6",
            "contract": "29c961bb216d9a5f16961d6a224fd6cb1feeb18b36e29e9a09c2733b0590a655"
        },
        "test%CounterTest.setUp()": {
            "method": "2c1737186c0be7909e9db344314dabfa10f5f492345ad17e87375ed13c1bcaa6",
            "contract": "29c961bb216d9a5f16961d6a224fd6cb1feeb18b36e29e9a09c2733b0590a655"
        },
        "test%CounterTest.testFuzz_SetNumber(uint256)": {
            "method": "323d5de4c923be8a393316579d3833bd4f00d4d81a5a85b3030a54efd2222e87",
            "contract": "29c961bb216d9a5f16961d6a224fd6cb1feeb18b36e29e9a09c2733b0590a655"
        }
    },
    "kompilation": "089cb7a3c95b4643e44147c41139b379b7a69797b3a8cb3d6a9e4acc41fb3bef",
    "kontrol": "0.1.32",
    "foundry": "a7794d1db53643ecf7e89e8be01ca0fcad6ca2cf381e68f4fdbe70d1be6934f9"
}
```
For any method of `CounterDigest` the `contract` hash is the same. So I don't see a point to loop through each function of the contract.

JuanCoRo: According to [the docs](https://docs.python.org/3/library/functions.html#any), `any` (and `all`, for that matter) evaluate lazily. So we wouldn't be computing all of the `contract_up_to_date` and then seeing if any of them are not up to date. It'd stop at the first positive.
In light of this, my previous comment, and that factoring out the contract digest logic down to the `Contract` class seems to deserve a PR of its own, I'd suggest leaving the code as is. What do you think?

src/kontrol/cli.py
JuanCoRo: Done!

src/kontrol/foundry.py
anvacaru: Yes, sounds good. We should open an issue for this!

JuanCoRo: Yep!! On it!

anvacaru (approved):

PR #755, opened 2024-08-07
Add proof cleaning capabilities to `kontrol clean`

JuanCoRo:
This PR adds the flags `--proofs` and `--old-proofs` to `kontrol clean` to remove the `out/proofs` directory always or when the proofs are outdated (respectively).

It also fixes a bug in the `Foundry.remove_old_proofs` function (credits to @anvacaru :tada:)


src/kontrol/foundry.py
```diff
@@ -736,7 +737,10 @@ def free_proof_version(
 
     def remove_old_proofs(self, force_remove: bool = False) -> bool:
         if force_remove or any(
-            not method.contract_up_to_date(Path(method.contract_digest))
+            # We need to check only the methods that get written to the digest file
+            # Otherwise we'd get vacuous positives
+            (method.is_test or method.is_testfail or method.is_setup)
```
palinatolmach, line 742: How about it being a constructor? Do they get written to the digest too? They are called `contractName.init`, and you need `--run-constructor` to produce it.

src/kontrol/foundry.py
```diff
@@ -1382,3 +1390,14 @@ def init_project(project_root: Path, *, skip_forge: bool) -> None:
         logger=_LOGGER,
         cwd=root,
     )
+
+
+def foundry_clean(foundry: Foundry, options: CleanOptions) -> None:
+    if options.proofs and options.old_proofs:
+        raise AttributeError('Use --proofs or --old-proofs, but not both!')
```
palinatolmach, line 1397: Does `options.proof` subsume `options.old_proofs`? If yes, maybe we can just use it if both are provided?

palinatolmach (approved):
lgtm, left two minor comments!

src/kontrol/foundry.py
JuanCoRo: I don't think it subsumes it, since  `options.proof` means "remove `proofs` unconditionally" and `options.old_proofs` means "remove `proofs` conditionally". But maybe there's another way of reading this?

src/kontrol/foundry.py
JuanCoRo: Funny middle ground! The constructor gets written **but** without a `contract` digest!
```json
{                                                                                                                                 [0/80]
    "methods": {                                                                                                                        
        "test%kontrol%proofs%CounterKontrol.prove_multiple_counters()": {                                                               
            "method": "cfb5371552b12281a0df8d5a68b32fe96b588f698cb23202bde26567304c7bf2",                                               
            "contract": "ccb6162f5aef84316ecc13c735a21873b460b273951ad139b2cd0dc60b3458f7"                                              
        },                                                                                                                              
        "test%kontrol%proofs%CounterKontrol.setUp()": {                                                                                 
            "method": "726a86ee444d9b687ff4d0f01e3279ac44e5c328f0796241fa9f7088fba75554",                                               
            "contract": "ccb6162f5aef84316ecc13c735a21873b460b273951ad139b2cd0dc60b3458f7"                                              
        },                                                                                                                              
        "CounterKontrol.init": {                                                                                                        
            "method": "109c35c430033ee2070cf2bc311cbbe0f294de7e6a1f5e27744f0103e574e325"                                                
        }                                                                                                                               
    },                                                                                                                                  
    "kompilation": "0c8253ebff1f9b32a476e80c4d5981fc66c9a16a6618b11710448a3b460e6078",                                                  
    "kontrol": "0.1.395",                                                                                                               
    "foundry": "d86f9cb86f4e182bd30d2eb1d1323ce62aa1c5b62ee0c91445a173007e7651b5",                                                      
    "build-options": "BuildOptions(target: KompileTarget.HASKELL, foundry_root: ., enum_constraints: False, emit_json: True, llvm_kompi\
le: True, llvm_library: False, enable_llvm_debug: False, llvm_kompile_type: None, llvm_kompile_output: None, llvm_proof_hint_instrument\
ation: False, llvm_proof_hint_debugging: False, read_only: False, o0: False, o1: False, o2: False, o3: False, ccopts: [], enable_search\
: False, coverage: False, gen_bison_parser: False, gen_glr_bison_parser: False, bison_lists: False, no_exc_wrap: False, outer_parsed_js\
on: False, ignore_warnings: [], requires: [], imports: [], spec_module: None, main_module: None, syntax_module: None, md_selector: k, i\
ncludes: [], definition_dir: None, depth: None, verbose: False, debug: False, regen: False, rekompile: False, no_forge_build: False, no\
_silence_warnings: False, command: build, config_file: kontrol.toml, config_profile: default)"                                          
}  
```
So, since we're looking a the contract digest it seems fine to leave it as is. Let me know if you agree!

src/kontrol/foundry.py
palinatolmach: Ok, that’s fair, let’s leave it like this!

src/kontrol/foundry.py
palinatolmach: I agree! But maybe I’d also suggest to address it altogether in the next PR, otherwise we might end up with constructor proofs accumulating in KaaS or something.

Thanks for checking!

JuanCoRo: That's right! I'd say it's a corner case that we only have constructors and not proofs, but if that's the case, the proofs won't be erased!

PR #757, opened 2024-08-08
Add `require "foundry.md"` to `kontrol init` generated lemmas file

JuanCoRo:
The `lemmas.k` file generated by `kontrol init` didn't include a `require "foundry.md"`, which needs to be imported in order to add any relevant lemmas. This PR addresses it.


palinatolmach (approved):

src/kontrol/utils.py
```diff
@@ -72,7 +72,9 @@ def append_to_file(file_path: Path, content: str) -> None:
 
 
 def empty_lemmas_file_contents() -> str:
-    return """module KONTROL-LEMMAS
+    return """require "foundry.md"
```
anvacaru, line 75: This should be a `requires`.

Issue #525, opened 2024-04-22
Tests on CI are run against obsolete Foundry version 


tothtamas28:
The test Docker image uses a certain version of Foundry as a layer:

https://github.com/runtimeverification/kontrol/blob/331e6d060627fe75969c566082d343274eeb7b4b/.github/workflows/Dockerfile#L6

This corresponds to a given commit: https://github.com/foundry-rs/foundry/commit/7545c7a2857a873fa1909ec4174032c4e4702116.

At some point, this version was listed under releases as well, but not anymore (Foundry removes releases for some reason from time to time).

In theory, it should be possible to install a certain commit with `foundryup -C` (or build from source), but it'd be better for usability to test against a recent enough version of Foundry that's distributed as a release.

Issue #534, opened 2024-04-25
Slowdown detected for new versions of forge-std


anvacaru:
Our current test-suite uses a hardcoded version of `forge-std` [linked here](https://github.com/foundry-rs/forge-std/tree/75f1746c949ae56377611e5f2128aa93d381431f), tagged by the hash `75f1746`.
After updating to the latest version, it seems that there is a ~72% slowdown.

To check this, I've used the project that gets created with `kontrol init`, but I've tweaked the test below to use `assert` instead of `assertEq`.

```solidity
    function test_Increment() public {
        counter.increment();
        assert(counter.number() == 1);
    }
```

Using the `master` branch of kontrol (v0.1.256), I've ran the test with both `forge-std` libraries.
forge-std version: "1.8.1": 55s
forge-std version: "1.6.0": 32s

I have the bug-report tar for the proof that has the slowdown ([we-transfer link here as the file has 155 MB](https://we.tl/t-ET4K1scB77))

anvacaru:
I investigated a bit the `CounterTest` example using the "old" `v1.6.0` version of `forge-std` that we are using in our test project and the latest one, `v1.9.2`:
- The build time increased by `30s` with `v1.9.2`
- The execution time increased with `30s` as well, but the time it takes to run the proof does not change that much:
     - with `v1.6.0`:
       - kore server start takes 12 seconds
       - initializing the proof takes 6 seconds
       - running the proof takes 11 seconds for a total of 29 seconds
    - with `v1.9.2`:
       - kore server start takes 38 seconds
       - initializing the proof takes 7 seconds
       - running the proof takes 12 seconds for a total of 57 seconds 
 - With the new version, there are only 11 additional steps. This does not indicate a general slowdown.

In addition to this, I ran a `setUp` and a proof from the ongoing engagement and the times are:
- with `v1.7.4` - default on the project: 
   -  setUp `6m18s`  
   - proof `15m4s`
- with `v1.9.2`:
    - setUp `6m51s`
    - proof `16m8s`
 
At this point, I think this is just an overhead caused by the codebase changes in `forge-std.`
I'm not sure why the kore server start takes longer; one educated guess is that it is because of the library's larger codebase. 


The Solidity files that are part of the lib are
```shell
./src/interfaces/IERC4626.sol
./src/interfaces/IERC165.sol
./src/interfaces/IERC721.sol
./src/interfaces/IERC1155.sol
./src/interfaces/IERC20.sol
[...]

PR #426, opened 2024-03-08
Stabilising produced EVM code

PetarMax:
This PR stabilises the EVM bytecode produced by the compiler by not including the IPFS hash and the CBOR metadata.

This is relevant, for example, for summary reusability in CSE.

PR #763, opened 2024-08-09
Add `--no-metadata` to `kontrol build`, bump Foundry version in CI

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/525.

Follow up to https://github.com/runtimeverification/kontrol/pull/426, which stabilizes the EVM bytecode produced by the compiler by not including the IPFS hash and the CBOR metadata.

~Blocked on https://github.com/runtimeverification/kontrol/issues/525 and https://github.com/runtimeverification/kontrol/issues/534, since most of our tests are run against an obsolete version of Foundry, so it has to be bumped too. But it means that we'll need to update our tests to use [native assertions](https://github.com/foundry-rs/forge-std/pull/514), and I think upgrading Foundry also made our tests slower as identified by @anvacaru.~ Update: the slowdown is associated with a newer version of `forge-std`, not Foundry.


anvacaru:
I don't think we're blocked on #534, as that's related to the `forge-std` lib we're fetching using the commit hash.

anvacaru (approved):

PR #774, opened 2024-08-14
Add new Kontrol CLI args

anvacaru:
- adding the `--use-hex-encoding` flag, also used in `show` to the `view-kcfg`.
- updates to the `kontrol.toml` file created during `kontrol init`.
- adding a new `--optimize-performance M` flag to Kontrol that sets:
```sh
  --assume-defined 
    (still a bit experimental, uses Booster implies endpoint first, 
     then if that doesn't work, goes to Kore, may spit out error-like messages)
  --no-log-rewrite 
    (does not log any rewrite rules, this is ok when we don't care about which 
     rules we've applied (and we mostly don't))
  --max-frontier-parallel M 
    (for me it's 6, depends on machine, appropriate env variable needs to be set 
     and compilation should be done without that variable set)
  --maintenance rate N 
    (we can figure out what is the correct ratio, I'm using 2*M, but now that the 
     execution is flying it should be larger, writes to disk slow us down)
  --smt-timeout 32000 
    (to be understood, this seems to work for Lido)
  --smt-retry-limit 0
    (no retrying, we need to set the limit high enough)
  --max-depth 100000 and --max-iterations 10000
    (should be enough)
```


palinatolmach (dismissed):
lgtm!

src/kontrol/cli.py
```diff
@@ -552,6 +552,18 @@ def parse(s: str) -> list[T]:
...
+    prove_args.add_argument(
+        '--optimize-performance',
+        type=int,
+        default=None,
+        dest='optimize_performance',
+        help=(
+            'Optimize performance settings for the proof execution. Takes an integer value that adjusts '
+            'parameters like frontier parallelism and maintenance rate. Will overwrite any settings on:'
+            "'assume-defined', 'log-success-rewrites', 'max-frontier-parallel', 'maintenance-rate', 'smt-timeout'"
+            ", 'smt-retry-limit', 'max-depth' and 'max-iterations'. Example: '--optimize-performance 6'"
```
PetarMax, line 564:
```suggestion
            'Optimize performance for proof execution. Takes the number of parallel threads to be used.'
            "Will overwrite other settings of 'assume-defined', 'log-success-rewrites', 'max-frontier-parallel',"
            "'maintenance-rate', 'smt-timeout', 'smt-retry-limit', 'max-depth' and 'max-iterations'."
```

src/kontrol/options.py
```diff
@@ -434,6 +440,24 @@ def get_argument_type() -> dict[str, Callable]:
...
+        If `optimize_performance` is not `None`, this method will adjust several
+        internal parameters to enhance performance. The integer value of `optimize_performance`
+        is used to set the level of parallelism for frontier exploration and maintenance rate.
+        """
+        if self.optimize_performance is not None:
+            self.assume_defined = True
+            self.log_succ_rewrites = True
+            self.max_frontier_parallel = self.optimize_performance
+            self.maintenance_rate = 2 * self.optimize_performance
+            self.smt_timeout = 32000
```
PetarMax, line 456:
```suggestion
            self.smt_timeout = 120000
```

src/kontrol/options.py
```diff
@@ -434,6 +440,24 @@ def get_argument_type() -> dict[str, Callable]:
...
+        """Applies a series of performance optimizations based on the value of the
+        `optimize_performance` attribute.
+
+        If `optimize_performance` is not `None`, this method will adjust several
+        internal parameters to enhance performance. The integer value of `optimize_performance`
+        is used to set the level of parallelism for frontier exploration and maintenance rate.
+        """
+        if self.optimize_performance is not None:
+            self.assume_defined = True
+            self.log_succ_rewrites = True
```
PetarMax, line 453:
```suggestion
            self.log_succ_rewrites = False
```

src/kontrol/options.py
```diff
@@ -434,6 +440,24 @@ def get_argument_type() -> dict[str, Callable]:
...
+    def apply_optimizations(self) -> None:
+        """Applies a series of performance optimizations based on the value of the
+        `optimize_performance` attribute.
+
+        If `optimize_performance` is not `None`, this method will adjust several
+        internal parameters to enhance performance. The integer value of `optimize_performance`
+        is used to set the level of parallelism for frontier exploration and maintenance rate.
+        """
+        if self.optimize_performance is not None:
+            self.assume_defined = True
```
palinatolmach, line 452: Do we want to/can add `--kore-rpc-command='kore-rpc-booster --no-post-exec-simplify '` too?

palinatolmach: And `--smt-arg smt.ematching=false`

src/kontrol/cli.py
```diff
@@ -552,6 +552,17 @@ def parse(s: str) -> list[T]:
...
     )
+    prove_args.add_argument(
+        '--optimize-performance',
+        type=int,
+        default=None,
+        dest='optimize_performance',
+        help=(
+            'Optimize performance for proof execution. Takes the number of parallel threads to be used.'
+            "Will overwrite other settings of 'assume-defined', 'log-success-rewrites', 'max-frontier-parallel',"
+            "'maintenance-rate', 'smt-timeout', 'smt-retry-limit', 'max-depth' and 'max-iterations'."
```
palinatolmach, line 563:
```suggestion
            "'maintenance-rate', 'smt-timeout', 'smt-retry-limit', 'max-depth', and 'max-iterations'."
```

palinatolmach (approved):

src/kontrol/options.py
```diff
@@ -434,6 +440,24 @@ def get_argument_type() -> dict[str, Callable]:
...
+            self.log_succ_rewrites = False
+            self.max_frontier_parallel = self.optimize_performance
+            self.maintenance_rate = 2 * self.optimize_performance
+            self.smt_timeout = 120000
```
palinatolmach, line 456: Is it not 32000 (as in @PetarMax's message) intentionally? :)

src/kontrol/options.py
palinatolmach: Makes sense, thanks!

src/kontrol/options.py
anvacaru: @PetarMax  suggested the change to `120000` in a comment above.

src/kontrol/options.py
```diff
@@ -434,6 +440,24 @@ def get_argument_type() -> dict[str, Callable]:
...
+        internal parameters to enhance performance. The integer value of `optimize_performance`
+        is used to set the level of parallelism for frontier exploration and maintenance rate.
+        """
+        if self.optimize_performance is not None:
+            self.assume_defined = True
+            self.log_succ_rewrites = True
+            self.max_frontier_parallel = self.optimize_performance
+            self.maintenance_rate = 2 * self.optimize_performance
+            self.smt_timeout = 32000
+            self.smt_retry_limit = 0
```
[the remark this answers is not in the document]
PetarMax: The idea is to have a sufficiently high max timeout and that's it - any further retries would just increase that timeout and we would lose the time in the first check in the end. There is no difference in performance, strategies, or knowledge learnt transferred between checks.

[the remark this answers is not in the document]
JuanCoRo: Makes sense, thank you!

PR #779, opened 2024-08-16
Include keccak & aux lemmas conditionally

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/715.
Partially addresses https://github.com/runtimeverification/kontrol/issues/780.

This PR
- adds `keccak.md` file containing (unsound) assumptions about keccak to facilitate reasoning, use it by default
- adds `kontrol_lemmas.md` file containing other Kontrol lemmas that facilitate reasoning (copied from [Morpho lemmas](https://github.com/runtimeverification/_kaas_morpho/blob/c11b20e50ba694e9828fb6034bcbfca297aee278/test/kontrol/morpho-lemmas.k) with the exclusion of keccak assumptions included in `keccak.md`)
- includes both during `kontrol build` by default
- adds `--no-keccak-lemmas`, `--auxiliary-lemmas` options to `kontrol build` which excludes and includes keccak and auxiliary lemmas, respectively
- fixes a typo in an error message (`"provided"`)

PR #785, opened 2024-08-23
Fix permissions on `keccak.md`

palinatolmach:
Fixes an issue introduced in https://github.com/runtimeverification/kontrol/pull/779 by making `out/kompiled/requires/keccak.md` (and any other filed copied to `requires` in future) writeable for the current user, group, and others (anyone with access to the file).

The files have to writeable so they can be deleted and replaced when `kontrol build --rekompile` is executed. The files not copied from `kontrol/kdist` had write access previously, but the ones we are now copying from `kdist` (such as `keccak.md`) don't, which  causes issues such as this one:

```
INFO 2024-08-23 08:58:19,927 kontrol.kompile - Copying requires path: /nix/store/g67pca4v4lx9zgna190kdwvrj4n0jarm-python3.10-kontrol-0.1.409/lib/python3.10/site-packages/kontrol/kdist/keccak.md ->                           
out/kompiled/requires/keccak.md                                                                                                                                                                                                
An error occurred while building your Kontrol project: [Errno 13] Permission denied: 'out/kompiled/requires/keccak.md'
```

### DIFF
--- a/src/kontrol/kompile.py
+++ b/src/kontrol/kompile.py
@@ -81,7 +81,7 @@ def foundry_kompile(
                 # Fetch current permissions
                 current_permissions = Path.stat(req_path).st_mode
                 # Grant write permissions
-                Path.chmod(req_path, current_permissions | Path.stat.S_IWUSR | Path.stat.S_IWGRP | Path.stat.S_IWOTH)
+                req_path.chmod(current_permissions | Path.stat.S_IWUSR | Path.stat.S_IWGRP | Path.stat.S_IWOTH)
             regen = True
 
     _imports: dict[str, list[str]] = {contract.name_with_path: [] for contract in foundry.contracts.values()}
--- a/src/kontrol/kompile.py
+++ b/src/kontrol/kompile.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import shutil
+import stat
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -75,6 +77,12 @@ def foundry_kompile(
         if regen or not req_path.exists():
             _LOGGER.info(f'Copying requires path: {req} -> {req_path}')
             shutil.copy(req, req_path)
+            # If the copied file is not writeable
+            if not os.access(req_path, os.W_OK):
+                # Fetch current permissions
+                current_permissions = os.stat(req_path).st_mode
+                # Grant write permissions
+                os.chmod(req_path, current_permissions | stat.S_IWUSR | stat.S_IWGRP | stat.S_IWOTH)
             regen = True
 
     _imports: dict[str, list[str]] = {contract.name_with_path: [] for contract in foundry.contracts.values()}
--- a/src/kontrol/kompile.py
+++ b/src/kontrol/kompile.py
@@ -79,7 +79,7 @@ def foundry_kompile(
             # If the copied file is not writeable
             if not os.access(req_path, os.W_OK):
                 # Fetch current permissions
-                current_permissions = Path.stat(req_path).st_mode
+                current_permissions = req_path.stat().st_mode
                 # Grant write permissions
                 req_path.chmod(current_permissions | Path.stat.S_IWUSR | Path.stat.S_IWGRP | Path.stat.S_IWOTH)
             regen = True
--- a/src/kontrol/kompile.py
+++ b/src/kontrol/kompile.py
@@ -4,7 +4,6 @@ import json
 import logging
 import os
 import shutil
-import stat
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -80,9 +79,9 @@ def foundry_kompile(
             # If the copied file is not writeable
             if not os.access(req_path, os.W_OK):
                 # Fetch current permissions
-                current_permissions = os.stat(req_path).st_mode
+                current_permissions = Path.stat(req_path).st_mode
                 # Grant write permissions
-                os.chmod(req_path, current_permissions | stat.S_IWUSR | stat.S_IWGRP | stat.S_IWOTH)
+                Path.chmod(req_path, current_permissions | Path.stat.S_IWUSR | Path.stat.S_IWGRP | Path.stat.S_IWOTH)
             regen = True
 
     _imports: dict[str, list[str]] = {contract.name_with_path: [] for contract in foundry.contracts.values()}
--- a/src/kontrol/kompile.py
+++ b/src/kontrol/kompile.py
@@ -4,6 +4,7 @@ import json
 import logging
 import os
 import shutil
+import stat
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


src/kontrol/kompile.py
```diff
@@ -75,6 +77,12 @@ def foundry_kompile(
         if regen or not req_path.exists():
             _LOGGER.info(f'Copying requires path: {req} -> {req_path}')
             shutil.copy(req, req_path)
+            # If the copied file is not writeable
+            if not os.access(req_path, os.W_OK):
+                # Fetch current permissions
+                current_permissions = os.stat(req_path).st_mode
+                # Grant write permissions
+                os.chmod(req_path, current_permissions | stat.S_IWUSR | stat.S_IWGRP | stat.S_IWOTH)
```
tothtamas28, line 85: Why is it necessary to have write access on these requires?

palinatolmach: We want to be able to delete and replace them when `kontrol build --rekompile` is executed. The files not copied from `kontrol/kdist` had write access previously, but the we are not copying from `kdist` (such as `keccak.md`) don't, and it causes issues such as this one:
```
INFO 2024-08-23 08:58:19,927 kontrol.kompile - Copying requires path: /nix/store/g67pca4v4lx9zgna190kdwvrj4n0jarm-python3.10-kontrol-0.1.409/lib/python3.10/site-packages/kontrol/kdist/keccak.md ->                           
out/kompiled/requires/keccak.md                                                                                                                                                                                                
An error occurred while building your Kontrol project: [Errno 13] Permission denied: 'out/kompiled/requires/keccak.md'
```

tothtamas28: > it causes issues such as this one

Please include the issue in the PR description.

tothtamas28, line 85: Consider using `Path.chmod`.

src/kontrol/kompile.py
```diff
@@ -75,6 +77,12 @@ def foundry_kompile(
         if regen or not req_path.exists():
             _LOGGER.info(f'Copying requires path: {req} -> {req_path}')
             shutil.copy(req, req_path)
+            # If the copied file is not writeable
+            if not os.access(req_path, os.W_OK):
+                # Fetch current permissions
+                current_permissions = os.stat(req_path).st_mode
```
tothtamas28, line 83: Consider using `Path.stat`.

tothtamas28 (approved):

src/kontrol/kompile.py
palinatolmach: Done, thank you!

tothtamas28:
Btw, is it possible the root of the issue is that these files have not been checked in to source control with the right permission?

src/kontrol/kompile.py
palinatolmach: Done in https://github.com/runtimeverification/kontrol/pull/785/commits/597820377463129484ee18bd5a52df6855957d09, thanks!

src/kontrol/kompile.py
palinatolmach: Done in https://github.com/runtimeverification/kontrol/pull/785/commits/597820377463129484ee18bd5a52df6855957d09, thanks!

src/kontrol/kompile.py
```diff
@@ -75,6 +76,12 @@ def foundry_kompile(
         if regen or not req_path.exists():
             _LOGGER.info(f'Copying requires path: {req} -> {req_path}')
             shutil.copy(req, req_path)
+            # If the copied file is not writeable
+            if not os.access(req_path, os.W_OK):
+                # Fetch current permissions
+                current_permissions = Path.stat(req_path).st_mode
+                # Grant write permissions
+                Path.chmod(req_path, current_permissions | Path.stat.S_IWUSR | Path.stat.S_IWGRP | Path.stat.S_IWOTH)
```
tothtamas28, line 84:
```suggestion
                req_path.chmod(current_permissions | Path.stat.S_IWUSR | Path.stat.S_IWGRP | Path.stat.S_IWOTH)
```

src/kontrol/kompile.py
```diff
@@ -75,6 +76,12 @@ def foundry_kompile(
         if regen or not req_path.exists():
             _LOGGER.info(f'Copying requires path: {req} -> {req_path}')
             shutil.copy(req, req_path)
+            # If the copied file is not writeable
+            if not os.access(req_path, os.W_OK):
+                # Fetch current permissions
+                current_permissions = Path.stat(req_path).st_mode
```
tothtamas28, line 82:
```suggestion
                current_permissions = req_path.stat().st_mode
```

palinatolmach:
@tothtamas28 it would probably help, but these files are part of the kompilation artifacts usually placed in the `/out` folder in Foundry projects, and it's usually not tracked by source control at all (here's [one example](https://github.com/morpho-org/morpho-blue/blob/8e35224dfc69eebf68ad67408bb1154173aeed16/.gitignore#L3) from a project we're looking at).

src/kontrol/kompile.py
palinatolmach: Oh, sorry, I misunderstood how it should be used. Thank you!

tothtamas28:
> these files are part of the kompilation artifacts usually placed in the `/out` folder in Foundry projects

What I'm wondering, if the template of these files (i.e. in the `KSRC_DIR`) have the correct permissions, and [`shutil,copy`](https://docs.python.org/3/library/shutil.html#shutil.copy) preserves file permissions, why do the files placed in `out` have the wrong permissions.

palinatolmach:
> What I'm wondering, if the template of these files (i.e. in the KSRC_DIR) have the correct permissions, and [shutil,copy](https://docs.python.org/3/library/shutil.html#shutil.copy) preserves file permissions, why do the files placed in out have the wrong permissions.

Ah, got it, thanks, it looks like they don't.

If `kontrol build` is run from a local checkout (`poetry run kontrol build`), then these files are copied with write permissions as they are in `src/kontrol/kdist`:
```
❯ ls -l src/kontrol/kdist
total 304
-rw-r--r--  1 palina  staff      0 Apr  9 20:53 __init__.py
drwxr-xr-x  5 palina  staff    160 Apr 11 16:36 __pycache__
-rw-r--r--  1 palina  staff  19147 Aug  9 15:15 assert.md
...
-rw-r--r--@ 1 palina  staff   3218 Aug 22 22:15 keccak.md
```

If we run `kontrol build` using a `kup` installation, then they're copied from something like `/nix/store/gyx8hr1kq30chx12bdziqv5yyz582ql3-python3.10-kontrol-0.1.409/lib/python3.10/site-packages/kontrol/kdist/`, and the `kdist` files only have read permissions there:
```
> ls -l /nix/store/gyx8hr1kq30chx12bdziqv5yyz582ql3-python3.10-kontrol-0.1.409/lib/python3.10/site-packages/kontrol/kdist/
-r--r--r--@ 1 root  wheel      0 Jan  1  1970 __init__.py
dr-xr-xr-x@ 8 root  wheel    256 Jan  1  1970 __pycache__
-r--r--r--@ 1 root  wheel  19147 Jan  1  1970 assert.md
...
-r--r--r--@ 1 root  wheel   3218 Jan  1  1970 keccak.md
```

In short, it seems that the `kdist` files distributed with `kup`-installed Kontrol are read-only, and I thought it's intentional.

tothtamas28:
> In short, it seems that the `kdist` files distributed with `kup`-installed Kontrol are read-only, and I thought it's intentional.

Makes sense, thank you for debugging this. The Nix store in general is read only, so your fix is indeed the right thing to do.

palinatolmach:
@tothtamas28 thanks for your help!

anvacaru (approved):

PR #787, opened 2024-08-23
Revising init/pyproject file to show correct Major version on master branch

F-WRunTime:
__init__.py: Revise major version to 1.0.0. to propogate to master branch

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "kontrol"
-version = "0.1.409"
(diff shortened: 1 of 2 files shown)


ehildenb:
I think we also need to update this file to bump this version number appropriately? https://github.com/runtimeverification/kontrol/blob/08f75436b983db017422e8cf5fd0f50991507962/package/version.sh#L32

ehildenb:
I also see that `pyproject.toml` didn't get updated. Maybe that needs to be updated as well if we're manually bumping this?

ehildenb (approved):

PR #801, opened 2024-08-29
Add `--schedule` to `kontrol prove`

palinatolmach:
This PR adds `--schedule`, `--mode`, `--chain-id`, and `--no-gas` options to `kontrol prove` (example: `kontrol prove --schedule CANCUN`) by using `EVMChainOptions` and `evm_chain_args` in Kontrol. To make the gas enabling mechanism in Kontrol compatible with `EVMChainOptions`, this PR also renames `use_gas` argument to `usegas`.


ehildenb:
I thikn, instead, wwe can import EVM Chain args:

- https://github.com/runtimeverification/evm-semantics/blob/96cf7e15bbe4e3481d70736d03c55ba754582111/kevm-pyk/src/kevm_pyk/cli.py#L897
- https://github.com/runtimeverification/evm-semantics/blob/96cf7e15bbe4e3481d70736d03c55ba754582111/kevm-pyk/src/kevm_pyk/cli.py#L442

Those ones use an Enum for the schedule, so already does validity parsing, and translating it into the Enum.

ehildenb:
Basically, we can subcglass `EVMChainOptions` into `ProveOptions`, and then make sure to use all the options three for setting up the initial state (which are all things that can be used: schedule, chainid, mode, and usegas.

src/kontrol/options.py
```diff
@@ -381,7 +390,7 @@ def default() -> dict[str, Any]:
             'reinit': False,
             'bmc_depth': None,
             'run_constructor': False,
-            'use_gas': False,
+            'usegas': False,
             'break_on_cheatcodes': False,
             'recorded_diff_state_path': None,
```
ehildenb, line 395: Since we're inverting the default here, can you double-check that the `--use-gas` option works locally, and that it by default is turned off?

src/kontrol/prove.py
```diff
@@ -948,26 +956,27 @@ def _init_cterm(
...
 
     if not trace_options:
         trace_options = TraceOptions({})
 
     jumpdests = bytesToken(_process_jumpdests(bytecode=program))
     init_subst = {
-        'MODE_CELL': KApply('NORMAL'),
-        'USEGAS_CELL': TRUE if use_gas else FALSE,
+        'MODE_CELL': KApply(evm_chain_options.mode),
+        'USEGAS_CELL': TRUE if evm_chain_options.usegas else FALSE,
```
ehildenb, line 977:
```suggestion
        'USEGAS_CELL': boolToken(evm_chain_options.usegas),
```

ehildenb (approved):
Overall looks good, if we double check that `--use-gas` is behaving correctly, and the default is to not have it turned on, we can merge. A few small suggestions.

src/kontrol/options.py
palinatolmach: Will check, thanks!

palinatolmach:
Thanks @ehildenb! Will check `--use-gas` and whether setting it in `foundry.toml` works too (as @anvacaru pointed out). I think the output might need to be updated (started a job [here](https://github.com/runtimeverification/kontrol/actions/runs/10634211935/job/29481041032)). I'll need to be afk for several hours, but I'll finish it and merge as soon as I'm back.

palinatolmach:
@ehildenb @anvacaru ok, I can confirm that both 
```
usegas = true
```
in `foundry.toml`, and `--use-gas` work correctly and turn on gas computations; otherwise, it's off by default.

PR #815, opened 2024-09-10
Allow failing tests when updating expected output

ehildenb:
Currently, when any test fails on a given PR, the update-expected-output job fails. That's not ideal, because it may be able to provide a partial update which can be nice for developers.

This allows tests to fail and it still continues to try and update expected output.


anvacaru (approved):

anvacaru:
@ehildenb, some tests use hardcoded node values that cannot be updated automatically, which can lead to errors (i.e. [latest job](https://github.com/runtimeverification/kontrol/actions/runs/10794828222/job/29940009799#step:6:942)) .

ehildenb:
Yeah that makes sense, I just still want as much output updated automatically as possible if we're going to go with this route.

PR #838, opened 2024-09-26
Add support for external libraries

anvacaru:
Embedded libraries (or ones with only internal/private functions) are deployed with the smart contract. However, libraries containing public/external functions must be deployed independently.
Contracts that use external libraries will have placeholders inside the bytecode representing the library's address. These placeholders are marked in the contract JSON file under `linkReferences`.

With this pull request, if a contract has `linkReferences`:
- for each link reference, fetch the library deployed bytecode using the name and path provided in the reference
- generate a new address
- create a new KApply account with the new address and the library deployed bytecode
- store the account in a list
- replace the placeholder with the new address
- when the root of the kcfg is generated, add the new list of accounts in the state


src/tests/integration/test-data/foundry/test/ExternalLibTest.t.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.13;
+
+import "forge-std/Test.sol";
+
+
```
palinatolmach, line 6: Extra empty line?
```suggestion
```

src/kontrol/prove.py
```diff
@@ -1409,3 +1425,59 @@ def _final_term(empty_config: KInner, program: KInner, config_type: ConfigType)
...
+    external_libs: list[KInner] = []
+    address_list: dict[str, str] = {}
+
+    for lib, ref_locations in contract.external_lib_refs.items():
+        ref_contract = foundry_contracts.get(lib)
+        if ref_contract is None:
+            raise ValueError(f'External library not found: {lib}')
+
+        if lib not in address_list:
+            random_address_bytes = urandom(20)
```
palinatolmach, line 1453: It all looks good to me, but I'm wondering if we want to deploy libraries at random addresses. Wouldn't it make it easier to debug if they were deterministic? 

E.g., in Halmos, they are: https://github.com/a16z/halmos/blob/52600ef1a12c3ab9ecfec6f75513bba72b78f4cc/src/halmos/sevm.py#L1141

src/kontrol/prove.py
```diff
@@ -1409,3 +1424,58 @@ def _final_term(empty_config: KInner, program: KInner, config_type: ConfigType)
...
+    external_libs: list[KInner] = []
+    address_list: dict[str, str] = {}
+
+    for lib, ref_locations in contract.external_lib_refs.items():
+        ref_contract = foundry_contracts.get(lib)
+        if ref_contract is None:
+            raise ValueError(f'External library not found: {lib}')
+
+        if lib not in address_list:
+            new_address_hex = hash_str(ref_contract.deployed_bytecode)[:40].ljust(40, '0')
```
palinatolmach, line 1452: Out of curiosity: wouldn't 
```
hash_str(ref_contract.deployed_bytecode)[:40].zfill(40)
```
work here? I just saw it being used later in the code too.

src/kontrol/prove.py
```diff
@@ -1409,3 +1424,58 @@ def _final_term(empty_config: KInner, program: KInner, config_type: ConfigType)
             KVariable('STORAGESLOTLIST_FINAL'),
         ],
     )
+
+
+def _process_external_library_references(contract: Contract, foundry_contracts: dict[str, Contract]) -> list[KInner]:
+    """Create a list of KInner accounts for external libraries used in the given contract.
+
+    This function identifies external library placeholders within the contract's deployed bytecode,
+    deploys the required external libraries at random addresses, replaces the placeholders with the
```
palinatolmach, line 1433:
```suggestion
    deploys the required external libraries at the address generated based on the first 20 bytes of the hash of the library's bytecode, replaces the placeholders with the
```

palinatolmach (approved):
lgtm, left two small comments!

src/kontrol/prove.py
anvacaru: `zfill` automatically adds `0` to the left, whereas `just` will add the given char to the right-hand side. Both options work in this case.

PR #840, opened 2024-09-27
allow consecutive startPranks

### LINKED ISSUE #389: Display the proof execution time

anvacaru:
Closes #389

This updates the `#setPrank` rule to fix a bug for which the execution gets stuck when a new `startPrank` is called while a prank is already active.


src/tests/integration/test-data/foundry/running-tests.log
PetarMax, line 1: Empty file?

src/kontrol/kdist/cheatcodes.md
```diff
@@ -1328,7 +1328,7 @@ Will also return true if REASON is `.Bytes`.
            <prevOrigin> _ => OG </prevOrigin>
            <newCaller> _ => NEWCALLER </newCaller>
            <newOrigin> _ => NEWORIGIN </newOrigin>
-           <active> false => true </active>
+           <active> _ => true </active>
```
PetarMax, line 1331: Yes, so this will allow the next prank to override the previous one rather than not fire.

anvacaru: Yes. Currently it gets stuck.

src/tests/integration/test-data/foundry/test/PlainPrankTest.t.sol
```diff
@@ -97,6 +97,15 @@ contract PlainPrankTest is Test {
         vm.stopPrank();
     }
 
+    function test_startPrank_consecutive() public {
+        AdditionalToken token = new AdditionalToken();
+        vm.startPrank(address(0));
+        token.incrementCount();
+        vm.startPrank(address(1));
+        vm.stopPrank();
+        assert(token.count() == 1);
```
PetarMax, line 106: Could we have an assertion about the pranking address here, just to see it's working properly?

PetarMax (approved):

PR #850, opened 2024-10-07
Additional EVM optimizations

PetarMax:
This PR introduces optional additional optimizations for some of the EVM opcodes. It eliminates the word stack overflow and underflow checks, relying on the hypothesis that EVM bytecode coming from Solidity will not cause a stack overflow or underflow. The speed-up on our real-world engagements ranges between 5% and 20%, as per the following tables:

|   | PR  | master | Speedup |
|---:|-----:|--------:|---------:|
| 1 | 38  | 38     | 0.00%   |
| 2 | 30  | 30     | 0.00%   |
| 3 | 38  | 41     | 7.32%   |
| 4 | 41  | 43     | 4.65%   |
| 5 | 309 | 360    | 14.17%  |
| 6 | 460 | 575    | 20.00%  |
| 7 | 148 | 176    | 15.91%  |

|     |   PR | master | speedup |
|----:|-----:|-------:|--------:|
|  T1 |  164 |    170 |   3.53% |
|  T2 |  128 |    132 |   3.03% |
|  T3 |  100 |    106 |   5.66% |
|  T4 |  321 |    342 |   6.14% |
|  T5 | 1610 |   1702 |   5.41% |
|  T6 |  297 |    334 |  11.08% |
|  T7 |  178 |    199 |  10.55% |
|  T8 |  103 |    110 |   6.36% |
|  T9 |  112 |    120 |   6.67% |
| T10 |  947 |    993 |   4.63% |

### DIFF
--- a/src/kontrol/foundry.py
+++ b/src/kontrol/foundry.py
@@ -865,7 +865,12 @@ def foundry_show(
                 [
                     KApply(
                         '<foundry>',
-                        [KVariable('KEVM_CELL'), KVariable('CHEATCODES_CELL'), KVariable('KEVMTRACING_CELL')],
+                        [
+                            KVariable('KEVM_CELL'),
+                            KVariable('STACKCHECKS_CELL'),
+                            KVariable('CHEATCODES_CELL'),
+                            KVariable('KEVMTRACING_CELL'),
+                        ],
                     ),
                     KVariable('GENERATEDCOUNTER_CELL'),
                 ],


src/kontrol/cli.py
```diff
@@ -357,6 +357,13 @@ def parse(s: str) -> list[T]:
         action='store_true',
         help='Include auxiliary Kontrol lemmas.',
     )
+    build.add_argument(
+        '--evm-rule-optimizations',
+        dest='evm_rule_optimizations',
+        default=None,
+        action='store_true',
+        help='Include optimized KEVM rules.',
```
anvacaru, line 365: Maybe a more detailed help message?
```suggestion
        help='Include optimized KEVM rules. Uses the assumption that the EVM bytecode
that comes compiled from Solidity will not result in word stack overflows or underflows.',
```

src/kontrol/kompile.py
```diff
@@ -121,6 +124,7 @@ def foundry_kompile(
             imports=_imports,
             keccak_lemmas=options.keccak_lemmas,
             auxiliary_lemmas=options.auxiliary_lemmas,
+            evm_rule_optimizations=options.evm_rule_optimizations,
```
anvacaru, line 127: Let's also set `evm_rule_optimizations` to `true` in `apply_optimizations(self)`. This will include the optimization in the `--optimize-performance N` flag.

anvacaru: Nevermind, this is a flag for build and `optimize-performance` is a flag for `prove`.

anvacaru (approved):

src/kontrol/kompile.py
```diff
@@ -63,6 +63,9 @@ def foundry_kompile(
         options.requires
         + ([KSRC_DIR / 'keccak.md'] if options.keccak_lemmas else [])
         + ([KSRC_DIR / 'kontrol_lemmas.md'] if options.auxiliary_lemmas else [])
+        + ([KSRC_DIR / 'evm_rule_optimizations.md'] if options.evm_rule_optimizations else [])
+        + []
```
palinatolmach, line 67: What's the purpose of these empty arrays?

palinatolmach (approved):

ehildenb:
Could we do this based on a cell like `<stackChecks> false </stackChecks>` instead of as lemma inclusion? That would allow us to turn it on/off without rekompiling, by changing the value of that cell.

src/kontrol/kdist/evm_rule_optimizations.md
```diff
@@ -0,0 +1,114 @@
+Highly optimized EVM rules
+==========================
+
+The provided rules speed up execution and rely on the hypothesis that EVM bytecode
+that comes compiled from Solidity will not result in word stack overflows or underflows.
+
+```k
+module EVM-RULE-OPTIMIZATIONS
```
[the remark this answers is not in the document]
PetarMax: Do you mean in the sense that we would like to be able to monitor their application or? They are applied so many times that the resulting KCFG would be unreadable, I think? Am I misunderstanding?

PetarMax:
> Could we do this based on a cell like `<stackChecks> false </stackChecks>` instead of as lemma inclusion? That would allow us to turn it on/off without rekompiling, by changing the value of that cell.

Yes, we could.

PetarMax:
Hm, but then how do I formulate these? Structurally as well, with the `<stackChecks>` cell? I don't think I can?
```
    rule #stackUnderflow(_, _) => false [priority(30)]
    rule #stackOverflow(_, _) => false [priority(30)]
```

src/kontrol/kdist/no_stack_checks.md
```diff
@@ -0,0 +1,145 @@
...
+      requires notBool ( STATIC andBool #changesState(OP, WS) )
+      [priority(31)]
+
+    rule [super.optimized.next.smv]:
+      <k> ( #next [ OP ] => #end EVMC_STATIC_MODE_VIOLATION ) ... </k>
+      <stackChecks> false </stackChecks>
+      <wordStack> WS </wordStack>
+      <static> STATIC:Bool </static>
+      requires STATIC andBool #changesState(OP, WS)
+      [priority(32)]
```
ehildenb, line 142: Should these be at different priorities relative to each other? I guess the point is that we're more frequently having `STATIC == false`, so setting that rule to try first gives a faster path?

PetarMax: I was going off of the fact that it is very unlikely to get a `EVMC_STATIC_MODE_VIOLATION` error from compiled code, so I ordered the priorities accordingly.

PetarMax:
It would appear that the `<stackChecks>` approach speeds up the execution further when stack checks are deactivated (NSC), but slows down the execution by ~3% when they are activated (SC). Data from one of the engagements:

|   | NSC |  SC | master | NSC vs master | SC vs master |
|--:|----:|----:|-------:|-------------:|--------------:|
| 1 |  37 |  40 |     38 |        2.63% |        -5.26% |
| 2 |  30 |  31 |     30 |        0.00% |        -3.33% |
| 3 |  34 |  42 |     41 |       17.07% |        -2.44% |
| 4 |  38 |  44 |     43 |       11.63% |        -2.33% |
| 5 | 245 | 368 |    360 |       31.94% |        -2.22% |
| 6 | 460 | 587 |    575 |       20.00% |        -2.09% |
| 7 | 119 | 182 |    176 |       32.39% |        -3.41% |

PetarMax:
Some interesting updates. There are essentially three version of this optimization:
1. V1: redefines stack overflow/underflow as always false, brings in a number of optimization rules. These simplifications are added *at build time*.
2. V2: Instead of redefining the stack overflow/underflow functions, it brings in two additional rules that override the treatment of `#next` so that stack overflow and underflow are never computed. Instead of adding all at build time, it introduces an additional configuration cell so that .
3. V3: V1, but with the two additional rules introduced by V3. This is what is currently in the PR.

The results on our largest real-world engagement as follows:

|     |   V1 |   V2 |   V3 | master | V1 speedup | V2 speedup | V3 speedup |
|----:|-----:|-----:|-----:|-------:|-----------:|-----------:|-----------:|
|  T1 |  164 |  161 |  160 |    170 |      3.53% |      5.29% |      5.88% |
|  T2 |  128 |  123 |  123 |    132 |      3.03% |      6.82% |      6.82% |
|  T3 |  100 |   93 |   93 |    106 |      5.66% |     12.26% |     12.26% |
|  T4 |  321 |  307 |  306 |    342 |      6.14% |     10.23% |     10.53% |
|  T5 | 1610 | 1529 | 1520 |   1702 |      5.41% |     10.16% |     10.69% |
|  T6 |  297 |  259 |  256 |    334 |     11.08% |     22.46% |     23.35% |
|  T7 |  178 |  157 |  156 |    199 |     10.55% |     21.11% |     21.61% |
|  T8 |  103 |   96 |   95 |    110 |      6.36% |     12.73% |     13.64% |
|  T9 |  112 |  107 |  107 |    120 |      6.67% |     10.83% |     10.83% |
| T10 |  947 |  907 |  892 |    993 |      4.63% |      8.66% |     10.17% |

and we see that V2 and V3 outperform V1, and that V3 is consistently slightly faster than V2. In addition, V2 exhibits a slowdown with respect to V1/V3 of ~3% when stack checks are turned on, probably due to the combination of an additional cell and additional high-priority rules that are always checked.

V1, given these results, is out of the picture.

The pros of the V2 approach are that we don't have to rekompile between runs of the same proof with and without stack checks. A semi-pro is that it the speed-up is comparable to V2. The cons are that we introduce a new cell into the configuration and that there is a slowdown when stack checks are turned on.

The pros of the V3 approach are that it is the fastest, and that there is no slowdown when stack checks are turned back on. The cons are that one does have to rekompile if one wants to use re-run a proof stack checks. The mitigation of that con is that our working hypothesis is strong and that we would want to re-run a proof with stack checks very rarely.

Given this data, my choice is V3. Thoughts?

PetarMax:
We are going with V2 in the end, everything is in place and CI is passing. @ehildenb

ehildenb (approved):

ehildenb:
I approved, but something to fix. When doing `kontrol show ... --to-kevm-rules`, it attempts to produce KEVM rules for the given KCFG that can be tested/used directly as summaries, which we have some clients using. This isn't tested very well, on CI, but changing the configuration structure of Kontrol can break it.

- In particular, [here](https://github.com/runtimeverification/kontrol/blob/9fa22ce7fb5b81c7eee8ea3b13519a0ce4689706/src/kontrol/foundry.py#L862) we extract the KEVM sub-configuration to produce rules runnable with KEVM, and this extraction pattern needs to be updated with the new cell.
- Second, the rules produced should not execute past rules that do not exist in KEVM, like the Kontrol cheatcodes or other instrumentation. So when new rules are added, it's good to include them in the list of rules that are done with `--break-on-cheatcodes` [here](https://github.com/runtimeverification/kontrol/blob/9fa22ce7fb5b81c7eee8ea3b13519a0ce4689706/src/kontrol/prove.py#L440), so that we can make sure that every time a non-KEVM rule fires we save off the node before and after. This ensures that `kevm show ... --to-kevm-rules` can filter out those KCFG edges (done [here](https://github.com/runtimeverification/kontrol/blob/9fa22ce7fb5b81c7eee8ea3b13519a0ce4689706/src/kontrol/foundry.py#L905)) which contains steps from Kontrol that aren't in KEVM.

We should also add a test of `kontrol show --to-kevm-rules`, because it's something that we use for another engagement.

PetarMax:
If these rules are included in `--break-on-cheatcodes`, they will make the feature completely unusable in any real-world debugging since they will fire on every `#next`, it’d be like `--break-every-step`, which simply does not scale. It is essential to be able to debug use of actual cheatcodes.

I am also not quite understanding in which context `--to-kevm-rules` is useful, hasn’t this been replaced by CSE? Which client would be running KEVM and not Kontrol? 

All of this indicates to me that these rules should perhaps be added to KEVM (just like the `<useGas>` rules), not Kontrol.

PetarMax:
When it comes to testing `--to-kevm-rules`, that feels like a job for the next PR.

src/kontrol/kdist/evm_rule_optimizations.md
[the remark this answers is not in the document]
ehildenb: After discussion, we decided not to include those ones in `--break-on-cheatcodes` to avoid breaking too often.

PR #853, opened 2024-10-07
Preserving all relevant symbolic variables

PetarMax:
It is possible for the `setUp` function of a test to branch on symbolic variables. If these branches are on equality against a concrete value, e.g., `{ ?WORD:Int #Equals 0 }`, then one of the resulting two branches would no longer have `?WORD` in its configuration, as it would have been substituted throughout with `0`. As a consequence, the call to `remove_useless_constraints` when grafting the initial nodes onto the `setUp` KCFG would eliminate ` ?WORD:Int #Equals 0 }` from the constraints. In turn, this would allow the backend to re-use `?WORD` when creating fresh symbolic variables, resulting in a malformed KCFG.

This PR corrects this bug by collecting all of the variables on which the branching has occurred and forbidding their removal.

### DIFF
--- a/src/kontrol/prove.py
+++ b/src/kontrol/prove.py
@@ -15,7 +15,7 @@ from kevm_pyk.utils import KDefinition__expand_macros, abstract_cell_vars, run_p
 from multiprocess.pool import Pool  # type: ignore
 from pyk.cterm import CTerm, CTermSymbolic
 from pyk.kast.inner import KApply, KSequence, KSort, KVariable, Subst
-from pyk.kast.manip import flatten_label, set_cell
+from pyk.kast.manip import flatten_label, free_vars, set_cell
 from pyk.kcfg import KCFG, KCFGExplore
 from pyk.kcfg.minimize import KCFGMinimizer
 from pyk.kore.rpc import KoreClient, TransportType, kore_server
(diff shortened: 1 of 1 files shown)


src/kontrol/prove.py
```diff
@@ -756,8 +756,19 @@ def _method_to_cfg(
...
+        # When minimizing constraints, we need to make sure not to forget any variables
+        # that might have been instantiated by a branching in the setup KCFG
+        keep_vars: set[str] = set().union(
+            *[
+                free_vars(constraint)
+                for split in cfg.splits()
+                for _, csubst in split.splits.items()
+                for constraint in csubst.constraints
+            ]
+        )
```
anvacaru, line 768: Can you try this instead? It should create the set directly instead of creating a list of sets and then unpacking them.
```suggestion
        keep_vars: set[str] = {
            var
            for split in cfg.splits()
            for _, csubst in split.splits.items()
            for constraint in csubst.constraints
            for var in free_vars(constraint)
        }
```

anvacaru (approved):
Approved with a suggestion.

src/kontrol/prove.py
PetarMax: Yes, this works (if CI passes).

PR #867, opened 2024-10-15
add `no-stack-checks` to `kontrol.toml` file

anvacaru:
Add the `--no-stack-checks` option to the toml file generated when a new project is created.

Additionally, this PR inverses the `dest` for the flags starting with `--no-...` with the `dest` also starting with `no_...`, such as
- `--no-metadata`, which is now used to set `metadata` (instead of `no_metadata`)
- `--no-silence-warnings`
- `--no-forge-build`
- `--no-stack-checks`

The logic of this flag processing is changed accordingly. 

The reason for this changes is due to the way [parse_toml_args](https://github.com/runtimeverification/k/blob/bd34ca161369a5e1731c61074ed04607c8cc5bad/pyk/src/pyk/cli/pyk.py#L607) handles options starting with `no-` in the `kontrol.toml` file: it strips the `no-`, and sets the remaining option string to the negation of the argument provided. That's needed to accommodate options such as
```
  --break-on-calls      Store a node for every EVM call made.
  --no-break-on-calls   Do not store a node for every EVM call made (default).
```
But prevents `no-stack-checks` and other similar args from being parsed correctly.


PetarMax (approved):

PetarMax:
We should note somewhere that Pi2-related usage of Kontrol should not use `--no-stack-changes`.

src/kontrol/options.py
```diff
@@ -471,7 +471,7 @@ def apply_optimizations(self) -> None:
             self.smt_retry_limit = 0
             self.max_depth = 100000
             self.max_iterations = 10000
-            self.no_stack_checks = True
+            self.stack_checks = True
```
PetarMax, line 474:
```suggestion
            self.stack_checks = False
```

PetarMax (approved):

PR #881, opened 2024-11-09
RichHandler: Increase width of the logger

anvacaru:
A downside of using the RichHandler is the logger's fixed width. When `--verbose` or `--debug` is used, lines longer than the default width are wrapped on a new line, and lines shorter are padded with whitespace.

Here, I'm increasing the width to 1500.


anvacaru:
Closing in favor of #882

PR #899, opened 2024-12-03
Implement kevm.forgetBranch cheatcode

anvacaru:
blocked on: ~https://github.com/runtimeverification/kontrol/pull/903~
blocked on: ~https://github.com/runtimeverification/kontrol-cheatcodes/pull/15~
blocked on: ~https://github.com/runtimeverification/evm-semantics/pull/2662~
blocked on: ~https://github.com/runtimeverification/k/pull/4700/~

Changes in this PR:
 - a new cheatcode `forgetBranch(uint256,uint8,uint256)`:
   - when called, should remove a constraint from the set of path constraints.
   - It puts a `#forget Int Int Int` production at the top of the K Cell. The `#forget` rule is set as a `cut-rule`.
 - Creating a new class, `FOUNDRYSemantics(KEVMSemantics)`, that is inheriting the KEVMSemantics.
 - Two new functions are added to `FOUNDRYSemantics`: 
   - `_check_forget_pattern(self, cterm: CTerm) -> bool`: checks if the `forget` cut-rule is at the top of the K Cell
   - `_exec_forget_custom_step(self, cterm: CTerm, cterm_symbolic: CTermSymbolic) -> KCFGExtendResult`:
       1. reads the values from the `#forget` K production and reconstructs the constraint.
       2. It simplifies the constraint using the `CTermSymbolic.simplify` endpoint and a `CTerm` created from an empty configuration.
       3. If the simplified constraint is part of the current path constraints, it gets removed, and a new `Step` is returned; otherwise, `None` is returned.
- overwrites `can_make_custom_step` and `custom_step` to check and run the `forget` pattern.

Usage example:
```Solidity
function test_forgetBranch(uint256 x) public {
    uint256 y;

    vm.assume(x > 200);
    kevm.forgetBranch(x, KontrolCheatsBase.ComparisonOperator.GreaterThan, 200);
    if(x > 200){
        y = 21;
    } else {
        y = 42;
    }
}
```


src/kontrol/foundry.py
```diff
@@ -82,6 +94,75 @@
...
+        )
+        self._cached_subst = abstract_pattern.match(cterm.cell('K_CELL'))
+        return self._cached_subst is not None
+
+    def _exec_forget_custom_step(self, cterm: CTerm, cterm_symbolic: CTermSymbolic) -> KCFGExtendResult | None:
+        """Given a CTerm, remove the constraint that's placed at the top of the K_CELL in the 'FOUNDRY-ACCOUNTS.forget' cut-rule.
+
+        :param cterm: CTerm of a proof node.
+        :param cterm_symbolic: CTermSymbolic instance
+        :return: If the K_CELL matches the forget_pattern and the constraint is successfully removed, returns a Step with depth 1 along with the new configuration, registering that the `cheatcode_forget` rule has been applied. If the constraint is not found, returns None.
```
palinatolmach, line 121: Obviously a nit, so feel free to ignore, but I think docstring lines are recommended to be <= 79 symbols.
```suggestion
        :return: If the K_CELL matches the forget_pattern and the constraint is 
            successfully removed, returns a Step with depth 1 along with the new 
            configuration, registering that the `cheatcode_forget` rule has been 
            applied. If the constraint is not found, returns None.
```

src/kontrol/kdist/cheatcodes.md
```diff
@@ -1691,6 +1711,8 @@ Selectors for **implemented** cheat code functions.
     rule ( selector ( "mockCall(address,bytes,bytes)" )            => 3110212580 )
     rule ( selector ( "mockFunction(address,address,bytes)" )      => 2918731041 )
     rule ( selector ( "copyStorage(address,address)" )             => 540912653  )
+    rule ( selector ( "replaceConstraint(uint256,uint8,uint256,bool)" ) => 1408652492 )
```
palinatolmach, line 1714: We should probably remove it or at least move to the "unimplemented cheatcode" section:
```suggestion
```

src/kontrol/foundry.py
palinatolmach, line 121: Obviously a nit, so feel free to ignore, but I think docstring lines are recommended to be <= 79 symbols, including function parameter descriptions: https://google.github.io/styleguide/pyguide.html#doc-function-args.
```suggestion
        :return: If the K_CELL matches the forget_pattern and the constraint is 
            successfully removed, returns a Step with depth 1 along with the new 
            configuration, registering that the `cheatcode_forget` rule has been 
            applied. If the constraint is not found, returns None.
```
```

File: src/kontrol/foundry.py [hunk 3]
```diff
@@ -82,6 +94,75 @@
...
+        """
+        _operators = ['_==Int_', '_!=Int_', '_<=Int_', '_<Int_', '_>=Int_', '_>Int_']
```
palinatolmach, line 123: It might be
```suggestion
        _operators = ['_==Int_', '_=/=Int_', '_<=Int_', '_<Int_', '_>=Int_', '_>Int_']
```

src/tests/integration/test-data/end-to-end-prove-all
```diff
@@ -1,14 +1,25 @@
 CounterTest.test_Increment()
+ForgetBranchTest.test_forgetBranch(uint256)
```
anvacaru, line 2: I only added this new test, rest of the changes in this file come from alphabetical sorting.

src/kontrol/foundry.py
```diff
@@ -155,14 +157,145 @@ def _exec_rename_custom_step(self, cterm: CTerm) -> KCFGExtendResult | None:
...
+        :return: `True` if the pattern matches and a custom step can be made; `False` otherwise.
+        """
+        abstract_pattern = KSequence(
+            [
+                KApply('cheatcode_forget', [KVariable('###TERM1'), KVariable('###OPERATOR'), KVariable('###TERM2')]),
+                KVariable('###CONTINUATION'),
+            ]
+        )
+        self._cached_subst = abstract_pattern.match(cterm.cell('K_CELL'))
+        return self._cached_subst is not None
```
ehildenb, line 174: The optimization of caching the `_cached_subst` seems pretty unnecessary to me. As far as I can tell, it will only fire _if_ `_check_forget_pattern` is true, which should only happpen when we have `cheatcode_forget` at the top of the K cell, which only happens maybe once or twice a test? So we may set the cached result once, and use it once, but no more than that, in a given test. So it introduces state dependencies/synchronization between `can_take_custom_step` and `custom_step` for maybe less than a second of savings a single time in a proof.

ehildenb: Do we have data about how much time it's saving?

ehildenb: We can change `_check_forget_pattern` to return a `Subst | None`, and it can still be tested as a boolean where it's used.

src/kontrol/foundry.py
```diff
@@ -155,14 +157,145 @@ def _exec_rename_custom_step(self, cterm: CTerm) -> KCFGExtendResult | None:
...
+            initial_cterm=initial_cterm,
+            constraints_to_remove=[pos_constraint, neg_constraint],
+            constraints_to_keep=constraints_to_keep,
+            constraints=set(cterm.constraints),
+            empty_config=empty_config,
+        )
+
+        # Update the K_CELL with the continuation
+        new_cterm = CTerm.from_kast(set_cell(cterm.kast, 'K_CELL', KSequence(subst['###CONTINUATION'])))
+        return Step(CTerm(new_cterm.config, new_constraints), 1, (), ['cheatcode_forget'], cut=True)
```
ehildenb, line 285: We shoudn't be returning a `Step` here, because that indicates taking a rewrite step of depth 1. We should instead be returning an abstraction.

That makes me think, we could maybe integrate this into `abstract_node`, it's used here by the exploration: https://github.com/runtimeverification/k/blob/547e2cc337d08ba19820a3c22956091c0a9aebf8/pyk/src/pyk/kcfg/explore.py#L227. And KEVM's node abstractor is here: https://github.com/runtimeverification/evm-semantics/blob/b40fd7b09fd79dcb3fa13fc6d8f9a02451e8a475/kevm-pyk/src/kevm_pyk/kevm.py#L125.

So that could look for the correct pattern, and remove the appropriate constraints.

src/kontrol/foundry.py
```diff
@@ -155,14 +157,145 @@ def _exec_rename_custom_step(self, cterm: CTerm) -> KCFGExtendResult | None:
...
+        neg_constraint = mlEqualsTrue(notBool(KApply(_operators[int(operator.token)], fst_term, snd_term)))
+        # To be able to better simplify, we maintain range constraints on the variables present in the constraint
+        constraint_vars: frozenset[str] = free_vars(fst_term).union(free_vars(snd_term))
+        constraints_to_keep: set[KInner] = _find_constraints_to_keep(cterm, constraint_vars)
+
+        # Set up initial configuration for constraint simplification, and simplify it to get all
+        # of the kept constraints in the form in which they will appear after constraint simplification
+        kevm = KEVM(kdist.get('kontrol.foundry'))
+        empty_config: CTerm = CTerm.from_kast(kevm.definition.empty_config(GENERATED_TOP_CELL))
+        initial_cterm, _ = cterm_symbolic.simplify(CTerm(empty_config.config, constraints_to_keep))
```
ehildenb, line 269: Why do we need to invoke the simplifier? I think we should really try not to have to do that.

Stevengre:
I think we can have both. Custom command is a debugging method with quick response, cheatcode is a fixed proof for CI?

PetarMax:
After having read through, I think we should definitely have both, because the separate custom command allows for quick debugging and then the Solidity-level forgetting allows for a continuous proof.

src/kontrol/foundry.py
PetarMax: Oh, is this not a rewrite step of depth 1? It does removes the top of the `<k>` cell, and the only reason it's not in the semantics as a rewrite rule is because there we don't have the required expressivity, so we delegate it to `custom_step`.

src/kontrol/foundry.py
PetarMax: All of the path constraints that we have will have been previously simplified before being added. Therefore, if we would like to remove a constraint that we believe is there, we need to simplify it first, otherwise it will not match.

src/kontrol/foundry.py
ehildenb: In that case we should be doing a rewrite to a state that it actually could rewrite to, then inserting an abstraction. If it can't be expressed as a rewrite then it's not a step. We could also do an abstraction, then a rewrite (return an `[Abstract(...), Step(...)]`.

PetarMax: That's a good point, I will look into it.

PetarMax: It is likely to hinder minimization since we can't minimize through covers.

lucasmt (approved):
The cheatcode seems to be working on our tests, so I'm approving this and we can revisit it later to revise the implementation/add a command-line version.

PR #902, opened 2024-12-05
Add cheatcodes that generate symbolic variables with custom names

anvacaru:
Using `custom_step` to rename symbolic variables generated by cheatcodes.
Blocked on : ~https://github.com/runtimeverification/kontrol-cheatcodes/pull/16~

In this PR: 

1. Adding a new `#rename` rule that all cheatcodes that generate symbolic variables would rewrite into.
```
    syntax RenameTarget ::= Int | Bytes | Map
    syntax KItem ::= #rename ( RenameTarget , String ) [symbol(foundry_rename)]
 // ---------------------------------------------------------------------------
    rule [rename]: <k> #rename(_,_) => .K ... </k>
```

The goal is to have a single cut-rule `rename`. 
2. In `KontrolSemantics`, we add:
- ` _check_rename_pattern(self, cterm: CTerm) -> bool`:
  -  checks if the `#rename` cut-rule is at the top of the K Cell.
-  ` _exec_rename_custom_step(self, cterm: CTerm) -> KCFGExtendResult | None`:
   - takes the newly generated symbolic variable (`target_var`) and new name (`new_name`) from the top of the K Rule.
   - ensures that the new name is unique in the CTerm, if not, it appends an indexed suffix.
   - using `Subst` replaces all instances of the `target_var` with a new `KVariable(new_name.token, target_var.sort)` in the given `CTerm` configuration and constraints.

3. Added new unit and show integration tests.
4. Bumping `kontrol-cheatcodes` and `forge-std` versions for `tests/integration/test-data/foundry`


src/kontrol/utils.py
```diff
@@ -25,6 +27,29 @@
...
+    :param name: name of a KVariable
+    :param cterm: cterm
+    :return: Returns the name if it's not used, otherwise appends a suffix.
+    :rtype: str
+    """
+    new_config = Subst({name: token(True)})(config)
+    if new_config == config:
+        return name
+    print(1, config)
+    print(2, new_config)
```
PetarMax, line 42: These prints probably need to go :)

anvacaru: Right, missed them. Thanks!

src/kontrol/utils.py
```diff
@@ -25,6 +27,29 @@
...
 
+def ensure_name_is_unique(name: str, config: KInner) -> str:
+    """Ensure that a given name for a KVariable is unique within the context of a CTerm.
+
+    :param name: name of a KVariable
+    :param cterm: cterm
+    :return: Returns the name if it's not used, otherwise appends a suffix.
+    :rtype: str
+    """
+    new_config = Subst({name: token(True)})(config)
```
PetarMax, line 38: Do we not have `free_vars` or something like that to figure out the variables we have? Isn't it easier to use that and then examine the set?

anvacaru: Yup, I've updated the function to use `free_vars`

PetarMax:
Slightly late to the party - it would be great if we could make a minor adjustment where the indices are not separated from the name, like the backend has been doing for a while now - `X`, `X0`, `X1`, etc.

Issue #817, opened 2024-09-11
Duplicated `extra_output` in `kontrol init --skip-forge`


palinatolmach:
We're adding `extra_output = ['storageLayout']` to `foundry.toml` on `kontrol init` without checking if it was present. If it's already present in `foundry.toml`, it might lead to the following issue:
```
Error: 
failed to extract foundry config:
foundry.toml error: TOML parse error at line 15, column 1
   |
15 | extra_output = ['storageLayout']
   | ^
duplicate key `extra_output` in table `profile.default`
 in foundry.toml TOML file
```

palinatolmach:
Closed by https://github.com/runtimeverification/kontrol/pull/910.

PR #910, opened 2024-12-10
Stop appending the storageLayout extra output to foundry.toml

anvacaru:
As it's not needed anymore.
closes: #817


palinatolmach (approved):
Nice! It will also close this issue: https://github.com/runtimeverification/kontrol/issues/817.

PR #914, opened 2024-12-13
Reduce the number of processes for stable integration test with booster

Stevengre:


palinatolmach:
Closing as we haven't got any OOM issues on booster CI jobs in a while.

PR #915, opened 2024-12-13
Fix nested struct array rule generation

palinatolmach:
Follow up to https://github.com/runtimeverification/kontrol/pull/830.

This PR ensures that the elements of nested structs up to an arbitrary depth all have `_{i}` indices appended to their names if they correspond to an array element.  It also factors out the processing of tuple array components into a separate function `process_tuple_array`.

This PR also adds a test that exercises the previously missing edge case, which manifested in variables on the LHS and in the `ensures` clause of the `#abiCalldata` rule in `contract.k` not having proper indexes (`..._0`) set.

 Please note that this test branches on the possible length of dynamically sized `strings` in the calldata, so it is meant to test `kontrol build` and is not run during the `kontrol prove` stage:

```solidity
pragma solidity =0.8.13;

import "forge-std/Test.sol";

interface ProjectConfig {
    struct ChainEnv {
        string chainName;
        string chainIndex;
    }

    struct Vault {
        string tokenName;
        address oracle;
        uint256 allowedToTrade;
        InputData params;
        uint256 reserves;
    }

    struct InputData {
        uint256 ltv;
        uint256 rate;
        uint256 exchangeRate;
        uint256 utilization;
    }

    struct AssetAddresses {
        address cToken;
        address dToken;
    }

    struct VaultWithAddresses {
        Vault vault;
        AssetAddresses addresses;
    }
}

contract NestedStructArrayTest is Test, ProjectConfig {
    // this test branches on dynamically sized calldata and not run in CI; it is only meant to test `kontrol build`
    function testVaultSignature(ChainEnv calldata environment, VaultWithAddresses[] memory listings) external { 
        assert(listings.length == 1);
    }
}
```

### DIFF
--- a/src/kontrol/solc_to_k.py
+++ b/src/kontrol/solc_to_k.py
@@ -176,6 +176,58 @@ class Input:
                 idx += 1
         return inputs
 
+    @staticmethod
+    def process_tuple_array(input: Input) -> list[Input]:
+        """
+        Processes an input tuple array, appending `_index` to the names of components
+        and recursively handling nested tuples.
+        """
+        processed_components = []
+
+        if input.array_lengths is None:
+            raise ValueError(f'Array length bounds missing for {input.name}')
+
+        for i in range(input.array_lengths[0]):
+            for _c in input.components:
+                if _c.type == 'tuple':
+                    # Recursively process nested tuple components
+                    sub_components = tuple(Input.add_index_to_components(_c.components, i))
+                else:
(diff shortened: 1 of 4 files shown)


ehildenb:
:heart_eyes: nice refactoring!

anvacaru (commented):
Nice! I was able to build my project using this fix

anvacaru (approved):

PR #916, opened 2024-12-15
Set `processed_link_refs` to `true` only in `setup` or constructor

palinatolmach:
Related: https://github.com/runtimeverification/kontrol/pull/861 and https://github.com/runtimeverification/kontrol/pull/838.

This PR fixes an issue identified by @lisandrasilva, where the second of two proofs calling a dynamically linked library fails when they're executed sequentially. The `<accounts>` cell of the second proof is missing an account corresponding to the library, which is present when the proof is run in isolation (or with `-j2` in parallel with the first one) and the proof is then passing. 

This PR changes the logic of re-deploying dynamically linked libraries by ensuring it doesn't happen _only_ if we've run `init` or `setUp` method for a contract, in which case it'll be grafted onto the proof; otherwise, the libraries will be deployed for every individual proof.

It also adds a second test `testSum` to `ExternalLibTest`.


src/tests/integration/test-data/foundry-prove-all
```diff
@@ -115,6 +115,7 @@ ExpectRevertTest.testFail_ExpectRevert_failAndSuccess()
 ExpectRevertTest.testFail_expectRevert_false()
 ExpectRevertTest.testFail_expectRevert_multipleReverts()
 ExternalLibTest.testSquare(uint256)
+ExternalLibTest.testSum(uint256,uint256)
```
anvacaru, line 118:
```suggestion
ExternalLibTest.testSum()
```

src/tests/integration/test-data/foundry-prove-skip-legacy
```diff
@@ -103,6 +103,7 @@ ExpectRevertTest.testFail_ExpectRevert_failAndSuccess()
 ExpectRevertTest.testFail_expectRevert_false()
 ExpectRevertTest.testFail_expectRevert_multipleReverts()
 ExternalLibTest.testSquare(uint256)
+ExternalLibTest.testSum(uint256,uint256)
```
anvacaru, line 106:
```suggestion
ExternalLibTest.testSum()
```

anvacaru (approved):
Good catch!

src/tests/integration/test-data/foundry-prove-all
palinatolmach: Thanks! Fixed in https://github.com/runtimeverification/kontrol/pull/916/commits/490164d73a693eac72fb88a36f7e31a25881329c.

src/tests/integration/test-data/foundry-prove-skip-legacy
palinatolmach: Fixed in https://github.com/runtimeverification/kontrol/pull/916/commits/490164d73a693eac72fb88a36f7e31a25881329c.

PR #921, opened 2024-12-18
Add from_option_string to RPCOptions

anvacaru:
To preserve the Argparse argument when the `dest` does not match the argument itself, `from_option_string` must define that mapping.
```
        args.add_argument(
            '--no-log-rewrites',
            dest='log_succ_rewrites',
            default=None,
            action='store_false',
            help='Do not log traces of any simplification and rewrite rule application.',
        )
```
 In this case, `(no-)log-rewrites` did not match `log_succ_rewrites` and the `RPCOptions` class did not have a `from_option_string` function. When `no-log-rewrites: true` was used in the TOML file, a new option, `log_rewrites` was set to `false` instead of setting `log_succ_rewrites` to `false.`

Having `log-rewrites` mapped to `log_succ_rewrites` in the `RPCOptions.from_option_string` class means that when the TOML file has `no-log-rewrites: true,` the `log_succ_rewrites` will be set to `False.`


palinatolmach (approved):
Nice find!

PR #941, opened 2025-01-23
Build Custom Kontrol Versions using a workflow

F-WRunTime:
- Adding a new feature to build kontrol images using kup and --overrides

Supports using custom test builds in kaas and local


.github/workflows/kup-build-kontrol.yml
```diff
@@ -0,0 +1,38 @@
...
+            docker push ghcr.io/runtimeverification/kontrol-custom:${{ github.run_id }}
+          else
+            docker stop --time=0 kontrol-build-with-kup-${{ github.run_id }}
+            echo "No overrides provided"
+            exit 1
+          fi
+      - name: 'Tear down Docker'
+        if: always()
+        run: |
+          docker stop --time=0 kontrol-build-with-kup-${{ github.run_id }}
```
ehildenb, line 38: IMO, we should instead I think take in the specific K/KEVM/Haskell/LLVM version to use. If they are not specified, it should default to what the repository `main` branch would use via the `deps/kevm` and `deps/k` files.

```suggestion
---
name: 'KUP Build Kontrol'
on:
  workflow_dispatch:
    inputs:
      kevm-version: 
        description: 'Branch/Tag to use for KEVM'
        required: false
        default: ''
      k-version: 
        description: 'Branch/Tag to use for K'
        required: false
        default: ''
      llvm-version: 
        description: 'Branch/Tag to use for LLVM Backend'
        required: false
        default: ''
      haskell-version: 
        description: 'Branch/Tag to use for Haskell Backend'
        required: false
        default: ''

jobs:
  build-kontrol:
    runs-on: [self-hosted, normal]
    steps:
      - name: 'Login to GitHub Container Registry'
        uses: docker/login-action@v2
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
      - name: 'Build Kontrol'
        shell: bash
        run: |
          set -o pipefail
          docker run --rm -it --detach --name kontrol-build-with-kup-${{ github.run_id }} ghcr.io/runtimeverification/kup:latest
          if [ ! -n "${{ inputs.kevm-version }}" ]; then
              KEVM_VERSION=$(get KEVM version from deps/kevm)
          else
              KEVM_VERSION="${{ inputs.kevm-version }}"
          fi
          if [ ! -n "${{ inputs.k-version }}" ]; then
              K_VERSION=$(get K version from deps/k)
          else
              K_VERSION="${{ inputs.k-version }}"
          fi
          if [ ! -n "${{ inputs.llvm-version }}" ]; then
              LLVM_VERSION=$(get LLVM version from deps/k)
          else
              LLVM_VERSION="${{ inputs.llvm-version }}"
          fi
          if [ ! -n "${{ inputs.haskell-version }}" ]; then
              HASKELL_VERSION=$(get HASKELL version from deps/k)
          else
              HASKELL_VERSION="${{ inputs.k-version }}"
          fi
          docker exec kontrol-build-with-kup-${{ github.run_id }} /bin/bash -c "kup install kontrol --override kevm ${KEVM_VERSION} --override kframework ${K_VERSION} --override llvm-backend ${LLVM_VERSION} --override haskell-backend ${HASKELL_VERSION}"
      - name: 'Tear down Docker'
        if: always()
        run: |
          docker stop --time=0 kontrol-build-with-kup-${{ github.run_id }}
```

README.md
```diff
@@ -71,7 +71,65 @@ To update the expected output of the tests, use the `--update-expected-output` f
...
+│   │   ├── llvm-backend - github:runtimeverification/llvm-backend (d5eab4b)
+│   │   │   ├── immer-src - github:runtimeverification/immer (4b0914f)
+│   │   │   └── rv-utils - github:runtimeverification/rv-nix-tools (a650588)
+│   │   └── rv-utils - follows kevm/k-framework/llvm-backend/rv-utils
+│   └── rv-utils - follows kevm/k-framework/rv-utils
+└── rv-utils - follows kevm/rv-utils
+```
+Now run a build using kup and specific dependency overrides:    
+
+`kup install kontrol --override kevm/haskell-backend "hash/branch name" --override kevm/k-framework/haskell-backend "hash/branch name"`
```
asavienko, line 118: What do you think? 
branch name -> branch_name
```suggestion
`kup install kontrol --override kevm/haskell-backend "hash/branch_name" --override kevm/k-framework/haskell-backend "hash/branch_name"`
```

asavienko (changes requested):
Let's implement "get KEVM version" before merging this PR 
```
$(get KEVM version from deps/kevm)
```

README.md
F-WRunTime: Working on it [over here](https://github.com/runtimeverification/devops-actions/actions/workflows/kup-build-kontrol.yml) so I can test it

F-WRunTime:
> Let's implement "get KEVM version" before merging this PR
> 
> ```
> $(get KEVM version from deps/kevm)
> ```

Agreed, Working on it over here so I can iterate and test before merging to kontrol 
https://github.com/runtimeverification/devops-actions/actions/workflows/kup-build-kontrol.yml

.github/workflows/kontrol-push-unfixed-deps.yml
```diff
@@ -0,0 +1,84 @@
...
+              LLVM_OVERRIDE="--override kevm/k-framework/llvm-backend ${{ inputs.llvm-version }}"
+          fi
+          if [ ! -n "${{ inputs.haskell-version }}" ]; then
+              K_VERSION=$(curl -s https://raw.githubusercontent.com/runtimeverification/kontrol/master/deps/k_release)
+              HASKELL_TAG=$(curl -s https://raw.githubusercontent.com/runtimeverification/k/v${K_VERSION}/deps/haskell-backend_release)
+              HASKELL_HASH=$(git ls-remote https://github.com/runtimeverification/haskell-backend.git refs/tags/${HASKELL_TAG} | awk '{print $1}')
+              HASKELL_OVERRIDE="--override kevm/k-framework/haskell-backend ${HASKELL_HASH}"
+          else
+              HASKELL_OVERRIDE="--override kevm/k-framework/haskell-backend ${{ inputs.haskell-version }}"
+          fi
```
jberthold, line 67: I think we should not generate `--override` options for components that aren't actually overridden.
1. I don't think there is _need_ to set the `XY_OVERRIDE` at all if we aren't actually overriding the component version. For example, by default `kontrol` version A will depend on `K` version B, so we won't need to explicitly override with a hash.
2. There may be a problem in the overrides for HS and LLVM backend here. Let's say we want to override the `K` dependency, then that different `K` should _bring in its own versions of HS and LLVM backends_ (unless we again override those. The way this is currently coded using explicit overrides for each component, we would read the HS and LLVM backend dependencies from the _default_ `K` dependency of `kontrol` instead of using the ones brought in by the overridden `K` version.

It may still be good to print out each version but then we need to set the `K_VERSION` in the two final `if` clauses above to `inputs.k_version` (if that variable is set) to retrieve the HS and LLVM versions unless overridden.

F-WRunTime: Thanks that's a good point, it might cause a contradiction and very unexplained behavior. 
Running new build now with this suggestions [here](https://github.com/runtimeverification/devops-actions/actions/runs/12939452551/job/36091733936)

ehildenb: Hmmmm, that's a good point. I guess we should just generate the `--override` for the ones the use specifies on the workflow prompt, and not for the others. We can do that with a bash array variable probably.

F-WRunTime: On the note of maintain printing "what" went into the build. I'm testing that as well in this run. 
Where it dumps the contents of 'kup list kontrol --inputs' to the summary. Hopefully it formats it nicely like the CLI does. 
This should then tell us everything we need straight from kup.

ehildenb:
```suggestion
          if [ ! -n "${{ inputs.k-version }}" && -n "${ inputs.kevm-version }}" ]; then
              K_TAG=$(curl -s https://raw.githubusercontent.com/runtimeverification/kontrol/master/deps/k_release)
              K_HASH=$(git ls-remote https://github.com/runtimeverification/k.git refs/tags/v${K_TAG} | awk '{print $1}')
              K_OVERRIDE="--override kevm/k-framework ${K_HASH}"
          else
              K_OVERRIDE="--override kevm/k-framework ${{ inputs.k-version }}"
          fi
          if [ ! -n "${{ inputs.llvm-version }}" && -n "${{ inputs.k-version }}" && -n "${{ inputs.kevm-version }}" ]; then
              K_VERSION=$(curl -s https://raw.githubusercontent.com/runtimeverification/kontrol/master/deps/k_release)
              LLVM_TAG=$(curl -s https://raw.githubusercontent.com/runtimeverification/k/v${K_VERSION}/deps/llvm-backend_release)
              LLVM_HASH=$(git ls-remote https://github.com/runtimeverification/llvm-backend.git refs/tags/v${LLVM_TAG} | awk '{print $1}')
              LLVM_OVERRIDE="--override kevm/k-framework/llvm-backend ${LLVM_HASH}"
          else
              LLVM_OVERRIDE="--override kevm/k-framework/llvm-backend ${{ inputs.llvm-version }}"
          fi
          if [ ! -n "${{ inputs.haskell-version }}" && -n "${{ inputs.k-version }}" && -n "${{ inputs.kevm-version }}" ]; then
              K_VERSION=$(curl -s https://raw.githubusercontent.com/runtimeverification/kontrol/master/deps/k_release)
              HASKELL_TAG=$(curl -s https://raw.githubusercontent.com/runtimeverification/k/v${K_VERSION}/deps/haskell-backend_release)
              HASKELL_HASH=$(git ls-remote https://github.com/runtimeverification/haskell-backend.git refs/tags/${HASKELL_TAG} | awk '{print $1}')
              HASKELL_OVERRIDE="--override kevm/k-framework/haskell-backend ${HASKELL_HASH}"
          else
              HASKELL_OVERRIDE="--override kevm/k-framework/haskell-backend ${{ inputs.haskell-version }}"
          fi
```

ehildenb: I made a suggestion, but I still think the approach of using an array we accumulate the overrides in that the user actually specifies is best.

F-WRunTime: I just leave them empty and leave them there. Works all the same.

F-WRunTime: Okay final run of this latest push is here
https://github.com/runtimeverification/devops-actions/actions/runs/12939752629

F-WRunTime:
Final example of latest test run through 
https://github.com/runtimeverification/devops-actions/actions/runs/12939752629

ehildenb (approved):

PR #976, opened 2025-03-07
Eliminate deadcode from the Python codebase

ehildenb:
Part of: https://github.com/runtimeverification/kontrol/issues/977.

I used the Python utility deadcode https://pypi.org/project/deadcode/ to detect unused code in this codebase. Still all of the methods in `__main__.py` which are of the form `def exec_...` are marked as dead by this tool because it does not understand the dynamic dispatch approach we used in `def main():`.

I suspect that porting the updated sourcemaps approach from Simbolik to this tool resulted in also porting a lot of functions which are not used in Kontrol.

I came across this refactoring as I was trying to understand how to avoid rekompiling every time `kontrol build` is called.


anvacaru:
It might be worth it to add `deadcode` to `pyproject.toml` and use it in the current `format` or `check` `Makefile` targets.

ehildenb:
I added an issue here about deadcode: https://github.com/runtimeverification/kontrol/issues/989

palinatolmach (approved):

Issue #170, opened 2023-11-10
Make calldata sugar generation optional


palinatolmach:
As @tothtamas28 suggested, in certain cases, we can avoid kompilation — and, therefore, having to rekompile every time we change test or smart contract code — by not generating K syntax for symbolic execution input, i.e., calldata syntactic sugar, which is only useful if the user writes lemmas

tothtamas28:
Comment from [Slack](https://runtimeverification.slack.com/archives/C05L7LWT5E3/p1699630818889109?thread_ts=1699609579.773179&cid=C05L7LWT5E3) by @ehildenb:

> note that for new aliases and macros, rekomplilation is not necessary anyway! We need to regenerate parsers perhaps, but not really because we're not even using these macros and aliases in the Java code, just in python. So we probably already don't need to rekompile!
>
>Also, even adding simplification lemma, as long as they are parsed (using the outer parsing capabilities pyk exposes, then inner parser on the rules), then we can just add those lemma using addmodule endpoint
>
>So basically, there really is never any reason to rekomplile already, even if the contracts change. Only time it's needed is if kevm or cheat codes themselves change. Everything else added via kontrol is contract specific macros or simplification lemma, all things we can add into context without kompliation.

Issue #977, opened 2025-03-07
Skip calling `kompile` when doing `kontrol build`


ehildenb:
Currently, when working with a Kontorl project, every change to the Soldiity code results in having to call `kompile` again. This is because we generate K code representing sugar for that specific contract and use that sugar in the proofs. In general, this isn't really necessary, we only are doing this to introduce macros for prettier syntax, and it slows down development of Kontrol proofs quite a bit.

We should:

- [x] Some initial cleanup. https://github.com/runtimeverification/kontrol/pull/976
- [x] Avoid generating any of the calldata sugar rules, instead generate the needed calldata directly in the proof init states. Also avoid generating the selector aliases. https://github.com/runtimeverification/kontrol/pull/978
- [x] Simplify the generated module structure now that no rules are being generated. https://github.com/runtimeverification/kontrol/pull/1003
- [x] Build different versions of Kontrol for each lemma body directly (`--no-aux-lemmas` and `--keccak-lemmas`), so they don't need to be included at `kontrol build ...` time. https://github.com/runtimeverification/kontrol/pull/1002
- [x] Do not rekompile the K definition when only Solidity code changes: https://github.com/runtimeverification/kontrol/pull/1007.
- [x] Minor cleanups to how extra lemmas are included in `kontrol prove ...`, and extending it to be included in all commands which use the RPC server. Also minor cleanups in `kontrol build ...` options parsing: https://github.com/runtimeverification/kontrol/pull/979.
- [ ] ~No need to actually build the base Kontrol definitions, other than to make sure that they do build. Maybe on CI we can shorten time by not building them? Maybe if a user is not supplying `--requires ... --module-imports ...` we could also have faster `kontrol build ...` time by just using one of the prebuilt base modules directly?~
- [x] Final cleanup and deadcode elimination.
  - Use `deadcode src/`
  - **DONE**: It's unclear if functions like `gen_bin_runtime`, `bin_runtime`, `solc_to_k`, etc... are used for anything except testing at this point. So they can probably be removed and then the repo updated/simplified quite a bit without them. https://github.com/runtimeverification/kontrol/pull/1005
- [x] Update Kontrol documentation with the following points:
  - How to use `kontrol prove ... --lemmas LEMMAS-FILE:LEMMAS-MODULE` (renamed from `--extra-module`).
[...]

PR #978, opened 2025-03-07
Do not generate calldata sugar rules

ehildenb:
Closes: https://github.com/runtimeverification/kontrol/issues/170

This PR is part of: https://github.com/runtimeverification/kontrol/issues/977

This PR removes a lot of the generated artifacts that were not being used in a meaningful way. In particular:

- The ABI calldata sugar rules are removed, because they were being macro-expanded immediately on proof construction anyway, so it's not like the user was ever seeing the sugared version of the states. Instead, we generate directly what the RHS of these rules were in our initial configuration (which involved `#abiCallData(...)` and still lists the function name), and add the preconditions (which were `ensures` clauses in the rules) directly as constraints to our initial state.
- The syntax declared for each contract is also not generated, as it was only used for making sugared versions of the calldata for rules.
- The selector alias rules are also not generated. This will perhaps be a UX change, because now if a configuration which has a selector alias somewhere on the stack is viewed by the user, now it instead will display as an integer instead of as `selector( "functionName" )`. We can probably replicate this replacement in Python directly if that is something users want back.
- All associated code that became dead via `deadcode src/` run on this repository is also removed.

In total, no sentences are generated anymore for the produced definitions. This means we can greatly simplify the build process, but for now I want to get this merged and tested, then I can apply a follow-up PR with the remaining simplifications to the build process.


src/tests/integration/test-data/show/node-refutation.expected
```diff
@@ -42,7 +42,7 @@
    │   method: test%MergeTest.test_branch_merge(uint256)
    │
    │  (90 steps)
-   ├─ 10
+   ├─ 12
```
palinatolmach, line 45: The PR looks good to me, but we probably don't expect the node numbering to change. Maybe that is flakiness in output update or integration tests which is not specific to this PR, I'm trying to reproduce locally.

palinatolmach: I can reproduce it, looks like the output depends on the relative order between `test_foundry_refute_node` and `test_foundry_merge_nodes`, which rely on the same `MergeTest.test_branch_merge` test. Apparently, most of the time `test_foundry_merge_nodes` is ran first in CI integration tests, so, in this PR, it probably makes sense to just revert the node numbering updates in `node-refutation.expected` as done in https://github.com/runtimeverification/kontrol/pull/978/commits/28dfb5a584b714d1bf6e1039520083656bed0a23.

I can't explain why we haven't run into this before, but I'm getting a similar behavior on other branches too. Will open a separate PR to address this.

palinatolmach (approved):

anvacaru (approved):

src/tests/integration/test-data/show/node-refutation.expected
ehildenb: Oh how interesting! Good catch! Thanks for doing an investigation, I'll wait for your PR to get merged. I was confused about the output changing and changing back.

Issue #980, opened 2025-03-10
Make `msg.sender` concrete


palinatolmach:
As suggested by @lisandrasilva, we should make `msg.sender` concrete instead of symbolic, and assign the same value to it as Foundry does (`0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`). Otherwise, it is causing issues when `delegatecall` is performed from a test contract.

palinatolmach:
Closed by https://github.com/runtimeverification/kontrol/pull/982.

PR #982, opened 2025-03-10
Make `msg.sender`, `tx.origin` concrete, add `--symbolic-caller` option

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/980.

This PR
- makes `msg.sender` and `tx.origin` concrete, set to the default Foundry sender (`137122462167341575662000267002353578582749290296` or `0x1804C8AB1F12E6BBF3894D4083F33E07309D1F38`) unless `--symbolic-caller` is provided or Kontrol is run in the CSE mode
- updates expected output
- fixes a typo in the `active_simbolik` parameter name (previously `active_symbolik`)


src/kontrol/prove.py
```diff
@@ -1052,13 +1060,25 @@ def _init_cterm(
 
     storage_constraints: list[KApply] = []
 
-    if config_type == ConfigType.TEST_CONFIG or active_symbolik:
+    if config_type == ConfigType.TEST_CONFIG or active_simbolik:
         init_account_list = _create_initial_account_list(contract_code, additional_accounts, recorded_state_entries)
+        origin_id = (
+            intToken(137122462167341575662000267002353578582749290296)
```
anvacaru, line 1066: Consider defining this value as a constant.

src/tests/integration/test-data/show/ArithmeticCallTest.test_double_add(uint256,uint256).cse.expected
```diff
@@ -274,7 +274,7 @@ module SUMMARY-TEST%ARITHMETICCALLTEST.TEST-DOUBLE-ADD(UINT256,UINT256):0
                      ( 491460923342184218035706888008750043977755113263 => 728815563385977040452943777879061427756277306518 )
                    </id>
                    <caller>
-                     ( 728815563385977040452943777879061427756277306518 => CALLER_ID:Int )
+                     ( 728815563385977040452943777879061427756277306518 => 137122462167341575662000267002353578582749290296 )
```
anvacaru, line 277: I thought this change shouldn't affect CSE proofs. Is this change expected?

src/kontrol/prove.py
palinatolmach: Thank you! Done in https://github.com/runtimeverification/kontrol/pull/982/commits/a3cf1416214036ed01034ba36d2b176f5ecaeb74.

src/tests/integration/test-data/show/ArithmeticCallTest.test_double_add(uint256,uint256).cse.expected
palinatolmach: It's not affecting the functions that get CSE'd (for example, the expected files for [this function](https://github.com/runtimeverification/kontrol/blob/a3cf1416214036ed01034ba36d2b176f5ecaeb74/src/tests/integration/test-data/show/Identity.applyOp(uint256).cse.expected) has not changed and still references `CALLER_ID`, `ORIGIN_ID`, and a symbolic contract address `C_IDENTITY_ID`). But the top-level test function has been treated as a normal proof with the concrete setup—except for using summaries instead of executing external calls, since we don't expect to reuse the summary of the test's execution.

lisandrasilva (approved):

anvacaru (approved):

PR #995, opened 2025-03-20
Pin Foundry version in `nix` to the CI one

palinatolmach:
Fixes the issue identified by @jberthold where the Foundry version installed with `nix develop` is not compatible with some of the integration tests as a newer version changes the bytecode of contracts and lemmas are not getting applied. This PR pins the version in `flake.nix` to the one we use in CI.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -245,17 +245,17 @@
         ]
       },
       "locked": {
-        "lastModified": 1740993113,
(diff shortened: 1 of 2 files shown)


juliankuners (approved):

jberthold:
This change makes the tests run in the [HS backend test script](https://github.com/runtimeverification/haskell-backend/blob/master/scripts/performance-tests-kontrol.sh) terminate, with fewer errors than observed before (some failures appear to be minor output differences)

jberthold (approved):

PR #1001, opened 2025-03-24
Fix `kompile` warnings.

anvacaru:
Fixing warnings related to the naming convention of unused vars.


palinatolmach (approved):

palinatolmach:
@anvacaru when you have time, can you please check why the end-to-end `AllowChangesTest.testAllowCalls` is [failing](https://github.com/runtimeverification/kontrol/actions/runs/14037559462/job/39308608320?pr=1001)? It doesn't look like it's caused by any of the changes, but it doesn't happen on [another PR](https://github.com/runtimeverification/kontrol/pull/983) that I just brought up-to-date with master, so I don't think it's caused by Foundry changes.

src/kontrol/kdist/kontrol_lemmas.md
```diff
@@ -197,9 +197,9 @@ module KONTROL-AUX-LEMMAS
     rule KI:KItem in ListItem(KJ:KItem) => KI ==K KJ [simplification]
 
     // Recursive list membership check for lists with multiple elements
-    rule KI:KItem in (ListItem(KI) Rest) => true [simplification]
-    rule KI:KItem in (ListItem(KJ) Rest) => KI in Rest [simplification]
-    rule KI:KItem in .List => false [simplification]
+    rule KI:KItem in (ListItem(KI) _REST) => true [simplification]
+    rule KI:KItem in (ListItem(_KJ) REST) => KI in REST [simplification]
+    rule _KI:KItem in .List => false [simplification]
```
anvacaru, line 202: @palinatolmach By reverting these changes, the proof is passing. However, what could cause the lemmas to stop from being applied?

palinatolmach: I tried rolling back the changes one by one, and I think it's caused by this lemma:
```
rule KI:KItem in (ListItem(_KJ) REST) => KI in REST [simplification]
```
which I guess should actually have been
```diff
- rule KI:KItem in (ListItem(_KJ) REST) => KI in REST [simplification]
+ rule KI:KItem in (ListItem(KJ) REST) => KI in REST requires KI =/=K KJ [simplification] 
```
so that it only applies if `KI` and `KJ` are different. We should probably investigate if it's correct that it was working before.

src/kontrol/kdist/kontrol_lemmas.md
```diff
@@ -197,9 +197,9 @@ module KONTROL-AUX-LEMMAS
     rule KI:KItem in ListItem(KJ:KItem) => KI ==K KJ [simplification]
 
     // Recursive list membership check for lists with multiple elements
-    rule KI:KItem in (ListItem(KI) Rest) => true [simplification]
-    rule KI:KItem in (ListItem(KJ) Rest) => KI in Rest [simplification]
-    rule KI:KItem in .List => false [simplification]
+    rule KI:KItem in (ListItem(KI) _REST) => true [simplification]
+    rule KI:KItem in (ListItem(_KJ) REST) => KI in REST [simplification]
```
palinatolmach, line 201:
```suggestion
    rule KI:KItem in (ListItem(KJ) REST) => KI in REST requires KI =/=K KJ [simplification]
```

palinatolmach (approved):

PR #1003, opened 2025-03-24
Simplify generated module structure

ehildenb:
Part of https://github.com/runtimeverification/kontrol/issues/977

Following https://github.com/runtimeverification/kontrol/pull/978, we don't need to generate any additional module structure because the generated modules in `contracts.k` are now just empty modules. Instead, we generate a flat module structure that includes the user-supplied lemmas directly into the main module and that's it.

In this PR:

- We no longer generate `contracts.k` file with the associated `KFlatModule` in them, instead the `foundry.k` imports the user-suppiled lemmas directly and the definition they supply directly.
- Output is updated accordingly.

### DIFF
--- a/src/kontrol/kompile.py
+++ b/src/kontrol/kompile.py
@@ -16,7 +16,6 @@ from pyk.utils import ensure_dir_path, hash_str
 
 from . import VERSION
 from .kdist.utils import KSRC_DIR
-from .solc_to_k import Contract, contract_to_main_module
 from .utils import _read_digest_file, _rv_blue, console, kontrol_up_to_date
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/tests/integration/conftest.py
```diff
@@ -73,29 +78,35 @@ def foundry(foundry_root_dir: Path | None, tmp_path_factory: TempPathFactory, wo
...
+                            'includes': (),
+                            'requires': [
+                                str(TEST_DATA_DIR / 'lemmas.k'),
+                                str(TEST_DATA_DIR / 'cse-lemmas.k'),
+                                str(TEST_DATA_DIR / 'pausability-lemmas.k'),
+                                str(TEST_DATA_DIR / 'symbolic-bytes-lemmas.k'),
+                            ],
+                            'imports': [
+                                'LoopsTest:SUM-TO-N-INVARIANT',
+                                'ArithmeticCallTest:CSE-LEMMAS',
```
palinatolmach, line 94: Now that we import the lemmas directly, we can probably remove one of the two `CSE-LEMMAS` imports:
```
'ArithmeticCallTest:CSE-LEMMAS',
'CSETest:CSE-LEMMAS',
```
so we also don't get duplicate imports in `foundry.k` as we do now in this PR:
```
    imports public CSE-LEMMAS
    imports public CSE-LEMMAS
```
As you've said, we were previously importing all lemmas into all modules, so we didn't need both imports before either, but this duplication is just more obvious now. Maybe we can come up with a way to disambiguate (or, probably better, disallow) similarly-named lemma modules as a separate PR.

palinatolmach (approved):

src/tests/integration/conftest.py
ehildenb: I included a call to `unique` in the `flattened_labels`, which should be good to keep this from happening for now.

ehildenb: I'm running the output-updater on it.

PR #1011, opened 2025-04-09
Migrate from `poetry` to `uv`

tothtamas28:
Currently, updates of kontrol dependencies are stuck due to failing builds of more recent python packages, see failing CI runs [here](https://github.com/runtimeverification/kontrol/pull/1008). In particular, the python dependency builds fail due to stale versions of python build system packages. In `poetry2nix` these are always by default taken, with precedence, from `nixpkgs`, which in our case is out-of-date. Overriding of build system packages poses quite challenging, imo due to changes in `buildPythonPackage` from `nixpkgs`. This was observed in this [pull request](https://github.com/runtimeverification/k/pull/4746#event-16803952080).

Instead of hot patching build system packages and buildPythonPackages until we have an up-to-date `nixpkgs` again, [`uv2nix`](https://github.com/pyproject-nix/uv2nix) and [`pyproject.nix`](https://github.com/pyproject-nix/pyproject.nix) can be used. In particular, this allows us to instead use up-to-date python build system packages from [`build-system-pkgs`](https://github.com/pyproject-nix/build-system-pkgs). This in turn will make the python builds in this aforementioned [pull request](https://github.com/runtimeverification/kontrol/pull/1008) be able to successfully build.

In addition, migrating from `poetry` to `uv` allows us to use any build system backend instead of only `poetry-core`. Also, `uv` already adheres to recent `pyproject.toml` standards.


tothtamas28:
@juliankuners @F-WRunTime a few notes on how `pyproject.toml` can be efficiently migrated for other repos:
* The script [`migrate-to-uv`](https://github.com/mkniewallner/migrate-to-uv) does 99% of the job.
* It will translate `git` dependencies into an entry in the `[tool.uv.sources]` table. This field is only used by `uv`, so `pip install <source-directory>` won't work correctly this way. Instead, we can use [direct references](https://hatch.pypa.io/1.9/config/dependency/#direct-references) of `hatch`. To enable this feature, `tool.hatch.metadata.allow-direct-references = true` has to be set in `pyproject.toml`.

juliankuners:
I migrated `poetry2nix` to `uv2nix` in this [branch](https://github.com/runtimeverification/kontrol/tree/setup-uv2nix).

In particular:
- `poetry2nix` was replaced by `uv2nix`
- devShell was updated to provide `uv`
- uv from the stale nixpkgs revision was too old
  - therefore, as a temporary solution, it was overridden with an up-to-date revision
- `uv2nix` uses the alias `lib.match -> builtins.match` that was introduced in a more recent nixpkgs
  - therefore, as a temporary solution, the nixpkgs used by `uv2nix` was overridden with a [custom patched nixpkgs](https://github.com/juliankuners/nixpkgs/commit/e9a77bb24d408d3898f6a11fb065d350d6bc71f1) with the respective patch
- the previously broken overlay was fixed and is now working correctly
- I refactored the `kontrol` and `kontrol-pyk` derivations into their own files outside of the flake

juliankuners:
The CI is failing specifically on ARM Nix MacOS when trying to build `uv`. I will investigate this later once I have access to a MacOS machine.

juliankuners:
I should note that this pull request uses pre-built wheels from pypi instead of source distributions and building from source. With `poetry2nix`, source distributions were supported with a provided set of overrides that make source distributions compile successfully, in the general case. `uv2nix`, or rather `pyproject-nix`, do not want to provide these overrides out-of-the-box due to maintenance overhead observed with `poetry2nix`. Instead, these should be provided externally. There does not seem to be a community maintained set of overrides yet. Therefore `uv2nix` recommends to use pre-built distributions that are very likely to work out-of-the-box.

When building `kontrol` with `uv2nix` with source distributions, called `sdist` and configurable [here](https://github.com/runtimeverification/kontrol/blob/de8fe53aaad90a4c671df2a385788f5a8e26f6dc/nix/kontrol-pyk/default.nix#L49), then the build fails mostly due to missing build system packages for dependencies. I created a python script to create these overrides automatically.

Later, though, the build fails when building `multiprocess`, because the build process builds `multiprocess-0.70.17-py313-none-any.whl` when it's supposed to build `multiprocess-0.70.17-py310-none-any.whl` instead. It builds the wheel for the wrong python version. `multiprocess` uses setuptools and building in `uv2nix` is done by calling `uv build --python=... [...]`. I suppose this wrong behaviour is caused by a bug in either `uv` or `setuptools` and could potentially be worked around with a patch. Though I stopped there due to this introducing more overhead than benefiting us, for now. In the future, more options might be available and tooling more stable. I might create a draft pull request to record my progress and describe the situation for the long term.

palinatolmach:
There're a few more instructions where `poetry` has to be changed to `uv`, e.g., here https://github.com/runtimeverification/kontrol/blob/0cacafba4c86e7f96f58eb78e56d1a5f5c01a4a3/CONTRIBUTING.md?plain=1#L31

juliankuners:
Thanks for the review @palinatolmach!

I updated the remaining documentation to use `uv` instead of `poetry`. I also searched for the term `poetry` in the entire repository and found remaining usage of `poetry` in the CI job `Update Expected Output`. I also updated that one remaining CI job.

CONTRIBUTING.md
```diff
@@ -27,9 +27,9 @@ The first step is to develop and test your changes locally.
 ##### 1. Build Kontrol from source
 ```
 kup install k.openssl.secp256k1 --version v$(cat deps/k_release)
-poetry install
-poetry run kdist clean
-CXX=clang++-14 poetry run kdist --verbose build -j2 kontrol.*
+uv sync
```
tothtamas28, line 30:
```suggestion
```

If `.venv` does not exists, `uv run` initializes it automatically.

.github/workflows/update-expected-output.yml
```diff
@@ -24,8 +24,8 @@ jobs:
           container-name: kontrol-ci-integration-${{ github.sha }}
       - name: 'Build Kontrol'
         run: |
-            docker exec -u github-user kontrol-ci-integration-${GITHUB_SHA} /bin/bash -c 'poetry install'
-            docker exec -u github-user kontrol-ci-integration-${GITHUB_SHA} /bin/bash -c 'CXX=clang++-14 poetry run kdist --verbose build -j`nproc` kontrol.*'
+            docker exec -u github-user kontrol-ci-integration-${GITHUB_SHA} /bin/bash -c 'uv sync'
```
tothtamas28, line 27:
```suggestion
```

Same here.

CONTRIBUTING.md
juliankuners: Thanks, I resolved this.

.github/workflows/update-expected-output.yml
juliankuners: Thanks, I resolved this.

README.md
```diff
@@ -25,35 +25,37 @@ The fastest way is via the [kup package manager], with which you can do to get t
...
 
+
 #### Build using the virtual environment
 
 In order to build `kontrol`, you need to build these specific targets.
 Note that you can use `kontrol.base` (without [keccak](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/kdist/keccak.md) or [aux](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/kdist/kontrol_lemmas.md) lemmas), `kontrol.aux` (with aux lemmas), `kontrol.keccak` (with keccak lemmas) and `kontrol.full` (with all lemmas) instead of `kontrol.*` (which builds them all).
 
 ```sh
-poetry run kdist --verbose build -j2 kontrol.*
+uv run kdist --verbose build -j2 kontrol.*
```
palinatolmach, line 43: I'm getting an error saying `no targets found` when running it in zsh without quotes:
```suggestion
uv run kdist --verbose build -j2 "kontrol.*"
```

palinatolmach: The same applies to two similar commands below.

CONTRIBUTING.md
```diff
@@ -27,9 +27,8 @@ The first step is to develop and test your changes locally.
 ##### 1. Build Kontrol from source
 ```
 kup install k.openssl.secp256k1 --version v$(cat deps/k_release)
-poetry install
-poetry run kdist clean
-CXX=clang++-14 poetry run kdist --verbose build -j2 kontrol.*
+uv run kdist clean
+CXX=clang++-14 uv run kdist --verbose build -j2 kontrol.*
```
palinatolmach, line 31: Same issue with quotes:
```suggestion
CXX=clang++-14 uv run kdist --verbose build -j2 "kontrol.*"
```

palinatolmach (approved):
Approved with a comment about wrapping `kontrol.*` in quotes.

README.md
juliankuners: Thank you, resolved now.

CONTRIBUTING.md
juliankuners: Thank you, resolved now.

.github/workflows/Dockerfile
F-WRunTime, line 1: The foundry version is no longer available on ghcr.io 

If there's a specific version we're aware of we want to be using lets pin that and use it. 
Two options I propose here we always just pull on latest to build kontrol, @palinatolmach what do you think? 

Digging through tags looks like we could pick maybe use rc-1 as a tag option?

tothtamas28: > we always just pull on latest

It needs to be a fixed version for reproducibility.

palinatolmach: There's a list of [breaking changes](https://book.getfoundry.sh/guides/v1.0-migration) some of which are relevant to our test suite. I'll [try](https://github.com/runtimeverification/kontrol/pull/1014) bumping the version to a more recent pre-stable nightly version first, and will then start moving towards rc-1.

F-WRunTime: How long has this nightly version been available/ Can we pull it and publish to our Packages in RV so we don't lose this version or can we pick a stable version outside of the ngihtly pool?

F-WRunTime:
> ly considerably faster even wi

Awesome. I didn't realize this was a fork for the nixos pkgs repository. Do we want to propose these changes back to their repo? And reference what need from there? 
or is this fix just specific to us?

juliankuners:
> > ly considerably faster even wi
> 
> Awesome. I didn't realize this was a fork for the nixos pkgs repository. Do we want to propose these changes back to their repo? And reference what need from there? or is this fix just specific to us?

That's just a back-port of an alias that was only later introduced in `nixpkgs`: `lib.match` -> `builtins.match`. I used the same `nixpkgs` revision that we use as a base, in case the respective `uv2nix` flake uses any packages, which it does. I specifically override the `nixpkgs` input in `uv2nix`. Otherwise `uv2nix` fails due to the missing alias in `nixpkgs.lib`

.github/workflows/Dockerfile
palinatolmach: I tried going for a stable `rc-1` in the end, but it looks like they're not available in the Nix overlay, and only `nightly` ones are. So I chose a recently released [nightly-1da4d324652b3a61f7c7128a6d28f9d6239e8218](https://github.com/foundry-rs/foundry/releases/tag/nightly-1da4d324652b3a61f7c7128a6d28f9d6239e8218) that [is available](https://github.com/shazow/foundry.nix/commit/afa5358db2f4d1d89a496f6482e32a4cb1b2e2da), at least to unblock all the PRs, but I'll check with @juliankuners on how dependent we are on versions available at https://github.com/shazow/foundry.nix next week.

F-WRunTime (approved):

.github/workflows/Dockerfile
palinatolmach: Fixed in https://github.com/runtimeverification/kontrol/pull/1015.

PR #1012, opened 2025-04-10
Fixed expectRevert being applied to calls to cheatcodes

### LINKED ISSUE #993: Align `expectRevert` and `prank` behavior with Foundry
At the moment, Kontrol has a strict requirement on the `vm.prank` and `vm.expectRevert` ordering and is not mirroring Foundry behavior. Kontrol reports the first test as failing but Foundry doesn't:
```
    function test_SetNumber_reverts() public {
        vm.expectRevert();
        vm.prank(address(0x11));
        counter.setNumberOwner(1);
    }

    function test_SetNumber_reverts_reordered() public {
        vm.prank(address(0x11));
        vm.expectRevert();
        counter.setNumberOwner(1);
    }
```

A client also mentioned that sometimes teams will put an external call inside the expect revert to get some state so having the prank afterward avoids pranking on the view call.

lisandrasilva:
This PR fixes the `expectRevert` cheatcode implementation. According to the Foundry documentation: 

> After calling expectRevert, calls to other cheatcodes before the reverting call are ignored.
This means, for example, we can call `prank` immediately before the reverting call.

In the previous implementation, the call to the cheatcode was being ignored in the sense that was not being executed. This PR still executes cheatcode calls but does not check if these calls revert or not. 

Fixes #993.


src/kontrol/kdist/cheatcodes.md
```diff
@@ -579,12 +567,13 @@ WThe `#checkRevert` will be used to compare the status code of the execution and
...
          <k> #next [ _OP:CallOp ] ~> (.K => #checkRevert ~> #updateRevertOutput RETSTART RETWIDTH) ~> #execute ... </k>
          <callDepth> CD </callDepth>
-         <wordStack> _ : _ : _ : _ : _ : RETSTART : RETWIDTH : _WS </wordStack>
+         <wordStack> _ : ACCTTO : _ : _ : _ : RETSTART : RETWIDTH : _WS </wordStack>
          <expectedRevert>
            <isRevertExpected> true </isRevertExpected>
            <expectedDepth> CD </expectedDepth>
            ...
          </expectedRevert>
+         requires ACCTTO =/=Int #address(FoundryCheat)
```
anvacaru, line 576: Does this check also have to be added to the rule `[foundry.set.expectrevert.2]` below?

lisandrasilva: Do you think it's possible/likely to delegate a call to the cheatcode address?

anvacaru: I looked into this and wrote two tests, one for `delegatecall` and one for `staticcall` calling a cheatcode. 

The assertion at the end fails for `delegatecall`, whereas it does not fail for `staticcall`.

```Solidity
    function testDelegatecallCheatcodeReverts() external {
        (bool success, ) = VM_ADDRESS.delegatecall(
            abi.encodeWithSignature("roll(uint256)", 1337)
        );
        assertTrue(success);
        assertEq(block.number, 1337);
    }

    function testStaticcallcallCheatcode() external view {
        (bool success, ) = VM_ADDRESS.staticcall(
            abi.encodeWithSignature("roll(uint256)", 1337)
        );
        assertTrue(success);
        assertEq(block.number, 1337);
    }
 ```

So maybe it's worth adding the rule to cover `staticcall` for now. Next, in another issue/PR, we should check that Kontrol has the same behavior and ensure that delegating into `VM_ADDRESS` does not execute the cheat code.

Comment (palinatolmach) [reply src/kontrol/kdist/cheatcodes.md:576]: I think it sounds good! Let's add a rule for `staticcall` here, and ensure that Kontrol behaves similarly wrt `delegatecall` as a follow-up.

Comment (lisandrasilva) [reply src/kontrol/kdist/cheatcodes.md:576]: Will do it! 
I also tried the following tests with Foundry:
```solidity
function testIncrementAsNotOwner_2() public {
        vm.expectRevert(Unauthorized.selector);
        (bool success, ) = VM_ADDRESS.delegatecall(
            abi.encodeWithSignature("roll(uint256)", 1337)
        );
        upOnly.increment();
    }
```
And this test is passing when it shouldn't, since the function does not revert as expected. On the other hand, the following test fails:
```solidity
function testIncrementAsNotOwner_3() public {
        vm.expectRevert(Unauthorized.selector);
        (bool success, ) = VM_ADDRESS.delegatecall(
            abi.encodeWithSignature("roll(uint256)", 1337)
        );
        vm.prank(address(0));
        upOnly.increment();
    }
```
It seems that `delegatecall` to `VM_ADDRESS` clears the `expectRevert` cheatcode.

palinatolmach (approved):
Looks good, and I think it's reasonable leave the `delegatecall` investigation to a follow-up PR.

PR #1015, opened 2025-04-17
Bump foundry version used in CI

palinatolmach:
This PR bumps the version of Foundry we use in CI to a recent pre-release [rc-1](https://github.com/foundry-rs/foundry/releases/tag/rc-1) to avoid the issues with an older version being inaccessible in `ghrc.io` currently blocking all CI runs for kontrol. It also updates expected output of end-to-end tests accordingly.

To avoid changes in the contract bytecode and lemmas in tests, this PR adds `optimizer = true` to the `foundry.toml` files in the integration and end-to-end tests, since the optimizer has been disabled by default in Foundry since the version we've used before. These lemmas and expected output would still have to be updated once we bump the version of the `forge-std` library, and I'll share the documentation for it later.

The [shazow/foundry.nix](https://github.com/shazow/foundry.nix) overlays are only available for `nightly` versions, so I overwrote the version we use in `flake.nix` to `rc-1` with `overrideAttrs`.


flake.nix
```diff
@@ -133,6 +133,45 @@
...
+                      sha256 = "1zq726d3vqnlp5bn5ijzia8xbw804g6kwxws02ddf390s5mr7946";
+                    };
+                  };
+
+                  src = final.fetchzip {
+                    inherit (sources.${system}) url sha256;
+                    stripRoot = false;
+                  };
+                in
+                  final.stdenv.mkDerivation {
```
juliankuners, line 166: Is there a specific reason, why you repackaged `foundry-bin` with a new `final.stdenv.mkDerivation` instead of overriding `foundry-bin` from `github:shazow/foundry.nix` with `prev.foundry-bin.overrideAttrs`?

In case there is a reason, the `github:shazow/foundry.nix` input and `foundry.overlay` overlay should be dropped entirely. Otherwise, I'd suggest to override the existing derivation. The derivation from `shazow` also seems more [thorough](https://github.com/shazow/foundry.nix/blob/main/foundry-bin/default.nix), though our successful CI/CD runs suggest that this light weight derivation should be enough as well.

juliankuners (changes requested):
left comment

flake.nix
palinatolmach: No specific reason, I just didn't know it was an option. Let me fix it!

palinatolmach: Done in https://github.com/runtimeverification/kontrol/pull/1015/commits/2fd1b6fdc785218e353a9901582d339aa84bc4cb.

src/tests/integration/test-data/show/ForgetBranchTest.test_forgetBranch(uint256).expected
```diff
@@ -7,28 +7,37 @@
 │   src: test/ForgetBranch.t.sol:8:21
 │   method: test%ForgetBranchTest.test_forgetBranch(uint256)
 │
-│  (755 steps)
+│  (1000 steps)
 ├─ 3
+│   k: JUMP 6022 ~> #pc [ JUMP ] ~> #execute ~> CONTINUATION:K
```
anvacaru, line 12: So it seems there is one additional node tracked now.

juliankuners (approved):
The nix changes look good to me!

anvacaru (approved):

PR #1019, opened 2025-04-27
Refactors to simplify import structure

ehildenb:
I'm using a tool called `pydeps` to analyze the import structure of Kontrol and understand where it can be simplified. This PR does so.

The original import graph looks like this:

kontrol

This PR does the following:

- Move function `hex_string_to_int` from `solc_to_k` to `utils`, a more natural place for it.
- Adjusts some imports in `kontrol.cli` to go directly to the modules they're importing from, instead of through `kontrol.prove`.
- Removes the commands `kontrol compile` and `kontrol to-dot` (which I believe are unused), and the associated code with them.
- Factors out `foundry_view` command to be reused in `__main__` instead of having the code inline (removes some imports from `__main__`.
- Refactor the logic around `--hevm` to remove it from being imported directly into `__main__`.

The new import graph loosk like this (a bit flatter):

kontrol

I was using the following command to compute the imports graph (using `pydeps`: https://pydeps.readthedocs.io/en/latest/):

```
% pydeps -o kontrol.png -T png --no-show --cluster --max-bacon 4 src/kontrol
```

### DIFF
--- a/src/kontrol/__main__.py
+++ b/src/kontrol/__main__.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 import logging
 import sys
 from collections.abc import Iterable
@@ -39,7 +38,6 @@ from .hevm import Hevm
 from .kompile import foundry_kompile
 from .prove import foundry_prove
 from .solc import CompilationUnit
-from .solc_to_k import solc_compile
 from .utils import _LOG_FORMAT, _rv_blue, _rv_yellow, check_k_version, config_file_path, console, loglevel
 
 if TYPE_CHECKING:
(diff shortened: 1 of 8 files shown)


ehildenb:
A further potential optimization would be to remove `Hevm` altogether, and inline its logic into where it's used. It would not be very much code motion to do so, and I think is reasonable given that we haven't seen any other predicates we're using to determine success/failure yet.

ehildenb:
Another potential optimization to make would be to split out `foundry.py` into more specific submodules (like we have for `kompile.py` and `prove.py`), so that `foundry.py` actually just becomes the `Foundry` object and related definitions, and the different files become the various CLI commands. That would probably also reveal places where import structure could be simplified.

palinatolmach (approved):

tothtamas28 (approved):

PR #1021, opened 2025-04-28
Factor out Foundry modules

ehildenb:
~Blocked on: https://github.com/runtimeverification/kontrol/pull/1019~

This PR continues the refactoring in https://github.com/runtimeverification/kontrol/pull/1019 to factor out more of the module structure. In particular:

- A new module `kontrol.display` is factored out with the commands `foundry_show` and `foundry_view`, and all display-related helpers are moved there.
- Functionality related to processing state-diffs is lifted out of `prove.py` and directly into `__main__.py`, so that `state_record` no longer needs to be imported into `prove.py`.
- All the helpers, and the main command for state loading, are removed from `foundry.py` and moved to `state_record.py`.
- The function `Foundry.custom_view`, which did not really use much else from the Foundry class, is moved to `display` module.

In terms of import graph structure, we have that:

- `state_record` is no longer imported into `foundry.py`, only into `__main__.py`, so it's better isolated.
- `solc` is no longer imported into `foundry.py`, just into `display.py` where it's used.

### DIFF
--- a/src/kontrol/__main__.py
+++ b/src/kontrol/__main__.py
@@ -24,15 +24,13 @@ from .foundry import (
     foundry_section_edge,
     foundry_simplify_node,
     foundry_split_node,
-    foundry_state_load,
     foundry_step_node,
     foundry_unrefute_node,
     init_project,
-    read_recorded_state_diff,
-    read_recorded_state_dump,
 )
 from .kompile import foundry_kompile
 from .prove import foundry_prove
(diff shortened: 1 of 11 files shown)


ehildenb:
The new import structure is as follows:

kontrol

ehildenb:
I moved `custom_view` into `display`, and the import graph now looks like this:

kontrol

tothtamas28 (approved):

Issue #1022, opened 2025-04-30
Fix `uv` version in GH actions


palinatolmach:
Based on the updates happening to `uv.lock` in https://github.com/runtimeverification/kontrol/pull/1008, it seems that the version of `uv` used in actions such as the one updating versions is not fixed, which causes version discrepancy issues. It should be set to the same one that's being used in `kup` and `nix build`.

juliankuners:
In the GitHub job `.github/workflows/update-version.yml`, `uv` is installed with `astral-sh/setup-uv@v5`. The latest update of `v5` was to version [`v5.2.2`](https://github.com/astral-sh/setup-uv/releases/tag/v5), which is probably the last update of `v5`, as [`v6.0.0`](https://github.com/astral-sh/setup-uv/releases/tag/v6.0.0) was released.

This means that from now on, most probably, the version of `uv` is fixed in CI/CD for now. Though it remains uncoupled from the version that is used by kup, `nix build`, and `nix develop`. I'd suggest that we reorder the jobs in `update-version.yml` to move the installation of `nix` forward and don't install `uv` with the `astral-sh/setup-uv` job. Instead we can use `uv` with the `nix develop` shell like this: `nix develop .# --command uv [...]`.

palinatolmach:
Thanks @juliankuners! I agree that's a good option, and, I guess, the other way is to do something like this: https://github.com/runtimeverification/kontrol/compare/master...pin-ci-uv but with the correct version, so we're more explicit about the version we're using. I've been trying to find the place in the codebase where the version used by kup, `nix build`, and `nix develop` is specified — can you please point me to it?

juliankuners:
@palinatolmach Yes, fixing the version in the CI job would be more explicit. I guess we could print the version of `uv` on the command line in the CI job logs in case of using the `nix develop` shell. We also would need/should have a CI job for updating `uv2nix` anyway, I think, and could explicitly fix both `uv` and `uv2nix` to the same respective version.

The version of `uv` is specified in the `uv2nix` repository: [https://github.com/pyproject-nix/uv2nix/blob/e00f7117aaebaa1f9ff846976d655c5a4d919321/pkgs/uv-bin/srcs.json](https://github.com/pyproject-nix/uv2nix/blob/e00f7117aaebaa1f9ff846976d655c5a4d919321/pkgs/uv-bin/srcs.json)

[...]

PR #1023, opened 2025-05-08
Pin `uv` version in CI to 0.7.2

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/1022.
Related: https://github.com/runtimeverification/mir-semantics/pull/547

This PR pins the version of `uv` used in CI to 0.7.2 for it to stay fixed and consistent with `mir-semantics`, pins the version in `pyproject-build-systems`, and updates the `uv.lock` and `flake.nix` files accordingly.

It also fixes a code quality issue identified by an updated `mypy`.

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -50,8 +50,7 @@ WORKDIR /home/${USER}/workspace
 
 ENV PATH=/home/${USER}/.cargo/bin:/home/${USER}/.local/bin:/usr/local/bin/:${PATH}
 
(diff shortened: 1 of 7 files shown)


juliankuners:
I'd suggest that we also add a GitHub job in another pull request that periodically updates `uv2nix` and `pyproject-build-systems` in the flake inputs as well as fetches the respective version of `uv` that is being used by nix and updates the version in the GitHub jobs. `uv` is updated quite frequently and this ensures that the two versions always match.

PR #1025, opened 2025-05-13
CSE: assume that cheatcodes' expected `CALLDEPTH` is smaller than the current one

palinatolmach:
Follow up to https://github.com/runtimeverification/kontrol/pull/1008.


ehildenb:
Would it work with condition added to LHS of the calldepth:

```
andBool notBool (ACTIVE_CELL orBool PRAKNDEPTH_CELL <Int CALLDEPTH_CELL)
andBool notBool (EXPECTED_REVERT_CELL orBool REVERTDEPTH_CELL <Int CALLDEPTH_CELL)
```

ehildenb:
Let's also add a test of `prank` of a function `f` that calls `g`, so that we can examine the behavior of the interaction of pranking and CSE.

PR #1040, opened 2025-06-19
Console.log Support for Kontrol

anvacaru:
This PR adds support for Foundry's `console.log` functionality in Kontrol proofs, enabling developers to see debug output during formal verification.

Added a new rule (`FOUNDRY_ACCOUNTS.console.log`) to intercept `STATICCALL` instructions to the console address (0x000000000000000000636F6e736F6c652e6c6f67).
This rule is also added to the list of `cut-point` rules and triggers a custom-step execution that decodes the calldata of the `log` function and prints it.
The calldata decoding is based on a function signature mapping (`CONSOLE_SELECTORS` maps all the function selectors from the `console` library to their parameter types) and uses the eth-abi library to decode the calldata based on the data types used in said log function.
The `console.log` output is displayed twice: first using the informational logging system (`--verbose` is required to see it) and then at the end of execution (also prints for non-verbose mode).
 

Some other changes:
 - Test contract balance: Updated initial test contract balance to `2^95 wei` to match Foundry's default.
 - Changed the `#address ( Contract )` tags from `macro` to `alias` to easily spot the predefined contracts in a KCFG.


src/kontrol/foundry.py
```diff
@@ -266,6 +278,31 @@ def _filter_constraints_by_simplification(
...
+        selector_token = subst['###SELECTOR']
+        data = subst['###DATA']
+        assert type(selector_token) is KToken
+
+        try:
+            if type(data) is KToken:
+                selector = int(selector_token.token)
+                if selector in CONSOLE_SELECTORS:
+                    param_types = CONSOLE_SELECTORS[selector]
+                    decoded = decode(param_types, ast.literal_eval(data.token))
```
palinatolmach, line 291: Can we catch and print out a warning if `decode` or `ast.literal_eval` fails for a given token?

palinatolmach (approved):
Looks good, I left one comment! 

 Please also consider factoring out this change to a separate PR: 
> Test contract balance: Updated initial test contract balance to 2^95 wei to match Foundry's default.

and adding a unit test for `console.log` calldata decoding, and an integration test for the Solidity test using `console.log`.

src/kontrol/foundry.py
anvacaru: I'm not sure I get this question. The `catch` on line 299 should catch any error with `decode` and `ast.literal_eval`, I think.

src/kontrol/foundry.py
```diff
@@ -266,6 +277,28 @@ def _filter_constraints_by_simplification(
...
 
+    def _exec_console_log_custom_step(self, subst: Subst, cterm: CTerm, _c: CTermSymbolic) -> KCFGExtendResult | None:
+        selector_token = subst['###SELECTOR']
+        data = subst['###DATA']
+        assert type(selector_token) is KToken
+
+        try:
+            if type(data) is KToken:
+                selector = int(selector_token.token)
+                print(data)
```
palinatolmach, line 288: Is this a debugging statement that should be removed by any chance?
```suggestion
                print(data)
```

palinatolmach (commented):
thanks for adding the test! added another comment:

src/kontrol/foundry.py
anvacaru: Great catch! Thanks!

PR #1043, opened 2025-07-03
deprecate execution of `testFail*` tests

anvacaru:
This PR removes support for Solidity test methods prefixed with `testFail`, `checkFail`, and `proveFail` when running Kontrol. Instead, a deprecation warning is displayed.

- Removed all testFail test entries from integration test data files
- Removed the `failing` parameter from `_final_cterm` and related logic that handled different success/failure expectations for `testFail` methods
image

### DIFF
--- a/src/kontrol/prove.py
+++ b/src/kontrol/prove.py
@@ -812,7 +812,7 @@ def _update_cterm_from_node(cterm: CTerm, node: KCFG.Node, config_type: ConfigTy
             'ACCOUNTS_CELL',
             'NUMBER_CELL',
             'TIMESTAMP_CELL',
-            'BASEFEE_CEL',
+            'BASEFEE_CELL',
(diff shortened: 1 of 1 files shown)


src/kontrol/prove.py
```diff
@@ -808,7 +812,7 @@ def _update_cterm_from_node(cterm: CTerm, node: KCFG.Node, config_type: ConfigTy
             'ACCOUNTS_CELL',
             'NUMBER_CELL',
             'TIMESTAMP_CELL',
-            'BASEFEE_CELL',
+            'BASEFEE_CEL',
```
palinatolmach, line 815: Probably a typo:
```suggestion
            'BASEFEE_CELL',
```

palinatolmach (approved):
Left a comment about a typo, otherwise looks good!

PR #1045, opened 2025-07-07
Changes to CLI output when a proof is failing

anvacaru:
Updating the CLI output. Changes:

1. Replaced the failure reason (and the remaining implication) with the `status code` and `output` values.

   - added a new `KontrolAPRFailureInfo(APRFailureInfo)` class with a `print_with_additional_info(status_codes: list[str], outputs: list[str])` method, which unlike the [APRFailureInfo.print()](https://github.com/runtimeverification/k/blob/40d1aa8505043451a08a7436ed076e7d85311756/pyk/src/pyk/proof/reachability.py#L1018) method, does not display the remaining implication but instead shows the status codes and output bytestrings of the failing nodes.

2. Pretty print K keywords using regular expressions.
   - Implemented a `replace_k_words` method which is using regular expressions to replace given K keywords.

3. Added telegram link.
 

Before:
image

Current implementation:
image


src/tests/integration/test-data/show/minimized/AssertTest.testFail_expect_revert().expected
```diff
@@ -529,7 +529,4 @@ Failing nodes:
 
 Join the Runtime Verification Discord server (https://discord.com/invite/CurfmXNtbN) or Telegram group (https://t.me/rv_kontrol) for support.
 
-See `foundry_success` predicate for more information:
-https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/kdist/foundry.md#foundry-success-predicate
-
```
anvacaru, line 534: It seems like there is a different method for `show` that handles the post-processing and display of the proof information:

[print_failure_information](https://github.com/runtimeverification/evm-semantics/blob/4838fe237e5c5b51caf0789bbacafbc25fd92cbb/kevm-pyk/src/kevm_pyk/utils.py#L179) called in [foundry_show](https://github.com/runtimeverification/kontrol/blob/7296feea3131144f7418bcbac90dabaaef3affc6/src/kontrol/display.py#L166).

palinatolmach: I mean, changes other than the success predicate link being removed.

palinatolmach (approved):

src/tests/integration/test-data/show/minimized/AssertTest.testFail_expect_revert().expected
anvacaru: Opened an issue for this #1048.

PR #1068, opened 2025-08-21
Show failing output in setUp and constructor

anvacaru:
Currently, when a setUp method fails, we only get a notification and a stack trace, which does not provide valuable insight.

In this PR, (1) I'm refactoring the logic used to process the output at the end of a proof to also process the output in case of a setUp/constructor failure; and (2), replacing `raise ValueError()` with `sys.exit()` to avoid displaying the stack trace.

Before:
image

After:
image


palinatolmach (approved):
lgtm, very timely change!

PR #1081, opened 2025-09-26
Add `KontrolTest` generation to `kontrol init`

palinatolmach:


src/kontrol/utils.py
```diff
@@ -716,3 +716,137 @@ def decode_log_message(token: str, selector: int) -> str | None:
...
+        uint256 offset,
+        uint256 width
+    ) internal view returns (uint256) {
+        // `offset` and `width` must not overflow the slot
+        assert(offset + width <= 32);
+        // Slot read mask
+        uint256 mask;
+        unchecked {
+            mask = (2 ** (8 * width)) - 1;
+        }
```
anvacaru, line 756: Some additional validation: width should be >0, address should be non-zero;
Additionally, `mask = (2 ** (8 * width)) - 1;` should overflow if `width` is 32 bytes.

```suggestion
    function _loadData(
        address contractAddress,
        uint256 slot,
        uint256 offset,
        uint256 width
    ) internal view returns (uint256) {
        require(contractAddress != address(0), 'Invalid contract address');
        require(width > 0, 'Width must be greater than 0');
        // `offset` and `width` must not overflow the slot
        assert(offset + width <= 32);
        
        // Slot read mask - handle width = 32 case to prevent overflow
        uint256 mask;
        if (width == 32) {
            mask = type(uint256).max;
        } else {
            unchecked {
                mask = (2 ** (8 * width)) - 1;
            }
        }
```

palinatolmach: thanks!

src/kontrol/utils.py
```diff
@@ -716,3 +716,144 @@ def decode_log_message(token: str, selector: int) -> str | None:
...
+    function _storeData(address contractAddress, uint256 slot, uint256 offset, uint256 width, uint256 value) internal {
+        // `offset` and `width` must not overflow the slot
+        assert(offset + width <= 32);
+        // and `value` must fit into the designated part
+        assert(width == 32 || value < 2 ** (8 * width));
+        // Slot update mask
+        uint256 maskLeft;
+        unchecked {
+            maskLeft = ~((2 ** (8 * (offset + width))) - 1);
+        }
```
lucasmt, line 780: Should we also split this into the `offset + width == 32` case, to make the intention clearer?

src/kontrol/utils.py
```diff
@@ -716,3 +716,144 @@ def decode_log_message(token: str, selector: int) -> str | None:
...
+        uint256 mappingSlot,
+        uint256 key,
+        uint256 subSlot
+    ) internal view returns (uint256) {
+        bytes32 hashedSlot = keccak256(abi.encodePacked(key, mappingSlot));
+        return _loadData(contractAddress, uint256(hashedSlot) + subSlot, 0, 32);
+    }
+    function _loadAddress(address contractAddress, uint256 slot) internal view returns (address) {
+        return address(uint160(_loadData(contractAddress, slot, 0, 20)));
+    }
```
lucasmt, line 827: We could add `_loadBytes32`, and `...Address` and `...Bytes32` versions of `_loadMapping...` and `_storeMapping...` as well. if we want to be complete.

src/kontrol/utils.py
```diff
@@ -716,3 +716,144 @@ def decode_log_message(token: str, selector: int) -> str | None:
...
+        // Current slot value
+        uint256 slotValue = uint256(vm.load(contractAddress, bytes32(slot)));
+        // Isolate and return data to retrieve
+        return mask & (slotValue >> shift);
+    }
+    function _storeData(address contractAddress, uint256 slot, uint256 offset, uint256 width, uint256 value) internal {
+        // `offset` and `width` must not overflow the slot
+        assert(offset + width <= 32);
+        // and `value` must fit into the designated part
+        assert(width == 32 || value < 2 ** (8 * width));
```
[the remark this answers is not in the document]
palinatolmach: Thanks! Addressed in https://github.com/runtimeverification/kontrol/pull/1081/commits/713bae53c80ce9c5b3d4aaefb95f9538fe2a047f.

src/kontrol/utils.py
```diff
@@ -716,3 +716,144 @@ def decode_log_message(token: str, selector: int) -> str | None:
...
+    function _loadData(
+        address contractAddress,
+        uint256 slot,
+        uint256 offset,
+        uint256 width
+    ) internal view returns (uint256) {
+        require(contractAddress != address(0), 'Invalid contract address');
+        require(width > 0, 'Width must be greater than 0');
+        // `offset` and `width` must not overflow the slot
+        assert(offset + width <= 32);
```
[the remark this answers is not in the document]
palinatolmach: Agreed, addressed in https://github.com/runtimeverification/kontrol/pull/1081/commits/713bae53c80ce9c5b3d4aaefb95f9538fe2a047f.

src/kontrol/utils.py
```diff
@@ -716,3 +716,144 @@ def decode_log_message(token: str, selector: int) -> str | None:
...
+        
+        // Slot read mask - handle width = 32 case to prevent overflow
```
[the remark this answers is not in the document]
palinatolmach: Agreed, thank you! I removed the special handling of `width == 32` but left a comment with what you've said in yours!

PR #1089, opened 2025-10-07
Add `pyproject-nix` overlay

juliankuners:
Previously, when RV nix derivations were using `poetry2nix`, overriding a nix input in nix was also overriding the python code and resources used in `poetry2nix`. This allowed for easy overrides of python and K semantics changes by utilizing nix overrides.

With the migration to `uv` and `uv2nix`, this was behavior was removed due to the respective code not documenting it's purpose. In addition, due to `uv2nix` relying on state-of-the-art python nix packaging tooling `pyproject-nix` instead of `nixpkgs`, ensuring identical override behaviour requires more nix logic.

This PR re-introduces this kind of behaviour with `uv2nix`, though it depends on newly introduced overlays in the k nix flake.

### Example
You can override k by running: `kup install kontrol --override kevm/k-framework ~/path/to/local-k-checkout`. This will now also build kontrol with the python and K code in `~/path/to/local-k-checkout`.

### DIFF
--- a/nix/kontrol-pyk-pyproject/default.nix
+++ b/nix/kontrol-pyk-pyproject/default.nix
@@ -1,0 +1,45 @@
+{
+  lib,
+  callPackage,
+  nix-gitignore,
+  stdenvNoCC,
+
+  uv2nix,
+  solc_version ? null,
+}:
+let
+  # patches cannot yet be applied to uv workspaces, so we use a derivation containing the src instead
+  src = stdenvNoCC.mkDerivation {
+    name = "kontrol-pyk-src";
+    src = callPackage ../kontrol-source { };
+
+    dontConfigure = true;
+    dontBuild = true;
+
+    postPatch = ''
+      ${lib.strings.optionalString (solc_version != null) ''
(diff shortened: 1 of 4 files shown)


juliankuners:
K releases are currently blocked. As this PR depends on recently introduced nix overlays in K, this PR will fail until a new release of K was published and merged into `mir-semantics`.

Also the respective [`pyproject-nix` overlay PR](https://github.com/runtimeverification/evm-semantics/pull/2802) in `kevm` must be merged prior to this PR.

nix/kontrol-pyk-pyproject/default.nix
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  # load a uv workspace from a workspace root
+  workspace = uv2nix.lib.workspace.loadWorkspace {
+    workspaceRoot = "${src}";
+  };
+
+  # create overlay
+  lockFileOverlay = workspace.mkPyprojectOverlay {
+    # prefer "wheel" over "sdist" due to maintance overhead
```
yiyi-wang-rv: Typo `maintenance`.

PR #1127, opened 2026-03-24
Exclude `out/lint` from the `Foundry.contracts` property glob.

anvacaru:
`forge lint` writes minimal artifacts that lack the AST section, causing the error below to always trigger on the second run, as the linter runs automatically.

```py
"Must have 'ast' field in solc output. Make sure `ast = true` is present in `foundry.toml` and run `forge clean`
``` 
In this PR, I'm omitting all the contracts under `out/lint` as the artifacts are duplicated and not needed by kontrol.

Other Cleanups:
- `find_enums` function definition is moved out of the loop, as originally it was re-created as a new function object on every iteration.
- Eliminated `str↔Path` round-trip: glob results stay as Path objects throughout.
- `json_path.stem` replaces `split('/')[-1][:-5]`: idiomatic and OS-safe.
- Three list-comprehension passes collapsed into a single sorted() generator expression.
- Added a default value for `dct.get('nodes', [])` in `find_enums` to avoid `KeyError` on leaf AST nodes.
- Compute `enum_max` once and reuse instead of calling `len(dct['members'])` twice.

### DIFF
--- a/src/kontrol/foundry.py
+++ b/src/kontrol/foundry.py
@@ -537,17 +537,16 @@ class Foundry:
         def find_enums(dct: dict) -> None:
             if dct['nodeType'] == 'EnumDefinition':
                 enum_name = dct['canonicalName']
-                enum_max = len([member['name'] for member in dct['members']])
+                enum_max = len(dct['members'])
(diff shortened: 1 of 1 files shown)


juliankuners (approved):
LGTM

PR #1129, opened 2026-03-30
Introducing changes to fix cachix pin and push failures in CI

F-WRunTime:
Failures with cachix are due to timeouts from cloudflare but the requests still complete. 

Per similar changes in [mir-semantics](https://github.com/runtimeverification/mir-semantics/pull/951/changes#diff-59cd4bfedd5e699e84dab588cd4c5bf931e01b112053cead0f7ce3d708d651d6), applying those changes to kontrol to clean up failed release jobs.

### DIFF
--- a/.github/scripts/check-cachix-pin.sh
+++ b/.github/scripts/check-cachix-pin.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Kup relies on cachix registry k-framework-binary.
+CACHE="k-framework-binary"
+OWNER_REPO="${OWNER_REPO:-$(git remote get-url origin | sed -E 's#(git@github.com:|https://github.com/)##; s#\.git$##')}"
+REV="${REV:-${GITHUB_SHA:-$(git rev-parse HEAD)}}"
+UNAME_S="$(uname -s)"
+UNAME_M="$(uname -m)"
+case "${UNAME_S}-${UNAME_M}" in
+  Linux-x86_64) SYSTEM="x86_64-linux" ;;
+  Linux-aarch64 | Linux-arm64) SYSTEM="aarch64-linux" ;;
+  Darwin-x86_64) SYSTEM="x86_64-darwin" ;;
+  Darwin-arm64) SYSTEM="aarch64-darwin" ;;
+  *)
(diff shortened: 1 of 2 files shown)


.github/scripts/check-cachix-pin.sh
```diff
@@ -0,0 +1,86 @@
...
+    )"
+    if [ -z "$STORE_PATH" ]; then
+      PIN_STATUS="pin-missing"
+      PUSH_STATUS="000"
+      ALL_OK=0
+      {
+        echo "key-${PKG}: ${KEY}"
+        echo "pin-status-${PKG}: ${PIN_STATUS}"
+        echo "push-http-${PKG}: ${PUSH_STATUS}"
+      }
```
anvacaru: If the Cachix API response shape ever changes or returns an unexpected structure, maybe we should also print the raw JSON.

anvacaru: Should these messages also be appended to `$SUMMARY`?

.github/scripts/check-cachix-pin.sh
```diff
@@ -0,0 +1,86 @@
...
+    PUSH_STATUS="$(curl -sS -o /dev/null -w '%{http_code}' "$PUSH_NARINFO_URL")" || PUSH_STATUS="000"
+    if [ "$PUSH_STATUS" != "200" ]; then
+      ALL_OK=0
+    fi
+
+    {
+      echo "key-${PKG}: ${KEY}"
+      echo "store-path-${PKG}: ${STORE_PATH}"
+      echo "pin-status-${PKG}: ${PIN_STATUS}"
+      echo "push-http-${PKG}: ${PUSH_STATUS}"
```
anvacaru, line 81: Same question here, should these messages also be appended to `$SUMMARY`?

.github/scripts/check-cachix-pin.sh
F-WRunTime: It's certainly a risk we can dump the json for review for that scenario. But probably best to just dump to logs. I think there's quite a bit in the message just fluff really that wouldn't be very informative with the exception of message structure changing.

.github/scripts/check-cachix-pin.sh
F-WRunTime: Yes certainly can

anvacaru (approved):

PR #1154, opened 2026-06-17
Fall back to prove TOML for RPC proof commands

### LINKED ISSUE #1134: `kontrol simplify-node --config-profile <profile>` ignores `kontrol.toml` and specified profile
`kontrol simplify-node --config-profile <profile>` ignores `kontrol.toml` and the specified profile.

This can be reproduced with the kontrol project in this repository: https://github.com/juliankuners/test-simplify-node

peter941221:
Fixes #1134

When `kontrol.toml` only defines proof-time settings under `[prove.<profile>]`, the RPC proof commands that accept `--config-file` and `--config-profile` read no TOML settings at all. `pyk.parse_toml_args(...)` only loads `[args.command]`, so `simplify-node`, `step-node`, `section-edge`, and `get-model` ignore the generated `kontrol.toml` layout and the reproducer from #1134.

This patch makes those commands load `[prove.<profile>]` first and then overlay any command-specific TOML section on top. Command-local overrides still win when both sections exist.

Tests cover:
- fallback from `[prove.default]` for all four RPC proof commands
- fallback from a selected non-default prove profile
- command-specific sections overriding the prove fallback

Validation:
- `uv run make check`
- `uv run pytest src/tests/unit/test_toml_args.py -q`
- `uv run make test` still fails on a clean `master` clone with the same existing `pytest-xdist` collection mismatch in `src/tests/unit/test_natspec.py`, so this PR does not introduce that failure

### DIFF
--- a/src/kontrol/__main__.py
+++ b/src/kontrol/__main__.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import sys
 from collections.abc import Iterable
+from copy import copy
 from typing import TYPE_CHECKING
 
 from pyk.cli.pyk import parse_toml_args
@@ -48,6 +49,7 @@ from .utils import (
 )
 
 if TYPE_CHECKING:
+    from argparse import Namespace
     from pathlib import Path
     from typing import Final, TypeVar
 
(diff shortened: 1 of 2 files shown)


peter941221:
@palinatolmach I tightened the regression coverage for #1134 and marked this ready for review.

Validation on this branch:
- uv run make check
- uv run pytest src/tests/unit/test_toml_args.py -q

I also checked uv run make test against a clean master clone. It hits the same existing pytest-xdist collection mismatch in src/tests/unit/test_natspec.py, so that failure is not introduced by this PR.

PR #1167, opened 2026-09-21
.github/with-docker: make compatible with podman

anvacaru:
Import the changes from https://github.com/runtimeverification/evm-semantics/pull/2881 and https://github.com/runtimeverification/evm-semantics/pull/2880 to add compatibility for Podman.

Change the `kontrol-ci-integration` container names to unique ones instead: `kontrol-ci-cse`, `kontrol-ci-e2e`, and `kontrol-ci-expected-output`.
Adding the  `--ulimit nofile=65536:65536` flag to the build command in the Docker Tests stage.


palinatolmach (approved):
lgtm, just a couple of minor notes:
- the comment in the `docker` job points to `.github/actions/with-docker` for the reasoning behind the pre-run `rm --force`, but `action.yml` has no comment there;
- `--rm` on `docker run` is now redundant due to the explicit teardown, so I believe it can be dropped, but should be harmless to keep too.

